PR #3, opened 2025-04-10
Add Makefile and initial CI workflow

stuart-mclaren-hpe:
There is just enough here to prevent checking in code that won't compile.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,22 @@
+# Copyright 2025 Hewlett Packard Enterprise Development LP
+
+name: Build
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        go: [ '1.22.3' ]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go }}
+      - run: go version
+      - run: make build
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,6 @@
+#(C) Copyright 2025 Hewlett Packard Enterprise Development LP
+#
+# Note: this Makefile works with GNUMake and BSDMake
+#
+build:
+	go build ./...


eamonnotoole (approved):

eamonnotoole (commented):
## 📋 Review: Build Configuration

**PR:** [#3 — Add Makefile and initial CI workflow](https://github.com/HPE/terraform-provider-hpe/pull/3)

**Scope:** 🎯 PR diff vs `origin/main` — 2 files · agents: patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 1 |

---

.github/workflows/build.yaml
```diff
@@ -0,0 +1,22 @@
...
+name: Build
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+jobs:
+  build:
+    runs-on: ubuntu-24.04
```
eamonnotoole: 🔵 **Recommendation** — Consider updating `actions/checkout` (to `@v4`) and `actions/setup-go` (to `@v5`). The `@v2` versions use deprecated Node.js runner versions and will generate warnings during CI runs.

Makefile
```diff
@@ -0,0 +1,6 @@
+#(C) Copyright 2025 Hewlett Packard Enterprise Development LP
+#
+# Note: this Makefile works with GNUMake and BSDMake
+#
+build:
+	go build ./...
```
eamonnotoole: 🔴 **Gap — `go build ./...` does not compile test files, so the PR's stated contract is not met.**

`go build` skips `_test.go` files entirely, so a test file that fails to typecheck passes this CI green. Verified: a package containing `var broken int = "this does not compile"` in `a_test.go` gives `go build ./...` → exit 0, while `go vet ./...` → exit 1 catching it.

For a Terraform provider, where unit and acceptance tests will be a large share of the tree, this is the main hole in *"just enough here to prevent checking in code that won't compile"*.

```make
build:
	go build ./... && go vet ./...
```

or add a `test:` target running `go test ./...` and invoke it from CI. `go vet ./...` already exits 0 on this tree, so adding it now cannot turn CI red.

Makefile
```diff
@@ -0,0 +1,6 @@
+#(C) Copyright 2025 Hewlett Packard Enterprise Development LP
+#
+# Note: this Makefile works with GNUMake and BSDMake
+#
+build:
```
eamonnotoole: 🟠 **Violation — `build` is not declared phony, giving a silently-green CI.**

This is a latent failure mode for the *only* check CI runs. Verified: with a `build` directory present, `make build` prints ``make: `build' is up to date.`` and exits 0 **without invoking the compiler at all** — CI reports success having compiled nothing. A `build/` output directory is a plausible near-term addition here, at which point CI silently stops testing anything.

```make
.PHONY: build
build:
	go build ./...
```

`.PHONY` is documented as a special target in both GNU make and BSD make, so this does not weaken the portability claim on line 3.

eamonnotoole (commented):
## 📋 Review: `terraform-provider-hpe` build infrastructure

**PR:** [#3 — Add Makefile and initial CI workflow](https://github.com/HPE/terraform-provider-hpe/pull/3)

**Scope:** 🎯 PR diff vs `origin/main` — 2 files · agents: patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 1 |
| 🟠 Violations | 2 |
| 🟡 Warnings | 3 |
| 🔵 Recommendations | 2 |

**Verdict:** 🔴 **FAIL** — 8 issues found (3 critical)

The changeset is two new files (`.github/workflows/build.yaml`, `Makefile`, 28 lines). There are no resources, data sources, schema or SDK surfaces in this checkout yet, so spec-coverage and structure review lanes were not applicable and UI validation was skipped. The review is confined to build/test infrastructure.

Judged against the PR's own stated contract — *"just enough here to prevent checking in code that won't compile"* — there is one real hole and two incorrect patterns. Everything below was verified empirically or against the GitHub API, not asserted from memory.

---

### 🔧 1. Build & CI Infrastructure

#### 🔴 Gaps

**🔴 `Makefile:6` — `go build ./...` does not compile test files, so the stated contract is not met**
`go build` skips `_test.go` files entirely. A test file that fails to typecheck passes this CI green. Verified: a package containing `var broken int = "this does not compile"` in `a_test.go` gives `go build ./...` → exit 0, while `go vet ./...` → exit 1 with `cannot use "this does not compile" (untyped string constant) as int value`. For a Terraform provider, where unit and acceptance tests will be a large share of the tree, this is the main gap in "prevent checking in code that won't compile". Fix: `go build ./... && go vet ./...`, or add a `test:` target running `go test ./...` and invoke it from CI. `go vet ./...` already exits 0 on this tree, so adding it now cannot turn CI red.

#### 🟠 Violations

**🟠 `.github/workflows/build.yaml:17-18` — both actions are pinned to a retired Node runtime**
`actions/checkout@v2` and `actions/setup-go@v2` each declare `runs: using: node12` in their `action.yml` (verified by reading the `v2` tag of both repos). The Node 12/16 runtimes are retired on GitHub-hosted runners, so these are force-migrated onto a runtime they were never built against and emit deprecation warnings on every run. Current majors are `actions/checkout@v7.0.1` and `actions/setup-go@v7.0.0` (verified via the releases API) — five majors behind. Fix: bump both to `@v7`. `setup-go` v7 additionally enables module caching by default, which v2 has no support for.

**🟠 `Makefile:5` — `build` is not declared phony, giving a silently-green CI**
This is a latent failure mode for the *only* check CI runs. Verified: with a `build` directory present, `make build` prints ``make: `build' is up to date.`` and exits 0 without invoking the compiler at all — CI reports success having compiled nothing. A `build/` output directory is a plausible near-term addition to this repo, at which point CI stops testing anything without any signal. Fix: add `.PHONY: build` above line 5. `.PHONY` is documented as a special target in both GNU make and BSD make, so it does not weaken the portability claim on line 3.

#### 🟡 Warnings

**🟡 `.github/workflows/build.yaml:12` — single-OS matrix leaves build-constrained files unchecked**
`go build` also skips files excluded by build constraints for the current platform, and the job runs only on `ubuntu-24.04` with no OS dimension in the matrix. A `//go:build windows` or `//go:build darwin` file that does not compile would never be caught — relevant for a provider that ships darwin and windows binaries. Latent today (no build-constrained files exist yet). Fix: add an `os:` dimension to the matrix with `runs-on: ${{ matrix.os }}`, or at minimum a `GOOS=windows go build ./...` step.

**🟡 `.github/workflows/build.yaml:10` — no `permissions:` block**
The job inherits the repository/org default `GITHUB_TOKEN` scopes rather than declaring least privilege. Fork PRs receive a read-only token under `pull_request` regardless, so exposure is limited to the push-to-`main` path, but a job that only reads source should say so explicitly. Fix: add a top-level `permissions:` block with `contents: read` above `jobs:`.

**🟡 `.github/workflows/build.yaml:15` — Go version hand-duplicated from `go.mod`, and not actually pinned**
`'1.22.3'` is duplicated between the matrix and the `go` directive in `go.mod:3`; they agree today but will drift. The matrix entry is also not a true pin: with the default `GOTOOLCHAIN=auto`, raising `go.mod`'s `go` directive above 1.22.3 makes the 1.22.3 toolchain installed by `setup-go` silently download and use the newer one. Fix: use `go-version-file: go.mod` (supported since `setup-go` v3.1) and drop the single-element matrix, or set `GOTOOLCHAIN: local` if the pinned version is meant to be authoritative.

#### 🔵 Recommendations

**🔵 `.github/workflows/build.yaml:9` — no `concurrency:` group**
Every push to an open PR queues an additional full run instead of superseding the previous one. Fix: add a `concurrency:` group keyed on `${{ github.workflow }}-${{ github.ref }}` with `cancel-in-progress: true`.

**🔵 `.github/workflows/build.yaml:1` — copyright header diverges from the repo's existing style**
The header reads `# Copyright 2025 Hewlett Packard Enterprise Development LP`, dropping the `(C)` marker used by `subprovider/subprovider.go:1` (`// (C) Copyright 2025 ...`) and by `Makefile:1` (`#(C) Copyright 2025 ...`). Fix: use `# (C) Copyright 2025 Hewlett Packard Enterprise Development LP` here and add the missing space after `#` in `Makefile:1`, so all three files match.

---

### ✅ Verified clean

- **`make build` is green** on this tree at `6db2196` (`go build ./...` → exit 0, working tree unchanged). `go vet ./...` → exit 0 and `gofmt -l .` → no output, so the recommended `go vet` addition would not turn CI red today.
- **The portability claim on `Makefile:3` holds.** The file uses only POSIX constructs, and the recipe on line 6 is indented with a real tab (confirmed with `od -c`), which BSD make requires. `build` is the first target, so bare `make` runs it.
- **Trigger config and runner image** (`build.yaml:5-9`, `:12`) are correct: `push`-to-`main` plus `pull_request` is right for this purpose, and `ubuntu-24.04` is properly version-pinned rather than `ubuntu-latest`.

---

*Reported for awareness only — excluded from the counts and verdict above.*

- **`go.mod:3`** — the `go` directive pins a full patch version (`go 1.22.3`) rather than the conventional `go 1.22`. Legal since Go 1.21, but it tightens `GOTOOLCHAIN` semantics and is the value the new CI matrix duplicates, so it is worth settling alongside the warning on `build.yaml:15`. Untouched by this PR.

---

.github/workflows/build.yaml
```diff
@@ -0,0 +1,22 @@
...
+    strategy:
+      matrix:
+        go: [ '1.22.3' ]
+    steps:
+      - uses: actions/checkout@v2
```
eamonnotoole: 🟠 **Violation — both actions are pinned to a retired Node runtime.**

`actions/checkout@v2` (line 17) and `actions/setup-go@v2` (line 18) each declare `runs: using: node12` in their `action.yml` — verified by reading the `v2` tag of both repos. The Node 12/16 runtimes are retired on GitHub-hosted runners, so these are force-migrated onto a runtime they were never built against and emit deprecation warnings on every run.

Current majors are `actions/checkout@v7.0.1` and `actions/setup-go@v7.0.0` (verified via the releases API) — five majors behind.

```yaml
      - uses: actions/checkout@v7
      - uses: actions/setup-go@v7
```

`setup-go` v7 additionally enables module caching by default, which v2 has no support for.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,22 @@
+# Copyright 2025 Hewlett Packard Enterprise Development LP
+
+name: Build
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
eamonnotoole: 🔵 **Recommendation — no `concurrency:` group.**

Every push to an open PR queues an additional full run instead of superseding the previous one.

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

.github/workflows/build.yaml
```diff
@@ -0,0 +1,22 @@
+# Copyright 2025 Hewlett Packard Enterprise Development LP
```
eamonnotoole: 🔵 **Recommendation — copyright header diverges from the repo's existing style.**

This reads `# Copyright 2025 ...`, dropping the `(C)` marker used by `subprovider/subprovider.go:1` (`// (C) Copyright 2025 ...`) and by `Makefile:1` (`#(C) Copyright 2025 ...`).

Suggest `# (C) Copyright 2025 Hewlett Packard Enterprise Development LP` here, plus the missing space after `#` in `Makefile:1`, so all three files match.

PR #4, opened 2025-04-10
Add linting

stuart-mclaren-hpe:
This uses the newer (v2) golangci-lint

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,0 +1,23 @@
+# Copyright 2025 Hewlett Packard Enterprise Development LP
+
+name: Lint
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        go: [ '1.24' ]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go }}
+      - run: go version
+      - run: make linter
+      - run: make lint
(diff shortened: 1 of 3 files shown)


.golangci.yml
```diff
@@ -0,0 +1,77 @@
...
+    paths:
+      - third_party$
+      - builtin$
+      - examples$
+issues:
+  max-same-issues: 0
+formatters:
+  enable:
+    - gofmt
+    - gofumpt
```
matthew-c-hpe, line 67: Reading the `gofumpt` description seems to suggest that it is a superset of `gofmt`, we shouldn't need to enable `gofmt` also. Unless I'm misinterpreting it.

> Enforce a stricter format than gofmt, while being backwards compatible. That is, gofumpt is happy with a subset of the formats that gofmt is happy with.

https://github.com/mvdan/gofumpt

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,23 @@
...
+  push:
+    branches:
+      - main
+  pull_request:
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        go: [ '1.22.3' ]
```
matthew-c-hpe, line 15: Why not use go 1.24?

stuart-mclaren-hpe: It matches what the build workflow is using.

I'm not against updating in the future, but I'm happy to roll with this for now, unless there is some practical reason not to?

matthew-c-hpe: There may be some edge cases of conventions adopted with more recent releases of go that we'd be as well to cover by using 1.24 from the beginning.

e.g. as of 1.24 errors should use const format strings:
https://github.com/google/fscrypt/issues/422

PR #14, opened 2025-04-11
Add initial acceptance test

stuart-mclaren-hpe:
Also add "test" CI workflow to run it.


.github/workflows/test.yaml
```diff
@@ -0,0 +1,22 @@
...
+      - main
+  pull_request:
+jobs:
+  test:
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        go: [ '1.24' ]
+    steps:
+      - uses: actions/checkout@v2
```
matthew-c-hpe, line 17: I think you might need double check the versions on these, checkout action is at version 4 now.

go.mod
```diff
@@ -1,32 +1,64 @@
 module github.com/HPE/terraform-provider-hpe
 
-go 1.22.3
+go 1.23.0
```
jasonquigleyhpecom, line 3: It may be worth, at this early stage, bumping the Go version to latest. What do you think?

stuart-mclaren-hpe: Yes. In a follow on pr.

PR #15, opened 2025-04-15
Remove singleton

stuart-mclaren-hpe:
The singleton (`cf`) in clientfactory was problematic from a testing
point of view (we could not run tests in parallel).

This patch removes it by adding a private `model.Submodel` to the SubProvider.

This will allow us to update code to pass a per-SubProvider instance client factory
to resources via the SubProvider's GetResources/GetDataSources calls. (Code in a follow on patch.)

We also update the SubProvider interface to add `contexts` (so they can be used
for logging if desired).


go.mod
```diff
@@ -1,32 +1,35 @@
 module github.com/HPE/terraform-provider-hpe
 
-go 1.22.3
+go 1.23.0
```
matthew-c-hpe, line 3: Could we use go 1.24 with toolchain 1.24.2?

stuart-mclaren-hpe: Can we do that in another PR? (Out of scope for this change)

internal/subproviders/morpheus/morpheus.go
```diff
@@ -42,18 +44,39 @@ func (SubProvider) Configure(f func(any)) error {
 	}
 }
 
-func (SubProvider) GetName() string {
+func (SubProvider) GetName(_ context.Context) string {
```
matthew-c-hpe, line 47: I think all these getters should be methods of a pointer to SubProvider.
As it stands, a `*morpheus.SubProvider` wouldn't implement the SubProvider Interface (nor would a morpheus.SubProvider).

stuart-mclaren-hpe: What I had in mind here is that the GetXXX methods are for reading values (so I reckoned it was safer to not provide the ability to modify the subprovider in those calls.)

There is an interface check here:
https://github.com/HPE/terraform-provider-hpe/pull/15/files#diff-8f325493dc55067e4b4dd0baa4ef56fb57d7811f44aca593d03a8f3283677341R17
Showing that the morpheus sub provider satisfies the provider's subprovider interface.

eamonnotoole: I get it now.  I just haven't seen that pattern much.

jasonquigleyhpecom (approved):

internal/subproviders/morpheus/morpheus.go
stuart-mclaren-hpe: Good feedback. I think if I remove state from the implementation of the interface I can make all methods use value receivers. I will look at that in a follow on change.

PR #18, opened 2025-04-22
Add test workflow

stuart-mclaren-hpe:
Will allow running, eg some acceptance tests during CI

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,22 @@
+# Copyright 2025 Hewlett Packard Enterprise Development LP
+
+name: Test
+
+on:
+  push:
+    branches:
+      - main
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yaml
```diff
@@ -0,0 +1,22 @@
...
+  push:
+    branches:
+      - main
+  pull_request:
+jobs:
+  test:
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        go: [ '1.24' ]
```
ZoeySimone, line 15: Would be good to bump the `go.mod` version to `1.24` too

stuart-mclaren-hpe: Yup. PR coming up

PR #69, opened 2025-06-03
user resource: write only password

stuart-mclaren-hpe:
Switch to a write only password. This `password_wo` is not
persisted in the state file.

We also introduce `password_wo_version` which can be used (later) in Update()
to determine if the password has been changed. (This is required because
terraform cannot determine if the write-only password has changed by
examining the state.)

### DIFF
--- a/internal/subproviders/morpheus/resources/user/resource.go
+++ b/internal/subproviders/morpheus/resources/user/resource.go
@@ -8,18 +8,16 @@ import (
 	"net/http"
 	"strconv"
 
+	"github.com/HewlettPackard/hpe-morpheus-go-sdk/sdk"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/types"
 
 	"github.com/HPE/terraform-provider-hpe/internal/subproviders/morpheus/configure"
 	"github.com/HPE/terraform-provider-hpe/internal/subproviders/morpheus/convert"
 	"github.com/HPE/terraform-provider-hpe/internal/subproviders/morpheus/errors"
-	"github.com/HewlettPackard/hpe-morpheus-go-sdk/sdk"
-
-	"github.com/hashicorp/terraform-plugin-framework/attr"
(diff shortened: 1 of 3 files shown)


internal/subproviders/morpheus/resources/user/resource_test.go
```diff
@@ -84,11 +84,20 @@ provider "hpe" {
...
+	resourceConfig := `
 resource "hpe_morpheus_user" "foo" {
 	username = "testacc-TestAccMorpheusUserRequiredAttrsOk"
 	email = "foo@hpe.com"
-	password = "Secret123!"
+	password_wo = "Secret123!"
+	role_ids = [3]
+}
+`
+	resourceConfigWo := `
```
matthew-c-hpe, line 96: Naming: Shouldn't this be `resourceConfig` and the above one be `resourceConfigWo`?

This one doesn't have the password write only field.

internal/subproviders/morpheus/resources/user/resource.go
```diff
@@ -235,6 +242,10 @@ func (r *Resource) Read(
 		return
 	}
 
+	if diags.HasError() {
```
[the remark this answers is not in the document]
stuart-mclaren-hpe: Yes, thanks

internal/subproviders/morpheus/resources/user/resource.go
```diff
@@ -7,19 +7,19 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
-	"strings"
 
 	"github.com/hashicorp/terraform-plugin-framework/path"
```
jasonquigleyhpecom, line 11: The imports are all mixed up. std -> 3rd party -> our packages -> packages in this repo

internal/subproviders/morpheus/resources/user/resource_test.go
```diff
@@ -9,9 +9,7 @@ import (
 
 	"github.com/HPE/terraform-provider-hpe/internal/provider"
 	"github.com/HPE/terraform-provider-hpe/internal/subproviders/morpheus"
-
```
jasonquigleyhpecom, line 12: Again, import order

internal/subproviders/morpheus/resources/user/resource.go
stuart-mclaren-hpe: We should automate these checks for ordering. (Can save reviewer bandwidth for other things.)

PR #86, opened 2025-06-10
user resource: add update tests

stuart-mclaren-hpe:
We hardcode each config to make the test logic simpler.
(Make bugs in the tests themselves less likely.)

We test cases where fields which are currently
set to a string value are set back to `null`.

We test the logic of `password_wo` + `password_wo_version`

### DIFF
--- a/internal/subproviders/morpheus/resources/user/resource_test.go
+++ b/internal/subproviders/morpheus/resources/user/resource_test.go
@@ -69,6 +69,7 @@ func TestAccMorpheusUserRequiredAttrsOk(t *testing.T) {
 		t.Skip("Skipping slow test in short mode")
 	}
 
+	// nolint: goconst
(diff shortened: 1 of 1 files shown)


internal/subproviders/morpheus/resources/user/resource_test.go
```diff
@@ -189,6 +190,441 @@ resource "hpe_morpheus_user" "foo" {
 	})
 }
 
+func TestAccMorpheusUserUpdateOk(t *testing.T) {
+	if testing.Short() {
+		t.Skip("Skipping slow test in short mode")
+	}
+
+	// nolint: goconst
```
jasonquigleyhpecom, line 198: not needed here

jasonquigleyhpecom (approved):

PR #119, opened 2025-07-02
Add ValueToAny convert function

stuart-mclaren-hpe:
This is the inverse of the previously added AnyToValue.

In (simplified) pseudocode:

```
x := AnyToValue(X)
Y := ValueToAny(x)
Y == X is true
```

It should be possible, for example to use ValueToAny when using a DynamicAttribute, to create the contents of a suitable POST request.

Similarly, it should be possible to use AnyToValue when using a DynamicAttribute, to map the contents of
a GET's json body to suitable form for state file output.


internal/subproviders/morpheus/convert/valuetoany.go
```diff
@@ -0,0 +1,145 @@
...
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+)
+
+// ValueToAny converts a Terraform Plugin Framework Value into a Go type.
+// This function handles null values, primitive types, and complex types
+// like objects, lists, sets, tuples and maps.
+func ValueToAny(_ context.Context, v attr.Value) (any, error) {
```
matthew-c-hpe, line 16: Any reason you're using a context if it's not being used in this function?

stuart-mclaren-hpe: You need a context if you want to later add, eg tflog.Debug.
We may get corner cases where we need to add debug/trace.
I wanted to add the ability to do that without having people update their calls to the function.
ValueToAny is probably the function most people will use, but the others are available, and
I didn't add contexts there.
I should probably add contexts to all, or make those private for now.

matthew-c-hpe (approved):

matthew-c-hpe (approved):
LGTM

PR #122, opened 2025-07-02
Add tests for ValueToAny

stuart-mclaren-hpe:
Test null cases, lists, sets, object, tuples, primitives.

### DIFF
--- a/internal/subproviders/morpheus/convert/valuetoany_test.go
+++ b/internal/subproviders/morpheus/convert/valuetoany_test.go
@@ -1,0 +1,526 @@
+package convert
+
+import (
+	"context"
+	"math/big"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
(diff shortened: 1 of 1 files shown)


internal/subproviders/morpheus/convert/valuetoany_test.go
```diff
@@ -0,0 +1,526 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestValueToAnyObject(t *testing.T) {
+	t.Parallel()
+
+	testCases := map[string]struct {
+		input    attr.Value
+		expected any
+		wantErr  bool
```
matthew-c-hpe, line 20: Would an `expectErr error` be more useful here if tests were to be added for failure path on complex conversions with nested objects?

stuart-mclaren-hpe: Potentially. I've no objection if someone wants to add that later.

matthew-c-hpe (approved):
LGTM

PR #130, opened 2025-07-04
Use resources for group and cloud data source tests

jasonquigleyhpecom:
This PR remove the old group and cloud test helpers.

- The group data source tests now use the new group resource.
- The cloud data source tests now use the legacy provider cloud resource.


matthew-c-hpe (approved):
LGTM

PR #159, opened 2025-08-06
Refactor permissions block to use Terraform Plugin Framework types

matthew-c-hpe:
This PR switches to using Terraform plugin framework types for the Role permissions block.

Key changes in this PR:
- Generated schema has been updated.
- Permissions block and all permissions it contains are `optional` and act as overrides. We only track them in state if the user has set them in the state file.
- The subset algorithm has been changed. We now use a simpler method involving `slices.IndexFunc` to determine whether permissions in state are still a valid subset of those in API state.
- Lots of boilerplate additions to handle conversions between generated SDK structs and Terraform Plugin Framework types.
- Added an implementation for the `ValidateConfig` method for the Role resource. This prevents a user from setting cloud permissions on a user role and group permissions on an account role.


internal/subproviders/morpheus/resources/role/resource.go
```diff
@@ -74,6 +54,454 @@ func (r *Resource) Schema(
 	resp.Schema = RoleResourceSchema(ctx)
 }
 
+// This function breaks out the logic of reading permissions from API response to store to state.
+func populateGetRoleAsStatePermissions(ctx context.Context, r *sdk.GetRole200Response) (PermissionsValue, diag.Diagnostics) {
+
+	var features []FeaturePermissionsValue
+	for _, v := range r.FeaturePermissions {
+		features = append(features, FeaturePermissionsValue{
+			Code:        types.StringValue(v.GetCode()),
```
stuart-mclaren-hpe, line 63: Can any of these be null?

eg Is there a case where Code (or others) should be of type `type.StringNull` 
Elsewhere (eg group resource) we do

```
convert.StrToType(g.Group.Code.Get())
```

internal/subproviders/morpheus/resources/role/resource.go
```diff
@@ -74,6 +54,454 @@ func (r *Resource) Schema(
...
+		diags = plan.Permissions.CatalogItemTypePermissions.ElementsAs(ctx, &catalogItemTypePermissions, false)
+		if diags.HasError() {
+			return diags
+		}
+
+		var addRoleCatalogItemTypePermissions []sdk.AddRolesRequestRoleCatalogItemTypePermissionsInner
+		for _, v := range catalogItemTypePermissions {
+			addRoleCatalogItemTypePermissions = append(addRoleCatalogItemTypePermissions, sdk.AddRolesRequestRoleCatalogItemTypePermissionsInner{
+				Access: v.Access.ValueString(),
+				Id:     int32(v.Id.ValueInt64()),
```
stuart-mclaren-hpe, line 350: why cast int64 to int32?

internal/subproviders/morpheus/resources/role/resource.go
```diff
@@ -236,15 +644,119 @@ func (r *Resource) Create(
...
+	}
+
+	// If the user provided a config with feature permissions as part of the create,
+	// then set the feature permissions to what was in the plan (optional).
 	if !plan.Permissions.IsNull() && !plan.Permissions.IsUnknown() {
-		state.Permissions = plan.Permissions
+
+		// Only feature permissions requires this more complicated create logic.
+		// This is because if the user sets feature permissions, we can only store to state
+		// the number of feature permissions that were set by the user.
```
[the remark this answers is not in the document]
matthew-c-hpe: Yes, I'll update for clarity

internal/subproviders/morpheus/resources/role/resource.go
matthew-c-hpe: Some of them can be null. I believe tasks can have a null code. I can change it to handle the null case for those where it applies.

internal/subproviders/morpheus/resources/role/resource.go
```diff
@@ -284,46 +796,141 @@ func (r *Resource) Read(
...
-		if err != nil {
-			resp.Diagnostics.Append(pdiags...)
+		// Edge case: Not a subset if there are no feature permissions in Terraform state,
+		// but there are feature permissions in API state.
+		if len(apiStateFeaturePermissions) > 0 && len(stateFeaturePermissions) == 0 {
 			resp.Diagnostics.AddError(
 				"read role resource",
-				fmt.Sprintf("role %d: failed to unmarshal permissions from api; permissions: %s",
-					id, apiPermissionStr),
+				fmt.Sprintf("role %d: feature permissions not a subset: 0 feature permissions in state, 1 or more feature permissions in API", id),
```
stuart-mclaren-hpe, line 880: These read a little like developer facing errors rather than user facing errors

internal/subproviders/morpheus/resources/role/resource.go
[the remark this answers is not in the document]
matthew-c-hpe: done

internal/subproviders/morpheus/resources/role/resource.go
```diff
@@ -236,15 +644,119 @@ func (r *Resource) Create(
...
+
+			var apiStateFeaturePermissions []FeaturePermissionsValue
+			diags = apiState.Permissions.FeaturePermissions.ElementsAs(ctx, &apiStateFeaturePermissions, false)
+			if diags.HasError() {
+				resp.Diagnostics.Append(diags...)
+
+				return
+			}
+
+			planFeaturePermissionsWithComputed := planFeaturePermissions
```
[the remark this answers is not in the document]
matthew-c-hpe: I assumed it was deep copying it so we'll do as you say and just modify the planFeaturePermissions variable, it's the same result.

internal/subproviders/morpheus/resources/role/resource.go
ZoeySimone: It'd probably be a good idea to update this in the openapi spec

PR #183, opened 2025-08-14
network resource: add update

stuart-mclaren-hpe:
For the Dynamic Attribute `config` block, we assume that changes to
the block are HTTP `PUT`-able, and thus supported by update.

The `config` block entries cannot be fetched by HTTP GET, so we assume
that a successful PUT (200) will have updated the values successfully.

This may not be true for all network types, but it seems a pragmatic first pass for
network support.

### DIFF
--- a/internal/subproviders/morpheus/resources/network/update.go
+++ b/internal/subproviders/morpheus/resources/network/update.go
@@ -28,16 +28,6 @@ func (r *Resource) Update(
 		return
 	}
 
-	client, err := r.NewClient(ctx)
-	if err != nil {
-		resp.Diagnostics.AddError(
-			"update network resource",
-			"failed to create client: "+err.Error(),
-		)
-
-		return
-	}
-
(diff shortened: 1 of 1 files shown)


internal/subproviders/morpheus/resources/network/update.go
```diff
@@ -0,0 +1,256 @@
...
+) {
+	var plan, state NetworkModel
+
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	client, err := r.NewClient(ctx)
```
jasonquigleyhpecom, line 31: Maybe you could move this closer to where it's used?

stuart-mclaren-hpe: sure

PR #192, opened 2025-08-19
dynamic attribute validation

stuart-mclaren-hpe:
Add basic validation for `config` type dynamic attributes

This cannot validate much, but it will ensure `{}`s, eg:

```
config = {
 foo = "bar"
}
```

and allow raising an error during plan for

```
config = "foo"
```

or

```
config = 1
```

Note that this is not tied to the name `config`, dynamic attributes with any name should work also:

```
foo = {} # ok
foo = 123  # errors
```


matthew-c-hpe (approved):
This looks good to me.

PR #200, opened 2025-08-20
network resource: add update tests

stuart-mclaren-hpe:
Add initial set of update tests for networks.

We check that individual modified attributes trigger a plan diff.
We check that all modifiable fields get changed on update.
We don't run `apply` for each individual modifiable attribute.


matthew-c-hpe (approved):
LGTM.

PR #218, opened 2025-08-25
network resource: add documentation note

stuart-mclaren-hpe:
Add note about `lifecycle` block


internal/subproviders/morpheus/resources/network/extended_description.md
matthew-c-hpe, line 1: Should this live in the same directory as the other documentation?

When approving https://github.com/HPE/terraform-provider-hpe/pull/216, I didn't put together that the embedded description may end up in the same directory as the resource.

PR #223, opened 2025-08-27
Update experimental Role documentation

matthew-c-hpe:
Updates the Role documentation in `templates-experimintal` to capture information about Roles in line with the first release of the provider.


eamonnotoole (dismissed):
Really good documentation @matthew-c-hpe .  Well done.

PR #225, opened 2025-08-27
Service Plan resource (CRD only)

naoise-osullivan-hpe:
- Base implementation of service plan resource.
- Update method omitted from resource.go.
- required and all attributes tested.
- Currently in experimental

### DIFF
--- a/internal/subproviders/morpheus/resources/serviceplan/resource.go
+++ b/internal/subproviders/morpheus/resources/serviceplan/resource.go
@@ -211,48 +211,37 @@ func setConfigInCreate(
 			ranges.MinMemory = plan.ConfigRanges.MinMemory.ValueInt64Pointer()
 		}
 		if !plan.ConfigRanges.MaxMemory.IsNull() {
-			maxMemory := plan.ConfigRanges.MaxMemory.ValueInt64Pointer()
-			ranges.MaxMemory = maxMemory
+			ranges.MaxMemory = plan.ConfigRanges.MaxMemory.ValueInt64Pointer()
 		}
 		if !plan.ConfigRanges.MinStorage.IsNull() {
-			minStorage := plan.ConfigRanges.MinStorage.ValueInt64Pointer()
-			ranges.MinStorage = minStorage
+			ranges.MinStorage = plan.ConfigRanges.MinStorage.ValueInt64Pointer()
 		}
 		if !plan.ConfigRanges.MaxStorage.IsNull() {
-			maxStorage := plan.ConfigRanges.MaxStorage.ValueInt64Pointer()
-			ranges.MaxStorage = maxStorage
+			ranges.MaxStorage = plan.ConfigRanges.MaxStorage.ValueInt64Pointer()
 		}
 		if !plan.ConfigRanges.MinCores.IsNull() {
-			minCores := plan.ConfigRanges.MinCores.ValueInt64Pointer()
-			ranges.MinCores = minCores
+			ranges.MinCores = plan.ConfigRanges.MinCores.ValueInt64Pointer()
 		}
 		if !plan.ConfigRanges.MaxCores.IsNull() {
-			maxCores := plan.ConfigRanges.MaxCores.ValueInt64Pointer()
-			ranges.MaxCores = maxCores
+			ranges.MaxCores = plan.ConfigRanges.MaxCores.ValueInt64Pointer()
 		}
 		if !plan.ConfigRanges.MinCoresPerSocket.IsNull() {
-			minCoresPerSocket := plan.ConfigRanges.MinCoresPerSocket.ValueInt64Pointer()
-			ranges.MinCoresPerSocket = minCoresPerSocket
+			ranges.MinCoresPerSocket = plan.ConfigRanges.MinCoresPerSocket.ValueInt64Pointer()
 		}
 		if !plan.ConfigRanges.MaxCoresPerSocket.IsNull() {
-			maxCoresPerSocket := plan.ConfigRanges.MaxCoresPerSocket.ValueInt64Pointer()
-			ranges.MaxCoresPerSocket = maxCoresPerSocket
+			ranges.MaxCoresPerSocket = plan.ConfigRanges.MaxCoresPerSocket.ValueInt64Pointer()
 		}
 		if !plan.ConfigRanges.MinPerDiskSize.IsNull() {
-			minPerDiskSize := plan.ConfigRanges.MinPerDiskSize.ValueInt64Pointer()
-			ranges.MinPerDiskSize = minPerDiskSize
+			ranges.MinPerDiskSize = plan.ConfigRanges.MinPerDiskSize.ValueInt64Pointer()
 		}
 		if !plan.ConfigRanges.MaxPerDiskSize.IsNull() {
-			maxPerDiskSize := plan.ConfigRanges.MaxPerDiskSize.ValueInt64Pointer()
-			ranges.MaxPerDiskSize = maxPerDiskSize
+			ranges.MaxPerDiskSize = plan.ConfigRanges.MaxPerDiskSize.ValueInt64Pointer()
 		}
 		if !plan.ConfigRanges.MinSockets.IsNull() {
-			minSockets := plan.ConfigRanges.MinSockets.ValueInt64Pointer()
-			ranges.MinSockets = minSockets
+			ranges.MinSockets = plan.ConfigRanges.MinSockets.ValueInt64Pointer()
 		}
 		if !plan.ConfigRanges.MaxSockets.IsNull() {
-			maxSockets := plan.ConfigRanges.MaxSockets.ValueInt64Pointer()
-			ranges.MaxSockets = maxSockets
+			ranges.MaxSockets = plan.ConfigRanges.MaxSockets.ValueInt64Pointer()
 		}
 
 		config.Ranges = ranges


go.sum
```diff
@@ -1,7 +1,5 @@
 dario.cat/mergo v1.0.0 h1:AGCNq9Evsj31mOgNPcLyXc+4PNABt905YmuqPYYpBWk=
 dario.cat/mergo v1.0.0/go.mod h1:uNxQE+84aUszobStD9th8a29P2fMDhsBdgRYvZOxGmk=
-github.com/HewlettPackard/hpe-morpheus-go-sdk v0.0.0-20250814115258-5d820c745ce9 h1:O83aE8ujL1IVEdSi/fyArGhBkEEmHwbUJL6QQJ6um90=
```
matthew-c-hpe, line 3: Likewise we can remove the old checksum of the previous sdk version whenever we update the SDK next so just undo this change.

internal/subproviders/morpheus/resources/serviceplan/resource.go
```diff
@@ -0,0 +1,637 @@
...
+	if !config.ConfigRanges.IsNull() {
+
+		// ValueBool returns false if value was set to false, or if state is null or unknown
+		if !config.CustomMaxMemory.ValueBool() {
+
+			if !config.ConfigRanges.MinMemory.IsNull() {
+				resp.Diagnostics.AddAttributeError(
+					path.Root("config_ranges.min_memory"),
+					"Conflicting attributes in configuration",
+					`min_memory set when custom_memory is not`+
```
matthew-c-hpe, line 548: For all these messages, you'll want to add a `\n` before the end of the first line and break them up into two sentences. Maybe just change `is not` on the first lines to `has not been set` for clarity.

Currently the messages will display as `min_memory set when custom_memory is notset custom_memory to true to add a min_memory value.`

The + operator on strings in Go appends one string immediately to the end of another.

The validation logic looks fine though.

internal/subproviders/morpheus/resources/serviceplan/resource.go
```diff
@@ -0,0 +1,637 @@
...
+		)
+		return state, diags
+	}
+
+	pricesetIDValues := []attr.Value{}
+	for _, ps := range sp.ServicePlan.PriceSets {
+		pricesetIDValues = append(pricesetIDValues, convert.Int64ToType(ps.Id))
+	}
+
+	pricesetIDSet, d := types.SetValue(types.Int64Type, pricesetIDValues)
```
matthew-c-hpe, line 81: No need to create a temporary variable `d`, we can reuse `diags` here.

internal/subproviders/morpheus/resources/serviceplan/resource.go
```diff
@@ -0,0 +1,637 @@
...
+
+	// config is defaulted to empty map by api
+	apiConfig := sp.ServicePlan.Config
+
+	// config top level fields moved to service plan top level
+	state.StorageSizeType = convert.StrToType(apiConfig.StorageSizeType.Get())
+	state.MemorySizeType = convert.StrToType(apiConfig.MemorySizeType.Get())
+
+	if apiConfig.Ranges != nil {
+		configRanges := ConfigRangesValue{}
```
matthew-c-hpe, line 95: Use the `NewConfigRangesValue` constructor here. You'll get some additional validation and you won't have to explicitly set the state to ValueKnown.
See the `populateGetRoleAsStatePermissions()` function in Role `resource.go` to see how to use it.

internal/subproviders/morpheus/resources/serviceplan/schema_gen.go
```diff
@@ -0,0 +1,1175 @@
...
+				CustomType: ConfigRangesType{
+					ObjectType: types.ObjectType{
+						AttrTypes: ConfigRangesValue{}.AttributeTypes(ctx),
+					},
+				},
+				Optional:            true,
+				Description:         "The min and max ranges that instances using this service plan must conform to.",
+				MarkdownDescription: "The min and max ranges that instances using this service plan must conform to.",
+			},
+			"cores_per_socket": schema.Int64Attribute{
```
matthew-c-hpe, line 112: I think this should use a `UseStateForUnknown()` plan modifier

internal/subproviders/morpheus/resources/serviceplan/schema_gen.go
```diff
@@ -0,0 +1,1175 @@
...
+					),
+				},
+				Default: stringdefault.StaticString("mb"),
+			},
+			"name": schema.StringAttribute{
+				Required:            true,
+				Description:         "Service plan name",
+				MarkdownDescription: "Service plan name",
+			},
+			"priceset_ids": schema.SetAttribute{
```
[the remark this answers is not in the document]
matthew-c-hpe: Finally, we may want to double check if this should be named `price_set_ids`. Maybe we can discuss this in standup.

internal/subproviders/morpheus/resources/serviceplan/schema_gen.go
```diff
@@ -0,0 +1,1175 @@
...
+				Description:         "Max number of cores",
+				MarkdownDescription: "Max number of cores",
+			},
+			"max_cpu": schema.Int64Attribute{
+				Optional:            true,
+				Computed:            true,
+				Description:         "Max number of CPUs",
+				MarkdownDescription: "Max number of CPUs",
+			},
+			"max_disks": schema.Int64Attribute{
```
matthew-c-hpe, line 171: I think this should use a UseStateForUnknown() plan modifier

internal/subproviders/morpheus/resources/serviceplan/schema_gen.go
```diff
@@ -0,0 +1,1175 @@
...
+					int64planmodifier.UseStateForUnknown(),
+				},
+			},
+			"max_cores": schema.Int64Attribute{
+				Optional:            true,
+				Computed:            true,
+				Description:         "Max number of cores",
+				MarkdownDescription: "Max number of cores",
+			},
+			"max_cpu": schema.Int64Attribute{
```
matthew-c-hpe, line 165: I think this should use a UseStateForUnknown() plan modifier

internal/subproviders/morpheus/resources/serviceplan/schema_gen.go
```diff
@@ -0,0 +1,1175 @@
...
+				MarkdownDescription: "Number of cores per CPU",
+			},
+			"custom_cores": schema.BoolAttribute{
+				Optional:            true,
+				Computed:            true,
+				Description:         "Can be used to enable / disable customizable cores",
+				MarkdownDescription: "Can be used to enable / disable customizable cores",
+				Default:             booldefault.StaticBool(false),
+			},
+			"custom_cpu": schema.BoolAttribute{
```
matthew-c-hpe, line 125: I think this should use a UseStateForUnknown() plan modifier

internal/subproviders/morpheus/resources/serviceplan/schema_gen.go
```diff
@@ -0,0 +1,1175 @@
...
+				ElementType: types.Int64Type,
+				Optional:    true,
+				Computed:    true,
+			},
+			"provision_type_code": schema.StringAttribute{
+				Required:            true,
+				Description:         "The provision type code of the Morpheus service plan",
+				MarkdownDescription: "The provision type code of the Morpheus service plan",
+			},
+			"sort_order": schema.Int64Attribute{
```
matthew-c-hpe, line 215: I think this should use a UseStateForUnknown() plan modifier

internal/subproviders/morpheus/resources/serviceplan/resource.go
```diff
@@ -0,0 +1,637 @@
...
+// helper function to set provisionType in addServicePlan struct
+func setProvisionTypeInCreate(
+	ctx context.Context,
+	client *sdk.APIClient,
+	plan *ServicePlanModel,
+	addServicePlan *sdk.AddServicePlansRequestServicePlan,
+) error {
+	provisionTypeCode := plan.ProvisionTypeCode.ValueString()
+
+	pTypes, hresp, err := client.ProvisioningAPI.ListProvisionTypes(ctx).Code(
```
matthew-c-hpe, line 149: Asking this because I'm unfamiliar with this part of the API.

Do we need to do this validation here?
Will the API not perform validation if the provision type code isn't found or if multiples are found as part of the logic to handle a POST to the service plans endpoint?

internal/subproviders/morpheus/resources/serviceplan/resource.go
```diff
@@ -0,0 +1,637 @@
...
+		)
+
+		return
+	}
+
+	// optional
+	if !plan.PricesetIds.IsNull() && !plan.PricesetIds.IsUnknown() {
+		var pricesetIds []int64
+		diags := plan.PricesetIds.ElementsAs(ctx, &pricesetIds, false)
+		resp.Diagnostics.Append(diags...)
```
matthew-c-hpe, line 305: for consistency with the rest of the provider, could you check for an error on diags first, THEN append to resp in the if statement?

internal/subproviders/morpheus/resources/serviceplan/resource.go
naoise-osullivan-hpe: --- api expects json object with provisionType id, we need to get the id from the code the user has given. n. 
        provisionType := sdk.AddClusterLayoutsRequestLayoutProvisionType{}
	provisionType.Id = *pTypeID
	addServicePlan.ProvisionType = provisionType

internal/subproviders/morpheus/resources/serviceplan/resource.go
```diff
@@ -0,0 +1,637 @@
...
+	id := *servicePlan.Id
+	plan.Id = types.Int64Value(id)
+
+	// write id as soon as possible
+	resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	state, pdiags := getServicePlanAsState(ctx, id, client)
```
matthew-c-hpe, line 409: no need for this to be called pdiags, `diags` is fine.

internal/subproviders/morpheus/resources/serviceplan/resource.go
```diff
@@ -0,0 +1,637 @@
...
+		resp.Diagnostics.AddError(
+			"read service plan resource",
+			"new client call failed with "+err.Error(),
+		)
+
+		return
+	}
+
+	id := plan.Id.ValueInt64()
+	state, pdiags := getServicePlanAsState(ctx, id, client)
```
matthew-c-hpe, line 457: rename pdiags to diags

internal/subproviders/morpheus/resources/serviceplan/resource.go
```diff
@@ -0,0 +1,637 @@
...
+	if pdiags.HasError() {
+		resp.Diagnostics.Append(pdiags...)
+		resp.Diagnostics.AddError(
```
matthew-c-hpe, line 460: Redundant append to diags, you already have them from `getServicePlanAsState`

internal/subproviders/morpheus/resources/serviceplan/resource_test.go
```diff
@@ -0,0 +1,334 @@
...
+
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config:             providerConfig + resourceConfig,
+				ExpectNonEmptyPlan: false,
+				Check:              checkFn,
+				PlanOnly:           false,
+			},
```
matthew-c-hpe, line 100: Add a step to ensure that there's no diff on a plan only after apply. This will show you whether you may need some of your attributes to use `UseStateForUnknown()` plan modifier.

matthew-c-hpe (changes requested):
Really good work, I've left a lot of comments so don't be discouraged, but you've done a lot of work here so it's natural there'll be some stuff to work through.

internal/subproviders/morpheus/resources/serviceplan/resource.go
```diff
@@ -0,0 +1,637 @@
...
+
+	// required
+	addServicePlan.SetName(name)
+	addServicePlan.SetCode(plan.Code.ValueString())
+	addServicePlan.SetMaxMemory(plan.MaxMemory.ValueInt64())
+	addServicePlan.SetMaxStorage(plan.MaxStorage.ValueInt64())
+
+	err = setProvisionTypeInCreate(ctx, client, &plan, addServicePlan)
+	if err != nil {
+		// setProvisionTypeInCreate checks for cert err
```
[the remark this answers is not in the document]
matthew-c-hpe: Ok, I think remove the comment because we're not really checking for a cert error.

internal/subproviders/morpheus/resources/serviceplan/resource.go
```diff
@@ -0,0 +1,631 @@
...
+		resp.Diagnostics.AddError(
+			"create service plan resource",
+			"set provision type POST failed : "+err.Error(),
```
matthew-c-hpe, line 295: just update this to clarify the context of POST - we're not sending a POST request until later on in this function, so the error message is confusing compared to the established pattern elsewhere in the provider for POST requests.

maybe something like `set provision type ID from code failed`?
or `set provision type ID in POST request failed`?

internal/subproviders/morpheus/resources/serviceplan/resource.go
```diff
@@ -0,0 +1,631 @@
...
+	if !config.ConfigRanges.IsNull() {
+
+		// ValueBool returns false if value was set to false, or if state is null or unknown
+		if !config.CustomMaxMemory.ValueBool() {
+
+			if !config.ConfigRanges.MinMemory.IsNull() {
+				resp.Diagnostics.AddAttributeError(
+					path.Root("config_ranges.min_memory"),
+					"Conflicting attributes in configuration",
+					`min_memory set when custom_memory has not been `+
```
matthew-c-hpe, line 542: these still need to be broken up into two sentences, they render like this:

```
│ Error: Conflicting attributes in configuration
│
│   with hpe_morpheus_service_plan.example_required,
│   on main.tf line 26, in resource "hpe_morpheus_service_plan" "example_required":
│   26: resource "hpe_morpheus_service_plan" "example_required" {
│
│ max_memory set when custom_memory has not been set custom_memory to true to add a max_memory value.
```

go.sum
matthew-c-hpe: think you only undid one of the lines instead of two, go.sum is still in the git diff.

internal/subproviders/morpheus/resources/serviceplan/resource.go
```diff
@@ -0,0 +1,631 @@
...
+
+		return
+	}
+
+	id := plan.Id.ValueInt64()
+	state, diags := getServicePlanAsState(ctx, id, client)
+	if diags.HasError() {
+		resp.Diagnostics.AddError(
+			"read service plan resource",
+			fmt.Sprintf("service plan %d: failed to read from api", id),
```
matthew-c-hpe, line 456: You can just append `diags` to `resp` instead of using `AddError()`. By using `AddError()` you're losing the error information in the `diags` returned from `getServicePlanAsState`

internal/subproviders/morpheus/resources/serviceplan/resource.go
```diff
@@ -0,0 +1,637 @@
...
+		addVolumes := plan.AddVolumes.ValueBool()
+		addServicePlan.AddVolumes = &addVolumes
+	}
+
+	if !plan.SortOrder.IsNull() {
+		sortOrder := plan.SortOrder.ValueInt64()
+		addServicePlan.SortOrder = &sortOrder
+	}
+
+	if !plan.MemorySizeType.IsNull() || !plan.StorageSizeType.IsNull() ||
```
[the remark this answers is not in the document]
matthew-c-hpe: I think we might be talking about different things

setConfigInCreate will get called if any of those three attributes are set (i.e. not null).

But in setConfigInCreate, the `sdk.AddServicePlansRequestServicePlanConfig` is only getting updated with values from these three fields if they have been set.

So you can be in a situation where `MemorySizeType()` has been set but the others not, so it short circuits the OR statement. And then you check again if `MemorySizeType()` has been set in order to set the `sdk.AddServicePlansRequestServicePlanConfig`'s `MemorySizeType`.
So the OR check on all of these fields here seems redundant.

internal/subproviders/morpheus/resources/serviceplan/schema_gen.go
```diff
@@ -111,9 +112,11 @@ func ServicePlanResourceSchema(ctx context.Context) schema.Schema {
 			},
 			"cores_per_socket": schema.Int64Attribute{
 				Optional:            true,
-				Computed:            true,
```
matthew-c-hpe: How come Computed is removed in this commit?

internal/subproviders/morpheus/resources/serviceplan/resource.go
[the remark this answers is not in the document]
naoise-osullivan-hpe: the reason for checking if at least one of them is not null is to avoid the edge case where i blindly call setConfigInCreate with all them null. In this case then i would be setting the request.config to a value a config {} struct, which as far as the api is concerned is not null, as it just stores config as a string. There is no real consequence doing this though so i can change it either way?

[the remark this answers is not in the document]
matthew-c-hpe: Fair enough. But I don't think it's possible to have an empty config that leaves the Terraform provider and gets sent to the API considering storage size type and memory size type both use default static strings (these will get set in the plan for create even if they're set to null in your config).

So I think it's safe to remove the OR checks. If stuff breaks down the line we can come back to it.

PR #237, opened 2025-09-01
Add read operations for the cloud resource

jasonquigleyhpecom:
This PR adds the read operations for the Morpheus cloud resource.

PR #238, opened 2025-09-01
Add create operation to HVM cloud

jasonquigleyhpecom:
This PR adds the create operation for the Morpheus HVM cloud.

This should be landed after #237


internal/subproviders/morpheus/resources/cloud/cloud_create.go
```diff
@@ -0,0 +1,205 @@
...
+
+	addCloudConfig := sdk.AddCloudsRequestZoneConfig{
+		AddCloudsRequestZoneConfigAnyOf: &sdk.AddCloudsRequestZoneConfigAnyOf{},
+	}
+
+	cloudTypeCode := "standard"
+
+	switch {
+	case !plan.ConfigHvm.IsNull():
+		cloudTypeCode = "standard"
```
stuart-mclaren-hpe, line 48: already set?

jasonquigleyhpecom: Yes, good catch

PR #239, opened 2025-09-01
network resource: make resource permissions computed

stuart-mclaren-hpe:
Handle the logic change around making network resource permissions computed.

Update tests accordingly.

Tweak documentation.

### DIFF
--- a/internal/subproviders/morpheus/resources/network/create_test.go
+++ b/internal/subproviders/morpheus/resources/network/create_test.go
@@ -5,6 +5,8 @@
 package network_test
 
 import (
+	"io/ioutil"
+	"path/filepath"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-testing/config"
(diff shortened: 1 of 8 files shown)


internal/subproviders/morpheus/resources/network/create_test.go
```diff
@@ -27,7 +29,7 @@ func TestAccMorpheusNetworkResourceCreateRequiredAttrsOk(t *testing.T) {
 variable "name" {
   description = "Network name"
   type        = string
-  default     = "terraform-network-minimal"
+  default     = "TestAccMorpheusNetworkResourceCreateRequiredAttrsOk"
```
jasonquigleyhpecom, line 32: Could hardcoded values here cause problems?

What about using `name := acctest.RandomWithPrefix(t.Name())`

stuart-mclaren-hpe: Think it's ok. See uniqueName on line 22 (or is it 24, github is confusing me)

jasonquigleyhpecom (approved):

PR #255, opened 2025-09-02
Commit docs-experimental to repo

matthew-c-hpe:
Commits the contents of `docs-experimental`, i.e. the generated directory and contents following a run of `make docs-experimental`, to the repo.

This is done so documentation can be handed over to QA without the need for QA to generate experimental documentation.

This directory should be kept up-to-date as more documentation is added (just run `make docs-experimental` when committing documentation).


matthew-c-hpe:
Closing as we rolled back our decision to hand over items in `experimental` to QA. 

QA handovers will be performed on resources moved out of `experimental` so that QA can evaluate it from the perspective of a user (i.e. the resource/data source is considered ready for use by users when moved out of experimental).

PR #256, opened 2025-09-02
Add instance resource example and test

ZoeySimone:
Adds an example of an HVM instance being provisioned through terraform.

Adds a drive by fix for the instance dynamic config caught by the testing.

### DIFF
--- a/internal/subproviders/morpheus/resources/instance/example.tf.tmpl
+++ b/internal/subproviders/morpheus/resources/instance/example.tf.tmpl
@@ -1,0 +1,52 @@
+data "hpe_morpheus_cloud" "vme_cloud" {
+  name = "HPE Alletra VME" 
+}
+
+data "hpe_morpheus_service_plan" "vme_512mb" {
+    name                = "1 CPU, 1GB Memory"
+    provision_type_code = "kvm"
+}
+
+resource "hpe_morpheus_instance" "example" {
+  name               = "{{.Name}}"
+  cloud_id           = data.hpe_morpheus_cloud.vme_cloud.id   # HPE Alletra VME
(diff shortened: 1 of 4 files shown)


internal/subproviders/morpheus/resources/instance/instance.go
```diff
@@ -162,6 +162,10 @@ func getInstanceAsState(
 	// config
 	state.Config = types.DynamicNull()
 
+	if !plan.Config.IsNull() && !plan.Config.IsUnknown() {
```
stuart-mclaren-hpe, line 165: Not mentioned in commit.
If you're doing things other than docs, should be in seperate pr, or mentioned in commit.

internal/subproviders/morpheus/resources/instance/example.tf.tmpl
```diff
@@ -0,0 +1,52 @@
+data "hpe_morpheus_cloud" "vme_cloud" {
+  name = "HPE Alletra VME"
```
matthew-c-hpe, line 2: Will the example test for this only work on the `feature` environment?
Would it be possible to provision a standard morpheus cloud as a dependency in the `resource_test.go` to provision an instance under?

Looks like the service plan is one of those seeded into the application so that should be ok.

ZoeySimone: I think it would be essentially impossible at the moment to completely separate the instance example from the `feature` system, and I think that's okay for now. We don't yet have some of the resources/datasources implemented once we do we can update the example further. 

I could try adding a `cloud` resource into the example, it shouldn't be difficult, but I am worried that the instance example could become too unwieldy to be useful to users.

PR #294, opened 2025-09-18
Refactor the environment datasource to avoid referencing SDK struct names

RDavisHPE:
Refactoring the environment datasource to avoid returning SDK struct names and instead populate the Terraform model directly within the `getEnvironment` function.

This reduces dependencies on the SDK's internal structures - which may change.
- Changed getEnvironmentByName and getEnvironmentByID to return error only.
- Populate the EnvironmentModel directly in getEnvironmentByName and getEnvironmentByID.

PR #295, opened 2025-09-18
Refactoring CreateEnvironment to avoid passing around SDK structs

RDavisHPE:
Refactored CreateEnvironment to return a pointer to the created TestEnvironment struct instead of the full SDK struct.

This prevents using SDK structs directly - as they may change.
- Added TestEnvironment to get necessary fields.
- Updated CreateEnvironment to return *TestEnvironment.
- Updated datasource tests to use TestEnvironment fields instead of SDK struct methods.


internal/subproviders/morpheus/datasources/environment/datasource_test.go
```diff
@@ -49,11 +49,11 @@ func TestAccMorpheusFindEnvironmentById(t *testing.T) {
 	}
 
 	t.Cleanup(func() {
-		testhelpers.DeleteEnvironment(t, environment.GetId())
+		testhelpers.DeleteEnvironment(t, environment.ID)
```
RDavisHPE, line 52: Not sure about the above lint failure but it seems to go away when I cherrypick [my other environment refactor.](https://github.com/HPE/terraform-provider-hpe/pull/294)

PR #300, opened 2025-09-19
Support outputting a debug env file

ZoeySimone:
Newer versions of `terraform-plugin-framework` support outputting the debug configuration into a file.

This allows us remove the manual steps when we want to run the provider with a debugger.

Update the go.mod with latest version of `terraform-plugin-framework` and update `main.go` to support writing out the debug config file.

### DIFF
--- a/main.go
+++ b/main.go
@@ -28,6 +28,11 @@ func main() {
 	p := provider.New(
 		version,
 		morpheus.New(),
+		// subprovider2.New(),
+		// subprovider3.New(),
+		// .
+		// .
+		// .
 	)
 
 	var opts []tf6server.ServeOpt


main.go
```diff
@@ -23,23 +25,32 @@ func main() {
...
 
-	opts := providerserver.ServeOpts{
-		Address: "registry.terraform.io/HPE/hpe",
-		Debug:   debug,
-	}
-
 	p := provider.New(
 		version,
 		morpheus.New(),
-		// subprovider2.New(),
```
stuart-mclaren-hpe, line 34: It would be good to preserve showing how additional 
subproviders would be added (however it would be done)

PR #319, opened 2025-09-29
Add a StructToDynamic converter

jasonquigleyhpecom:
This PR adds a helper to convert a struct into a dynamic attribute

### DIFF
--- a/internal/subproviders/morpheus/convert/anytovalue.go
+++ b/internal/subproviders/morpheus/convert/anytovalue.go
@@ -4,6 +4,8 @@ package convert
 
 import (
 	"context"
+	"encoding/json"
+	"errors"
 	"fmt"
 	"math/big"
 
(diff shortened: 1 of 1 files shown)


eamonnotoole (dismissed):

eamonnotoole (approved):

internal/subproviders/morpheus/convert/anytovalue.go
```diff
@@ -362,3 +364,104 @@ func TupleToValue(
 
 	return types.TupleValueMust(tupleType.ElemTypes, vl), nil
 }
+
+func StructToDynamic(ctx context.Context, s any) (types.Dynamic, error) {
```
stuart-mclaren-hpe, line 368: anytovalue.go is specifically for converting
from Go type "any" to a terraform value.

This seems to be json conversion, so I think
we should put it in a different file.

PR #322, opened 2025-09-30
Adding User Datasource to Morpheus subprovider

RDavisHPE:
Adding the User datasource to the Morpheus subprovider so that data about users can be queried in Terraform configurations using the `data "hpe_morpheus_user"` block.
- Implemented datasource interface in datasource.go
- Added user datasource to datasources.go
- Defined access objects in Morpheus API spec and updated hpe-morpheus-go-sdk and hpe-morpheus-provider codespec with new OpenAPI version respectively.
- Ran go get github.com/HewlettPackard/hpe-morpheus-go-sdk@main and go mod tidy for SDK bump
- Ran make docs for documentation


internal/subproviders/morpheus/datasources/user/datasource.go
```diff
@@ -0,0 +1,540 @@
...
+
+		return diags
+	}
+
+	// Handle Access complex object
+	if access, accessOk := user.GetAccessOk(); accessOk {
+		buildAppTemplates := func() types.Set {
+			appTemplatesValues := []attr.Value{}
+			for _, in := range access.GetAppTemplates() {
+				appTemplatesValues = append(appTemplatesValues,
```
RDavisHPE: Was curious if I could use convert.ToSetType for these but looking at examples from instance it would seem this involves referencing SDK struct names explicitly (which we would like to avoid in case they change):
```go
tags, d := convert.ToSetType(
	ctx,
	resp.GetInstance().Tags,
	func(
		in sdk.AddInstance200ResponseAllOfOneOfInstanceTagsInner,
	) TagsValue {
		return TagsValue{
			Name:  convert.StrToType(in.Name),
			Value: convert.StrToType(in.Value),
			state: attr.ValueStateKnown,
		}
	},
)
```

matthew-c-hpe:
Could you separate out the sdk version bump into a separate PR?

internal/subproviders/morpheus/datasources/user/schema_gen.go
```diff
@@ -0,0 +1,8249 @@
...
+			"last_name": schema.StringAttribute{
+				Computed: true,
+			},
+			"last_updated": schema.StringAttribute{
+				Computed: true,
+			},
+			"linux_key_pair_id": schema.Int64Attribute{
+				Computed: true,
+			},
+			"linux_password": schema.StringAttribute{
```
matthew-c-hpe: I don't think there's much point in having linux or windows passwords available in the data source given that the API hides them as `"************"` if they're set.

internal/subproviders/morpheus/datasources/user/schema_gen.go
```diff
@@ -0,0 +1,8249 @@
...
+	DateCreated          types.String        `tfsdk:"date_created"`
+	DefaultPersona       DefaultPersonaValue `tfsdk:"default_persona"`
+	DisplayName          types.String        `tfsdk:"display_name"`
+	Email                types.String        `tfsdk:"email"`
+	Enabled              types.Bool          `tfsdk:"enabled"`
+	FirstName            types.String        `tfsdk:"first_name"`
+	Id                   types.Int64         `tfsdk:"id"`
+	IncludeAccess        types.Bool          `tfsdk:"include_access"`
+	IsUsing2fa           types.Bool          `tfsdk:"is_using2fa"`
+	LastLoginDate        types.String        `tfsdk:"last_login_date"`
```
matthew-c-hpe: In a Terraform and Morpheus context, I'm not sure how useful it would be to have the date and time-related properties `LastLoginDate`, `DateCreated`, `LastUpdated` available in the data source. I think we could leave them out.

internal/subproviders/morpheus/datasources/user/schema_gen.go
```diff
@@ -0,0 +1,8249 @@
...
+	return NewAccessValueMust(AccessValue{}.AttributeTypes(ctx), attributes), nil
+}
+
+func (t AccessType) ValueType(ctx context.Context) attr.Value {
+	return AccessValue{}
+}
+
+var _ basetypes.ObjectValuable = AccessValue{}
+
+type AccessValue struct {
```
matthew-c-hpe, line 1030: In the new provider we're trying to adopt the common naming convention of Morpheus resources as per the UI and Documentation, this means that the schema will have to be adjusted if we will include the access attribute in the data source. See the Role resource implementation for guidance

internal/subproviders/morpheus/datasources/user/schema_gen.go
```diff
@@ -0,0 +1,8249 @@
...
+	AppTemplates     basetypes.SetValue `tfsdk:"app_templates"`
```
matthew-c-hpe: AppTemplates should be renamed to Blueprints

internal/subproviders/morpheus/datasources/user/schema_gen.go
```diff
@@ -0,0 +1,8249 @@
...
+	CatalogItemTypes basetypes.SetValue `tfsdk:"catalog_item_types"`
+	Features         basetypes.SetValue `tfsdk:"features"`
+	InstanceTypes    basetypes.SetValue `tfsdk:"instance_types"`
+	Personas         basetypes.SetValue `tfsdk:"personas"`
+	ReportTypes      basetypes.SetValue `tfsdk:"report_types"`
+	Sites            basetypes.SetValue `tfsdk:"sites"`
+	TaskSets         basetypes.SetValue `tfsdk:"task_sets"`
```
matthew-c-hpe: TaskSets should be renamed to Workflows

internal/subproviders/morpheus/datasources/user/schema_gen.go
```diff
@@ -0,0 +1,8249 @@
...
+	Tasks            basetypes.SetValue `tfsdk:"tasks"`
+	VdiPools         basetypes.SetValue `tfsdk:"vdi_pools"`
+	Zones            basetypes.SetValue `tfsdk:"zones"`
```
matthew-c-hpe: Zones should be renamed to Clouds

internal/subproviders/morpheus/datasources/user/schema_gen.go
```diff
@@ -0,0 +1,8249 @@
...
+var _ basetypes.ObjectValuable = AccessValue{}
+
+type AccessValue struct {
+	AppTemplates     basetypes.SetValue `tfsdk:"app_templates"`
+	CatalogItemTypes basetypes.SetValue `tfsdk:"catalog_item_types"`
+	Features         basetypes.SetValue `tfsdk:"features"`
+	InstanceTypes    basetypes.SetValue `tfsdk:"instance_types"`
+	Personas         basetypes.SetValue `tfsdk:"personas"`
+	ReportTypes      basetypes.SetValue `tfsdk:"report_types"`
+	Sites            basetypes.SetValue `tfsdk:"sites"`
```
matthew-c-hpe: Sites should be renamed to Groups.

internal/subproviders/morpheus/datasources/user/datasource.go
```diff
@@ -0,0 +1,540 @@
...
+		},
+	}, roleValues)
+	if roleDiags.HasError() {
+		diags.Append(roleDiags...)
+
+		return diags
+	}
+
+	// Handle Access complex object
+	if access, accessOk := user.GetAccessOk(); accessOk {
```
matthew-c-hpe: Break the building up of the access object out into a separate function

internal/subproviders/morpheus/datasources/user/schema_gen.go
```diff
@@ -0,0 +1,8249 @@
...
+	IncludeAccess        types.Bool          `tfsdk:"include_access"`
+	IsUsing2fa           types.Bool          `tfsdk:"is_using2fa"`
+	LastLoginDate        types.String        `tfsdk:"last_login_date"`
+	LastName             types.String        `tfsdk:"last_name"`
+	LastUpdated          types.String        `tfsdk:"last_updated"`
+	LinuxKeyPairId       types.Int64         `tfsdk:"linux_key_pair_id"`
+	LinuxPassword        types.String        `tfsdk:"linux_password"`
+	LinuxUsername        types.String        `tfsdk:"linux_username"`
+	LoginAttempts        types.Int64         `tfsdk:"login_attempts"`
+	LoginCount           types.Int64         `tfsdk:"login_count"`
```
matthew-c-hpe: I don't think having `LoginAttempts` or  `LoginCount` available in the data source is that useful, I think we could leave them out.

internal/subproviders/morpheus/datasources/user/schema_gen.go
```diff
@@ -0,0 +1,8249 @@
...
+			},
+		},
+	}
+}
+
+type UserModel struct {
+	Access               AccessValue         `tfsdk:"access"`
+	Account              AccountValue        `tfsdk:"account"`
+	AccountExpired       types.Bool          `tfsdk:"account_expired"`
+	AccountId            types.Int64         `tfsdk:"account_id"`
```
matthew-c-hpe: I think this should be renamed to `TenantId` in line with UI/documentation naming conventions.

RDavisHPE:
> Could you separate out the sdk version bump into a separate PR?

👍 
https://github.com/HPE/terraform-provider-hpe/pull/332

internal/subproviders/morpheus/datasources/user/schema_gen.go
RDavisHPE: Might still be useful in case someone just wants to check whether those values are set (null check), but I don’t feel strongly about it.

internal/subproviders/morpheus/datasources/user/schema_gen.go
RDavisHPE: Initially I was sort of thinking it could be useful for aggregating users created before/after a certain time - but just remembered the datasource works pretty explicitly with just a single user - pretty happy removing these.

internal/subproviders/morpheus/datasources/user/datasource.go
```diff
@@ -0,0 +1,540 @@
...
+	id int64,
+	data *UserModel,
+	apiClient *sdk.APIClient,
+) diag.Diagnostics {
+	diags := diag.Diagnostics{}
+
+	getUserRequest := apiClient.UsersAPI.GetUser(ctx, id)
+
+	// Add includeAccess parameter if specified
+	if !data.IncludeAccess.IsNull() && data.IncludeAccess.ValueBool() {
```
matthew-c-hpe: If you end up leaving in the access field, I think we should simplify the data source schema by omitting `include_access` from it and instead always setting `IncludeAccess` in the request to true here.

That way it will be more consistent with the other data source schemas that we have - so far the only attributes we've allowed the user to set have been those which filter what Morpheus resource the data source returns (`id` and resource `name`).

RDavisHPE: Seems fair - my main reason for including it was that the access blob is huge (kilobytes of data) so the user might not want to download/manage all of that.

internal/subproviders/morpheus/datasources/user/schema_gen.go
RDavisHPE: https://github.com/hpe-cds/hpe-morpheus-provider-code-spec/pull/134

internal/subproviders/morpheus/datasources/user/datasource.go
RDavisHPE: Defaulting includeAccess query parameter to be true :+1:

internal/subproviders/morpheus/datasources/user/datasource.go
RDavisHPE: Seperated - with the caveat that we are now using a SDK struct name in the function signature so if that gets renamed it will be a breaking change.

internal/subproviders/morpheus/datasources/user/datasource.go
```diff
@@ -0,0 +1,546 @@
...
+							"access": types.StringValue(in.GetAccess()),
+							"id":     types.Int64Value(in.GetId()),
+							"name":   types.StringValue(in.GetName()),
+						}),
+				)
+			}
+			if len(blueprintsValues) == 0 {
+				return types.SetNull(BlueprintsValue{}.Type(ctx))
+			}
+			set, _ := types.SetValue(BlueprintsValue{}.Type(ctx), blueprintsValues)
```
matthew-c-hpe: With the function based implemention you've done here for building these up, we lose out on diagnostics information if the nested access types fail to correctly construct a set. I think this could be made more resilient by exiting early and returning those diagnostics if it fails to construct the nested types.

I'd recommend taking a look at Role's `populateGetRoleAsStatePermissions()` and changing these to a similar pattern so that we can propagate the diagnostics of any failed set conversion to the caller of `getAccessAsState`.

RDavisHPE: Updated - much cleaner now (if a little longer with the added checks)

matthew-c-hpe (approved):
LGTM, thanks

PR #323, opened 2025-09-30
Adding User Datasource Acceptance Tests to Morpheus subprovider

RDavisHPE:
Adding the User datasource acceptance tests to the Morpheus subprovider.
- Added acceptance tests in datasource_test.go
- Added test terraform templates and examples under user datasource folder.

### DIFF
--- a/internal/subproviders/morpheus/datasources/user/datasource_test.go
+++ b/internal/subproviders/morpheus/datasources/user/datasource_test.go
@@ -1,0 +1,422 @@
+// (C) Copyright 2025 Hewlett Packard Enterprise Development LP
+
+package user_test
+
+import (
+	"os"
+	"regexp"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-framework/providerserver"
+	"github.com/hashicorp/terraform-plugin-go/tfprotov6"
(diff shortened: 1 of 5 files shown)


internal/subproviders/morpheus/datasources/user/datasource_test.go
```diff
@@ -0,0 +1,431 @@
...
+				Config:      config,
+				Check:       checkFn,
+				ExpectError: regexp.MustCompile(expected),
+			},
+		},
+	})
+}
+
+// Test to verify that all of the attributes from a created user can be read
+func TestAccMorpheusFindUserVerifyAttributes(t *testing.T) {
```
matthew-c-hpe, line 255: I'm guilty of having included this sort of `VerifyAttribute` test, but going off recent discussions around testing it may be overkill. Maybe we should discuss these with the team.

RDavisHPE: Seems we're keeping this as per the discussion. :+1:

matthew-c-hpe (approved):
Approving, but we'll need a follow-up PR to rename `account` field to `tenant`

PR #325, opened 2025-10-01
Implementing Update method to ServicePlan resource

RDavisHPE:
Implementing Update method for ServicePlan resource to handle updates to existing service plans in Morpheus as part of the Terraform provider.
- Implemented Update method in internal/subproviders/morpheus/resources/serviceplan/resource.go
- Implemented acceptance tests in update_test.go

Also updated `ValidateConfig` to allow unknown variables for custom_max_memory, custom_max_storage, add_volumes and custom_cores as this allows for use of variables in the acceptance tests without triggering validation.


internal/subproviders/morpheus/resources/serviceplan/resource.go
```diff
@@ -398,16 +503,163 @@ func (r *Resource) Create(
...
+	var plan, state ServicePlanModel
+
+	// Get prior state (has the ID) before touching any Value* methods.
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if state.Id.IsNull() || state.Id.IsUnknown() {
```
matthew-c-hpe: Assuming Import and Create have been coded correctly, it should be impossible to get to the point where ID isn't set when we perform an update, so this check is redundant.

Create and Import for Service Plan have enough guarantees to make sure ID isn't null or unknown.

internal/subproviders/morpheus/resources/serviceplan/resource.go
```diff
@@ -398,16 +503,163 @@ func (r *Resource) Create(
...
+	}
+
+	if !plan.PriceSetIds.IsNull() && !plan.PriceSetIds.IsUnknown() {
+		var priceSetIDs []int64
+		diags := plan.PriceSetIds.ElementsAs(ctx, &priceSetIDs, false)
+		resp.Diagnostics.Append(diags...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+		var pricesets []sdk.AddServicePlansRequestServicePlanPriceSetsInner
```
matthew-c-hpe: You should be able to skip this step and convert straight to `UpdateServicePlansRequestServicePlanPriceSetsInner`

internal/subproviders/morpheus/resources/serviceplan/update_test.go
```diff
@@ -0,0 +1,446 @@
...
+					resource.TestCheckResourceAttr("hpe_morpheus_service_plan.test", "config_ranges.min_cores_per_socket", "2"),
+					resource.TestCheckResourceAttr("hpe_morpheus_service_plan.test", "config_ranges.max_cores_per_socket", "20"),
+				),
+			},
+		},
+	})
+}
+
+// Helper function to merge config variables
+func mergeConfigVars(base config.Variables, overrides config.Variables) config.Variables {
```
matthew-c-hpe: You don't need to merge the config vars; the point of using default values in the variables you're using in your config is so that you can simply provide a single `config.Variable` for each step and it'll detect changes only on that. For the rest, it'll fall back to your default values. So you can remove this and reduce the base Config vars to something like stuart does, i.e. just a single variable.

internal/subproviders/morpheus/resources/serviceplan/resource.go
RDavisHPE: Good catch - removed this check 👍

internal/subproviders/morpheus/resources/serviceplan/resource.go
RDavisHPE: Surprised I missed that one! Also noticed that the struct takes `AddServicePlanRequestServicePlanPriceSetsInner`: 
```go
type UpdateServicePlansRequestServicePlan struct { /* ... */
	// List of price sets to include in service plan
	PriceSets            []AddServicePlansRequestServicePlanPriceSetsInner `json:"priceSets,omitempty"`
 }
```

internal/subproviders/morpheus/resources/serviceplan/update_test.go
RDavisHPE: Good catch - removed that function and set the only base config var to `name` 👍

internal/subproviders/morpheus/resources/serviceplan/resource.go
matthew-c-hpe: Yeah - that's down to how OpenAPI Generator will reuse generated nested structs with equivalent shapes. Can be pretty annoying and confusing and potentially prone to stuff breaking (names of struct types).

internal/subproviders/morpheus/resources/serviceplan/update_test.go
```diff
@@ -0,0 +1,440 @@
...
+					resource.TestCheckResourceAttr("hpe_morpheus_service_plan.test", "config_ranges.max_storage", "4"),
+					resource.TestCheckResourceAttr("hpe_morpheus_service_plan.test", "config_ranges.min_memory", "2097152"),
+					resource.TestCheckResourceAttr("hpe_morpheus_service_plan.test", "config_ranges.max_memory", "4194304"),
+					resource.TestCheckResourceAttr("hpe_morpheus_service_plan.test", "config_ranges.min_cores", "2"),
+					resource.TestCheckResourceAttr("hpe_morpheus_service_plan.test", "config_ranges.max_cores", "4"),
+					resource.TestCheckResourceAttr("hpe_morpheus_service_plan.test", "config_ranges.min_sockets", "2"),
+					resource.TestCheckResourceAttr("hpe_morpheus_service_plan.test", "config_ranges.max_sockets", "20"),
+					resource.TestCheckResourceAttr("hpe_morpheus_service_plan.test", "config_ranges.min_cores_per_socket", "2"),
+					resource.TestCheckResourceAttr("hpe_morpheus_service_plan.test", "config_ranges.max_cores_per_socket", "20"),
+				),
```
matthew-c-hpe, line 440: Let's add another step to test an update of `storage_size_type` and `memory_size_type`, or include them in here. We want to make sure there's no side effects from updating them. See how the API behaves first though, separate to the acceptance tests via curl or Morpheus CLI

RDavisHPE: Added to final apply (could also think of adding a check that max_storage and max_memory haven't changed in their respective types individual checks?)

Going off experimentation they don't change with changing the types so this is probably enough :+1:

RDavisHPE: Added to final apply. If we wanted we could also have:
```go
{ // memory_size_type change isolation test - verify no side effects on max_memory
	Config: baseConfigText,
	ConfigVariables: config.Variables{
		"name":             config.StringVariable(uniqueName),
		"memory_size_type": config.StringVariable("gb"),
	},
	Check: resource.ComposeTestCheckFunc(
		// Verify the size type changed
		resource.TestCheckResourceAttr("hpe_morpheus_service_plan.test", "memory_size_type", "gb"),
		// Verify max_memory value unchanged (no side effects)
		resource.TestCheckResourceAttr("hpe_morpheus_service_plan.test", "max_memory", "4294967296"),
	),
	}, { // storage_size_type change isolation test - verify no side effects on max_storage
	Config: baseConfigText,
	ConfigVariables: config.Variables{
		"name":              config.StringVariable(uniqueName),
		"storage_size_type": config.StringVariable("mb"),
	},
	Check: resource.ComposeTestCheckFunc(
		// Verify the size type changed
		resource.TestCheckResourceAttr("hpe_morpheus_service_plan.test", "storage_size_type", "mb"),
		// Verify max_storage value unchanged (no side effects)
		resource.TestCheckResourceAttr("hpe_morpheus_service_plan.test", "max_storage", "0"),
	),
},
```
However through experimentation it seems updating the number is done through the UI so this is unlikely to break. (edit: didn't realize I sent the previous reply)

matthew-c-hpe (approved):
Looks good to me, just update the description to capture the changes to `ValidateConfig`

PR #327, opened 2025-10-01
Add some sdk function aliases

jasonquigleyhpecom:
This PR add a couple of function aliases for working with the SDK cloud endpoints

PR #329, opened 2025-10-01
Add create for generic cloud

jasonquigleyhpecom:
This PR add create functionality for generic clouds

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/HPE/terraform-provider-hpe
 go 1.24.1
 
 require (
-	github.com/HewlettPackard/hpe-morpheus-go-sdk v0.0.0-20250919151946-3abcc4685d7e
+	github.com/HewlettPackard/hpe-morpheus-go-sdk v0.0.0-20251001122913-1f822867c74d
 	github.com/cenkalti/backoff/v5 v5.0.2
 	github.com/google/go-cmp v0.7.0
 	github.com/hashicorp/terraform-plugin-framework v1.16.0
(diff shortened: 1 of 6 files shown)


jasonquigleyhpecom:
This PR is part of a stack of 5 bookmarks:

1. `trunk()`
1. [cloud-func-aliases](https://github.com/HPE/terraform-provider-hpe/pull/327)
1. [cloud-read-generic](https://github.com/HPE/terraform-provider-hpe/pull/328)
1. **cloud-create-generic ← this PR**
1. [cloud-update-generic](https://github.com/HPE/terraform-provider-hpe/pull/330)
1. [cloud-tests-generic](https://github.com/HPE/terraform-provider-hpe/pull/331)

---
*Created with [jj-stack](https://github.com/keanemind/jj-stack)*

matthew-c-hpe (dismissed):
Code looks good to me, but separate to this should `config` and `hvm_config` have `ConflictsWith` validators? Or was the decision to allow both simultaneously intentional?

As I understand, the way it is now allows for additional properties to be set in the `config` attribute e.g. if using `hvm_config`. This could be useful down the line if something got added to the API for configuring a HVM cloud and we did not yet update `hvm_config` to match that (I'm aware a user could just use `config` and ignore `hvm_config` entirely to provision a HVM cloud if they wanted to).

jasonquigleyhpecom:
> Code looks good to me, but separate to this should `config` and `hvm_config` have `ConflictsWith` validators? Or was the decision to allow both simultaneously intentional?
> 
> As I understand, the way it is now allows for additional properties to be set in the `config` attribute e.g. if using `hvm_config`. This could be useful down the line if something got added to the API for configuring a HVM cloud and we did not yet update `hvm_config` to match that (I'm aware a user could just use `config` and ignore `hvm_config` entirely to provision a HVM cloud if they wanted to).

The validators were acting strange so I took them out for now in preference to getting it over the line

matthew-c-hpe (approved):

PR #331, opened 2025-10-01
Add tests for generic cloud

jasonquigleyhpecom:
This PR adds additional tests for generic clouds.

N.B. There is no import test yet because we don't have creds to test anything other than a standard cloud.

### DIFF
--- a/internal/subproviders/morpheus/resources/cloud/example_generic.tf
+++ b/internal/subproviders/morpheus/resources/cloud/example_generic.tf
@@ -1,0 +1,38 @@
+resource "hpe_morpheus_cloud" "example" {
+  # Required fields
+  name      = "TestCloud"
+  tenant_id = 1
+  group_id  = 1
+
+  # General configuration
+  code             = "aCode"
+  labels           = ["aLabel1", "aLabel2"]
+  enabled          = true
+  location         = "somewhere"
+  visibility       = "public"
+  cloud_type_code  = "standard"
+
+  # Agent and provisioning settings
+  agent_install_mode       = "ssh"
+  auto_recover_power_state = true
+
+  # Cost and guidance settings
+  costing_mode  = "costing"
+  guidance_mode = "off"
+
+  # Security settings
+  security_mode = "off"
+
+  # Shove into config
+  appliance_url            = "https://somewhere.com"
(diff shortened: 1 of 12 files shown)


jasonquigleyhpecom:
This PR is part of a stack of 5 bookmarks:

1. `trunk()`
1. [cloud-func-aliases](https://github.com/HPE/terraform-provider-hpe/pull/327)
1. [cloud-read-generic](https://github.com/HPE/terraform-provider-hpe/pull/328)
1. [cloud-create-generic](https://github.com/HPE/terraform-provider-hpe/pull/329)
1. [cloud-update-generic](https://github.com/HPE/terraform-provider-hpe/pull/330)
1. **cloud-tests-generic ← this PR**

---
*Created with [jj-stack](https://github.com/keanemind/jj-stack)*

internal/subproviders/morpheus/resources/cloud/example_generic.tf
```diff
@@ -0,0 +1,38 @@
...
+  auto_recover_power_state = true
+
+  # Cost and guidance settings
+  costing_mode  = "costing"
+  guidance_mode = "off"
+
+  # Security settings
+  security_mode = "off"
+
+  # Shove into config
```
matthew-c-hpe, line 26: I don't think we need to document that these end up in `config`, a user shouldn't care about this implementation detail.

internal/subproviders/morpheus/resources/cloud/example_generic.tf.tmpl
```diff
@@ -0,0 +1,38 @@
...
+  auto_recover_power_state = true
+
+  # Cost and guidance settings
+  costing_mode  = "costing"
+  guidance_mode = "off"
+
+  # Security settings
+  security_mode = "off"
+
+  # Shove into config
```
matthew-c-hpe: I don't think we need to document that these end up in config, a user shouldn't care about this implementation detail.

jasonquigleyhpecom: yikes, that was for my own sanity while testing - thanks for catching it

matthew-c-hpe (approved):

PR #339, opened 2025-10-07
hpe_morpheus_datastore commit 1

eamonnotoole:
In this PR we add the hpe_morpheus_datastore schema_gen.go.  We also add the relevant changes to resources_experimental.go commented out.


internal/subproviders/morpheus/resources_experimental.go
```diff
@@ -21,6 +21,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource"
```
stuart-mclaren-hpe, line 21: comments intentional?

jasonquigleyhpecom (approved):

internal/subproviders/morpheus/resources_experimental.go
stuart-mclaren-hpe: I see from the commit this is intentional.
Is this a pattern we should adopt? (I haven't tended to do this, so just wondering)

eamonnotoole: Yes.

PR #366, opened 2025-10-13
Use state values for Visibility and Active on Update

eamonnotoole:
If they aren't set in the config.  Vivek found an issue where the PUT
would return a 500 on the Update call during Create if Visiblity and 
Active hadn't been set in the config.  On debugging the issue is that 
if those values aren't set in the config then Morpheus sets them for 
you. We can't use a PlanModifier (UseStateForUnknown) since that 
requires that the State exists and is written out, which isn't the case 
on Create.  So we add code to use the State values for both - which 
are set to those returned by the API - if the values are Unknown in 
the Plan, i.e. not set by the user.

Note that When we Create we need to run an Update directly afterward
to set the resourcePermissions and tenants.

### DIFF
--- a/internal/subproviders/morpheus/resources/datastore/datastore_update.go
+++ b/internal/subproviders/morpheus/resources/datastore/datastore_update.go
@@ -36,11 +36,25 @@ func updateDatastore(
 	updateDatastore := sdk.NewUpdateCloudDatastoresRequestDatastoreWithDefaults()
 	updateDatastore.AdditionalProperties = make(map[string]any)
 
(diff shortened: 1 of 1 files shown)


matthew-c-hpe:
Why not just use `UseStateForUnknown` plan modifiers to do this?

Edit: Plan modifiers won't work because we need to perform an update immediately after create. The `Update()` method never gets called with knowledge of the plan at this point so it won't work as intended for Create.

matthew-c-hpe (dismissed):

matthew-c-hpe (approved):

PR #373, opened 2025-10-21
Switch to mux 2

eamonnotoole:
In this PR we:
- add the SDK v2.0 Morpheus provider.go and config.go to internal/sdkv2
- remove all of the resource and datasoure definitions
- adjust the Schema so that it matches that in framework
- switch the MarkdownDescription fields in the Framework provider Schema to Description to match that in SDK v2.0 which doesn't support MarkdownDescription
- use Hashicorp mux in main.go to combine the Framework and SDK v2.0 Morpheus providers
- regenerate the docs, there is a minor change to docs/index.md


internal/sdkv2/morpheus/config.go
```diff
@@ -0,0 +1,106 @@
...
+				Summary:  "SSL Certificate Verification Disabled",
+				Detail: `
+SSL certificate verification is disabled. To enable verification, set "secure = true" in your
+provider configuration or set environment variable MORPHEUS_API_SECURE to true.
+provider "morpheus" {
+	url = "https://..."
+	.
+	.
+	.
+	secure = true <-- set to true to enable certificate verification
```
ZoeySimone: This needs to be updated to match the schema change

internal/sdkv2/morpheus/config.go
```diff
@@ -0,0 +1,106 @@
...
+proceed with caution and understand the security implications of doing so. You can
+disable certificate verification by either setting the provider argument
+"secure = false" in your provider configuration or by setting the environment
+variable MORPHEUS_API_SECURE to false.
+provider "morpheus" {
+	url = "https://..."
+	.
+	.
+	.
+	secure = false <-- set to false to disable certificate verification
```
ZoeySimone: Needs to be updated to match the schema

main.go
```diff
@@ -51,9 +56,26 @@ func main() {
 		)
 	}
 
+	// sdkv2 Morpheus provider server
+	originalMorpheus, err := tf5to6server.UpgradeServer(context.Background(), sdkv2Morpheus.Provider().GRPCProvider)
```
matthew-c-hpe: originalMorpheus -> legacyMorpheus ?

eamonnotoole: Done.

internal/sdkv2/morpheus/config.go
eamonnotoole: Done.

ZoeySimone (dismissed):

internal/sdkv2/morpheus/config.go
```diff
@@ -0,0 +1,58 @@
...
+// Config is the configuration structure used to instantiate the Morpheus
+// provider.  Only Url and AccessToken are required.
+type Config struct {
+	Url             string
+	AccessToken     string
+	RefreshToken    string // optional and unused
+	Username        string
+	Password        string
+	ClientId        string
+	TenantSubdomain string
```
matthew-c-hpe: comment out TenantSubdomain until we decide if we need it (or remove it altogether, additionally from the schema, but I think we're unsure about why exactly it was included in the old provider).

eamonnotoole: It is gone from the Schema, I'll comment out here.

matthew-c-hpe (approved):

PR #374, opened 2025-10-23
Improve hpe_morpheus_datastore import

eamonnotoole:
Previously hpe_morpheus_datastore import would not populate the resource_permissions or tenants since that information is not returned by the datastores endpoint.  In this PR we add code to call the Cloud datastore or Cluster datastore endpoint to populate these stanzas on import.  Note that we only populate the groups list in resource_permissions for now.

We've updated the import acceptance test and the documentation.


internal/framework/subproviders/morpheus/resources/datastore/datastore_read.go
```diff
@@ -89,8 +89,44 @@ func getDatastoreAsState(
...
+		state.Tenants = plan.Tenants
+
 	case clusterRefType:
 		state.AssociatedResourceType = types.StringValue(associatedResourceTypeCluster)
+		// For imports, we need to get the tenants and resource permissions from the cluster datastore API
+		if plan.Name.IsNull() || plan.Name.IsUnknown() {
+			tenants, resourcePermissions, pdiags := populateClusterDatastoreInformation(
+				ctx, id, *datastore.RefId, client)
+			diags = append(diags, pdiags...)
+			if diags.HasError() {
```
ZoeySimone: Checking `diags.HasError()` but populate functions only add warnings. I'm guessing this error check wouldn't trigger

internal/framework/subproviders/morpheus/resources/datastore/datastore_read.go
```diff
@@ -321,6 +353,249 @@ func getDatastoreAsState(
...
+// only to be called on Import
+func populateClusterDatastoreInformation(
+	ctx context.Context,
+	id, clusterId int64,
+	client *sdk.APIClient,
+) (types.Set, ResourcePermissionsValue, diag.Diagnostics) {
+	var diags diag.Diagnostics
+
+	// Get cluster datastore information
+	cdResp, hresp, err := client.ClustersAPI.GetClusterDatastore(ctx, int32(clusterId), id).Execute() // nolint:gosec
```
ZoeySimone: Not going to hold this PR hostage, but the`int32` cast here is an indicator that the openapi spec needs to be updated

internal/framework/subproviders/morpheus/resources/datastore/datastore_read.go
```diff
@@ -321,6 +353,249 @@ func getDatastoreAsState(
...
+// only to be called on Import
+func populateCloudDatastoreInformation(
+	ctx context.Context,
+	id, cloudId int64,
+	client *sdk.APIClient,
+) (types.Set, ResourcePermissionsValue, diag.Diagnostics) {
+	var diags diag.Diagnostics
+
+	// Get cloud datastore information
+	cdResp, hresp, err := client.CloudsAPI.GetCloudDatastores(ctx, float32(cloudId), id).Execute()
```
ZoeySimone: Same as the `int32` cast below.

internal/framework/subproviders/morpheus/resources/datastore/datastore_read.go
eamonnotoole: You're correct.  I can remove the check.

internal/framework/subproviders/morpheus/resources/datastore/datastore_read.go
eamonnotoole: I can do that, there are a couple of examples that I'd need to update.

internal/framework/subproviders/morpheus/resources/datastore/datastore_read.go
eamonnotoole: I'll fix-up the spec.

ZoeySimone (approved):

PR #387, opened 2025-10-30
Add legacy provider convert_types.go

matthew-c-hpe:
Adds the `convert_types.go` file from the legacy provider.

We're trusting that whatever's here works for the legacy provider code, but we can revisit this if we need to fix or extend it.

### DIFF
--- a/internal/sdkv2/morpheus/convert/convert_types.go
+++ b/internal/sdkv2/morpheus/convert/convert_types.go
@@ -6,7 +6,7 @@ import (
 
 // This file contains some helper methods for things like
 // converting values to one type or another.
-// Suprisingly this is not as simple as it should be
+// Surprisingly this is not as simple as it should be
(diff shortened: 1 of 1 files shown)


internal/sdkv2/morpheus/convert/convert_types.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+// This file contains some helper methods for things like
+// converting values to one type or another.
+// Suprisingly this is not as simple as it should be
+// but I probably don't know what I am doing though...
+// Hey, let's convert all our IDs to strings though, for real.
+
+func ToInt64(i interface{}) int64 {
+	return StringToInt64(i.(string))
```
ZoeySimone: Check the assertion actually worked

jasonquigleyhpecom (dismissed):

jasonquigleyhpecom (approved):

PR #392, opened 2025-10-31
Add sdkv2 resource: hpe_morpheus_cluster_mks_vsphere

matthew-c-hpe:
Adds the `hpe_morpheus_cluster_mks_vsphere` to the sdkv2 provider.

Resource renamed from `morpheus_vsphere_mks_cluster` to `hpe_morpheus_cluster_mks_vsphere`.

Type assertions have been made type safe.
Nil pointer access have been made safe.
Slice 0th element accesses have been made safe.

Docs template and generated documentation have been added.


examples/resources/hpe_morpheus_cluster_mks_vsphere/import.sh
```diff
@@ -0,0 +1 @@
+terraform import hpe_morpheus_cluster_mks_vsphere.tf_example_vsphere_mks_cluster 1
```
jasonquigleyhpecom, line 1: Is this meant to be commited?

matthew-c-hpe: yes, this is how they did the examples in legacy provider.

jasonquigleyhpecom: Never mind, I just got to the template

internal/sdkv2/morpheus/helpers/morpheus.go
```diff
@@ -0,0 +1,9 @@
+package helpers
+
+import (
+	"os"
+)
+
+// global stuff here
+
+var UseForce = (os.Getenv("USE_FORCE") == "true")
```
jasonquigleyhpecom, line 9: Is the use of this documented anywhere?

matthew-c-hpe: It's used in the legacy provider for vsphere mks cluster, aws instance, vsphere instance, mvm instance. It's not documented in the legacy code why this is used for the `force` option in the API call.  I'd need to check the Morpheus API source code to see what it does exactly.

jasonquigleyhpecom: I've seen it used for deleting clouds as there is no way in the API to remove a cloud from a group

jasonquigleyhpecom (approved):

PR #400, opened 2025-11-03
Add image resource

ZoeySimone:
Adds an experimental image resource


internal/framework/subproviders/morpheus/resources/image/delete.go
```diff
@@ -0,0 +1,72 @@
...
+		return
+	}
+
+	waitForDeleted := func() (*sdk.GetInstance200Response, error) {
+		_, httpResp, err := client.LibraryAPI.GetVirtualImage(ctx, id.ValueInt64()).Execute()
+		if err != nil && httpResp.StatusCode != http.StatusNotFound {
+			return nil, backoff.Permanent(err)
+		}
+
+		// 404 status code counts as a successful delete
```
RDavisHPE: Do we actually need this if if the previous check confirms StatusCode is StatusNotFound? Suppose its nice to be explicit but I think the comment does enough and I spent a little time sanity checking the double "return nil, nil" as I was reading

internal/framework/subproviders/morpheus/resources/image/create.go
```diff
@@ -0,0 +1,327 @@
...
+	// if !plan.File.IsNull() && !plan.File.IsUnknown() {
+	// 	if _, err := os.Stat(plan.File.ValueString()); err != nil {
+	// 		resp.Diagnostics.AddError("could not read the specified image file", err.Error())
+
+	// 		return
+	// 	}
+	// }
+
+	// image_type
+	if !plan.ImageType.IsNull() && !plan.ImageType.IsUnknown() {
```
RDavisHPE: I believe ImageType is required so this check should not be necessary. I usually do required fields first for my resources but that's a stylistic choice.

internal/framework/subproviders/morpheus/resources/image/create.go
```diff
@@ -0,0 +1,327 @@
...
+		reqImage.VirtualImage.SetMinDisk(plan.MinDisk.ValueInt64() * 1024 * 1024 * 1024)
+	}
+
+	// min_ram
+	if !plan.MinRam.IsNull() && !plan.MinRam.IsUnknown() {
+		reqImage.VirtualImage.SetMinRam(plan.MinRam.ValueInt64() * 1024 * 1024 * 1024)
+	}
+
+	// name
+	if !plan.Name.IsNull() && !plan.Name.IsUnknown() {
```
RDavisHPE: Name is also required so shouldn't be null/unknown so you can probably remove this check

internal/framework/subproviders/morpheus/resources/image/create.go
ZoeySimone: Yeah, it's not necessary, but I do prefer keeping it the same as all the other attributes because my silly brain likes consistency. I am happy to make a change if you think I should

I (do my best to) sort the attributes alphabetically to make it easier to check if I missed anything.

internal/framework/subproviders/morpheus/resources/image/delete.go
ZoeySimone: The check on line 49 doesn't necessarily check that the status code is a 404, it only does so if the err value is also set.

I think having an extra check here is a good idea, even though the code does look a bit strange.

Looking closer at the code here made me notice that the status check if checking the wrong http response, so I made a change to fix that.

RDavisHPE: Nice one - that snuck past me too

internal/framework/subproviders/morpheus/resources/image/create.go
RDavisHPE: I like the alphabetical ordering idea more than required at the top honestly!

With that in mind - I _do_ think the code would be a bit clearer without the checks for the sake of a four-line removal. I like consistency too but I think philosophically if the code doesn't do anything it should be removed - and would keep the required-ness explicit in the code while still going with your alphabetic ordering.

internal/framework/subproviders/morpheus/resources/image/delete.go
RDavisHPE: Devil's advocate: could this function return a 404 _and_ an error? I imagine not but that was the one edge case that cropped up as I read your message.

RDavisHPE (dismissed):

internal/framework/subproviders/morpheus/resources/image/delete.go
jasonquigleyhpecom: The call always returns an error with a status code >= 300 so the following check is not required

ZoeySimone: Thanks. I've removed the unnecessary check.

I forgot that this is an SDK error and not an stdlib one.

RDavisHPE (approved):

PR #405, opened 2025-11-04
Alletra MP HVM datastore enableransomware change

eamonnotoole:
In this PR we adapt to the change in the Alletra MP HVM datastore plugin config where "enableRansomware" (bool) is switched to "enableransomware" (string).  We leave "enable_ransomware" (bool) in the datastore resource schema.


internal/framework/subproviders/morpheus/resources/datastore/datastore_create.go
```diff
@@ -206,3 +206,14 @@ func (r *Resource) Create(
 		return
 	}
 }
+
+func convertEnableRansomwareBool(input bool) (string, bool) {
```
jasonquigleyhpecom: We might want to generalise this in the future.

internal/framework/subproviders/morpheus/resources/datastore/datastore_create_cluster.go
```diff
@@ -18,8 +18,8 @@ var (
 	resourcePermissionsFuncCluster = sdk.NewSaveClusterDatastoreRequestDatastoreResourcePermissionsWithDefaults
 	permissionsSitesFuncCluster    = sdk.NewGetAlerts200ResponseAllOfChecksInnerAccountWithDefaults
 	datastoreTypeFuncCluster       = sdk.NewGetAlerts200ResponseAllOfChecksInnerAccountWithDefaults
-	nfsConfigFuncCluster           = sdk.NewSaveClusterDatastoreRequestDatastoreConfigAnyOfWithDefaults
-	alletrampHvmConfigFuncCluster  = sdk.NewSaveClusterDatastoreRequestDatastoreConfigAnyOf2WithDefaults
+	nfsConfigFuncCluster           = sdk.NewNFSDatastoreConfigurationWithDefaults
+	alletrampHvmConfigFuncCluster  = sdk.NewAlletraMPHVMDatastoreConfigurationWithDefaults
```
jasonquigleyhpecom: Would these be better named with Func at the end?

internal/framework/subproviders/morpheus/resources/datastore/datastore_read.go
```diff
@@ -654,3 +655,14 @@ func (r *Resource) Read(
 
 	resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
 }
+
+func convertEnableRansomwareString(input string) (bool, bool) {
```
jasonquigleyhpecom: Again, we might need to think about a generalised version of this in the convert package

jasonquigleyhpecom (dismissed):

internal/framework/subproviders/morpheus/resources/datastore/datastore_create.go
```diff
@@ -206,3 +206,14 @@ func (r *Resource) Create(
...
+	enableRansomwareMap := map[bool]string{
+		true:  "on",
```
ZoeySimone: The plugin code I mentioned in Slack yesterday also has a check for `(value == 'true')` we should maybe add `"true"` and `"false"` into the map

eamonnotoole: I'll add a convert function

internal/framework/subproviders/morpheus/resources/datastore/datastore_read.go
eamonnotoole: Done.

internal/framework/subproviders/morpheus/resources/datastore/datastore_create_cluster.go
eamonnotoole: Done.

internal/framework/subproviders/morpheus/convert/convert.go
```diff
@@ -61,6 +62,50 @@ func BoolToType(b *bool) types.Bool {
...
 }
 
+func StringToBool(s string) types.Bool {
+	switch strings.ToLower(s) {
+	case "on", "true", "yes":
+		return types.BoolValue(true)
+	case "off", "false", "no":
+		return types.BoolValue(false)
+	default:
+		return types.BoolValue(false)
```
ZoeySimone, line 73: It would make these functions a bit more annoying to use. But I wonder if we should error/warn if we reach the `default` case as it might produce unexpected results

eamonnotoole: I'll add an error to this function's returned values.

ZoeySimone (dismissed):

internal/framework/subproviders/morpheus/convert/convert.go
```diff
@@ -61,6 +62,50 @@ func BoolToType(b *bool) types.Bool {
 	return types.BoolValue(*b)
 }
 
+func StringToBool(s string) (types.Bool, error) {
+	switch strings.ToLower(s) {
+	case "on", "true", "yes":
+		return types.BoolValue(true), nil
+	case "off", "false", "no":
```
jasonquigleyhpecom: I think that unless we have the specific string to set to true, we should default to false

ZoeySimone (approved):

internal/framework/subproviders/morpheus/convert/convert.go
eamonnotoole: Done.

PR #433, opened 2025-11-06
Fix issue with datastore data-source and no resource_permissions

eamonnotoole:
The datastore data-source failed erroneously when no resource_permissions were specified on creation of the datastore resource.  This PR fixes that issue.  We add an acceptance test for this scenario.

### DIFF
--- a/internal/framework/subproviders/morpheus/datasources/datastore/datasource.go
+++ b/internal/framework/subproviders/morpheus/datasources/datastore/datasource.go
@@ -12,8 +12,10 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/types"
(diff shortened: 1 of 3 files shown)


internal/framework/subproviders/morpheus/datasources/datastore/datasource.go
```diff
@@ -316,67 +318,63 @@ func populateCloudDatastoreInformation(
...
+			return types.SetNull(TenantsType{}), NewResourcePermissionsValueNull(), retErr
+		}
 	}
 
 	// Populate ResourcePermissions, we'll only do Groups for now
+	var resourcePermissions ResourcePermissionsValue
 	rp, ok := cloudDatastore.GetResourcePermissionOk()
-	if !ok || rp == nil {
-		retErr := fmt.Errorf("datastore %d missing resource permissions in response", id)
+	if ok && rp != nil {
```
jasonquigleyhpecom, line 349: We were returning an error if there were no permission, is it okay not to have permissions?

eamonnotoole: Yeah, that's the issue that Vivek found.  If you don't set permissions on create then the datastore won't have any.

jasonquigleyhpecom (approved):

PR #434, opened 2025-11-06
Adding Sweeper to TestMain in Policy

RDavisHPE:
Adding a sweeper function to the TestMain for the Policy resource. This is probably the simplest approach to sweeper functionality - with some overhead if another resource acceptance test uses a policy they will likely have to call the sweeper manually (for policy this is unlikely - but might be worth considering if other resources adopt this pattern.)
- Added sweeper in sweeper_test.go
- Included sweeper call in TestMain in resource_test.go

### DIFF
--- a/internal/framework/subproviders/morpheus/resources/policy/sweep_test.go
+++ b/internal/framework/subproviders/morpheus/resources/policy/sweep_test.go
@@ -1,0 +1,93 @@
+// (C) Copyright 2025 Hewlett Packard Enterprise Development LP
+
+package policy_test
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"net/http"
+	"strings"
+
+	morpheuserrors "github.com/HPE/terraform-provider-hpe/internal/framework/subproviders/morpheus/errors"
+	"github.com/HPE/terraform-provider-hpe/internal/framework/subproviders/morpheus/test/sweep"
+)
+
+// Policies whose name begins with this string will be eligible for deletion
+const testPolicyPrefix = "TestAccMorpheusPolicy"
+
+// sweepPolicies cleans up test policies after tests complete
+func sweepPolicies() error {
+	ctx := context.Background()
+
+	client, err := sweep.NewSweepClient(ctx)
+	if err != nil {
+		// If we can't create a client (e.g., env vars not set), just log and continue
+		log.Printf("[INFO] Cannot create sweep client, skipping policy sweep: %v", err)
+		return nil
+	}
+
+	policies, hresp, err := client.PoliciesAPI.ListPolicies(ctx).
+		Phrase(testPolicyPrefix).Execute()
+	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/framework/subproviders/morpheus/resources/policy/sweep_test.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+// Policies whose name begins with this string will be eligible for deletion
+const testPolicyPrefix = "TestAccMorpheusPolicy"
+
+// sweepPolicies cleans up test policies after tests complete
+func sweepPolicies() error {
+	ctx := context.Background()
+
+	client, err := sweep.NewSweepClient(ctx)
```
jasonquigleyhpecom, line 23: Why not use the standard API client?

RDavisHPE: I believe the standard client we use for resources are attached to the resource itself i.e. 
```go
client, err := r.NewClient(ctx)
```
so that is inaccessible via the tests.

Using clientfactory I believe we would have to populate it with authorization via the enviornmental variables. This is already handled by the sweep's client function:
```go
func NewSweepClient(ctx context.Context) (*sdk.APIClient, error) {
	var username, password string

	url, ok := os.LookupEnv("TF_VAR_testacc_morpheus_url")
	if !ok {
		return nil, errors.New("TF_VAR_testacc_morpheus_url not set")
	}

	token, ok := os.LookupEnv("TF_VAR_testacc_morpheus_access_token")
	if !ok {
		username, ok = os.LookupEnv("TF_VAR_testacc_morpheus_username")
		if !ok {
			return nil, errors.New(
				"one of TF_VAR_testacc_morpheus_access_token or " +
					"TF_VAR_testacc_morpheus_username must be set",
			)
		}

		password, ok = os.LookupEnv("TF_VAR_testacc_morpheus_password")
		if !ok {
			return nil, errors.New(
				"one of TF_VAR_testacc_morpheus_access_token or " +
					"TF_VAR_testacc_morpheus_password must be set",
			)
		}
	}

	// If set to any value, use insecure
	_, insecure := os.LookupEnv("TF_VAR_testacc_morpheus_insecure")
	var opts []clientfactory.ClientOption
	if insecure {
		opts = append(opts, clientfactory.WithInsecureTLS())
	}

	client := clientfactory.NewAPIClient(
		ctx,
		url,
		username,
		password,
		token,
		opts...,
	)

	return client, nil
}
```

This could also have the advantage of if we keep both approaches and need to change something in particular for sweeper it would apply to both those which use the test/sweep approach and handling directly in TestMain. However if there is a cleaner way of getting a client with the relevant credentials outside of Resource it could apply here.

jasonquigleyhpecom: fair enough

jasonquigleyhpecom (dismissed):

ZoeySimone (approved):

PR #436, opened 2025-11-06
sdkv2: Switch to using NotFoundInResponseError

matthew-c-hpe:
Switches to using NotFoundInResponseError for nil checks on pointers to objects in sdkv2 `result` objects.

### DIFF
--- a/internal/sdkv2/morpheus/resources/cluster/resource_cluster_mks_vsphere.go
+++ b/internal/sdkv2/morpheus/resources/cluster/resource_cluster_mks_vsphere.go
@@ -414,7 +414,7 @@ func getClusterWorkers(client *morpheus.Client, clusterId int64) ([]morpheus.Clu
 	}
 
 	if workerResp.Workers == nil {
-		return []morpheus.ClusterWorker{}, helpers.NotFoundInResponseError("workerResp.Workers")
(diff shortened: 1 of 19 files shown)


internal/sdkv2/morpheus/resources/cluster/resource_cluster_mks_vsphere.go
```diff
@@ -722,7 +722,7 @@ func resourceClusterMKSVSphereCreate(ctx context.Context, d *schema.ResourceData
 	}
 
 	if result.Cluster == nil {
-		return diag.FromErr(helpers.NilPointerError("result.Cluster"))
+		return diag.FromErr(helpers.NotFoundInResponseError("result.Cluster"))
```
jasonquigleyhpecom: `result.Cluster` is an internal representation. I don't think we want that to be passed back to the customer

jasonquigleyhpecom (dismissed):

internal/sdkv2/morpheus/resources/cluster/resource_cluster_mks_vsphere.go
matthew-c-hpe: Fair enough. I think we still want the e.g. `Cluster` part.

jasonquigleyhpecom (dismissed):

jasonquigleyhpecom (approved):

PR #450, opened 2025-11-07
Add sdkv2 resource: hpe_morpheus_job_task

matthew-c-hpe:
Adds the `hpe_morpheus_job_task` data source.

The provider has been updated to be type safe on type assertions.

Adds template and generated docs.


internal/sdkv2/morpheus/resources/job/resource_job_task.go
```diff
@@ -0,0 +1,604 @@
...
+	var result map[string]any
+	if err := json.Unmarshal(resp.Body, &result); err != nil {
+		log.Fatal(err)
+	}
+
+	jobId := fmt.Sprintf("%v", result["id"])
+	// Successfully created resource, now set id
+	d.SetId(jobId)
+
+	resourceJobTaskRead(ctx, d, meta)
```
jasonquigleyhpecom, line 286: We should be using the diagnostics returned

matthew-c-hpe: You're right. I'll put up a separate PR to address this in all prior resources, including this one - this is how the code was originally, and so we'll have to go back and update all the other Create methods.

jasonquigleyhpecom (approved):

PR #452, opened 2025-11-07
sdkv2 add integration_ansible

jasonquigleyhpecom:


examples/resources/morpheus_integration_ansible/resource.tf
```diff
@@ -0,0 +1,19 @@
+data "morpheus_key_pair" "tf_example_key_pair" {
```
matthew-c-hpe, line 1: I've been removing unimplemented data sources from examples so far, but if we're gonna add them eventually before all the parity changes get merged back into the main dev branch, then I think this is fine to keep.

just something to be aware of.

matthew-c-hpe (approved):

PR #482, opened 2025-11-11
Implementing Policy Datasource

RDavisHPE:
Implementing a datasource for policies in the Morpheus subprovider so that
`hpe_morpheus_policy` can be used in Terraform configurations.
- Added datasource.go to define the datasource logic.


internal/framework/subproviders/morpheus/datasources/policy/datasource.go
```diff
@@ -0,0 +1,1039 @@
...
+	// Handle Tenants (Accounts)
+	if len(policy.Accounts) > 0 {
+		tenantValues := []attr.Value{}
+		tenantObjectType := types.ObjectType{
+			AttrTypes: map[string]attr.Type{
+				"id":   types.Int64Type,
+				"name": types.StringType,
+			},
+		}
+		for _, account := range policy.Accounts {
```
ZoeySimone, line 956: No need to make changes, but we have the `convert.ToSetType` function which would reduce some typing here

ZoeySimone (approved):

PR #520, opened 2025-11-12
sdkv2_morpheus_tenant

jasonquigleyhpecom:


examples/resources/morpheus_tenant/resource.tf
```diff
@@ -0,0 +1,15 @@
+data "morpheus_tenant_role" "example" {
```
matthew-c-hpe: should be `hpe_morpheus_role` - role datasource is generalised

matthew-c-hpe (approved):

PR #553, opened 2025-11-13
Implement in-place updates for the Image resource

ZoeySimone:
Adds in-place updating for most attributes in the Image resource.

`url` and `config_azure` will continue to require a replace as those fields may require extra logic to ensure
correct updating of the underlying image.

`description` requires a replace due to the API not supporting description updates at the moment.


eamonnotoole (approved):

eamonnotoole:
@ZoeySimone You should probably add a note to the docs about the restrictions that you mention in the commit message and source.

PR #590, opened 2025-11-17
parity_ds_morpheus_servicenow_workflow

jasonquigleyhpecom:


matthew-c-hpe:
This should be renamed to `workflow_servicenow`.

examples/data-sources/morpheus_servicenow_workflow/data-source.tf
```diff
@@ -0,0 +1,8 @@
+data "morpheus_integration" "servicenow_prod" {
```
matthew-c-hpe: this should be `hpe_morpheus_integration`

matthew-c-hpe (approved):

PR #605, opened 2025-11-17
parity_ds_morpheus_storage_volume

jasonquigleyhpecom:


internal/sdkv2/morpheus/datasources/storage/storage_volume.go
```diff
@@ -0,0 +1,157 @@
...
+	} else {
+		return diag.FromErr(helpers.TypeAssertFailError("Result", resp.Result))
+	}
+
+	if result.StorageVolume == nil {
+		return diag.FromErr(helpers.NotFoundInResponseError("StorageVolume"))
+	}
+
+	storageVolume := result.StorageVolume
+	d.SetId(convert.Int64ToString(int64(storageVolume.ID.(float64))))
```
matthew-c-hpe: insert a type assertion to int64 for the ID before this - the API returns an int

Then just pass the int64 to the convert function

matthew-c-hpe (approved):

PR #630, opened 2025-11-25
Adding Acceptance Tests for Policy Datasource

RDavisHPE:
Adding Acceptance Tests for Policy Datasource to ensure
- A policy can be found by name using the example
- A policy can be found by id using the example
- A policy datasource with no valid terms fails
- A policy datasource with both name and id specified fails
- A policy datasource correctly sets all attributes - especially those in the static schema blocks.
- A policy datasource correctly populates the cloud, user, role or group attribute if the associated resource type corresponds to that resource type.


internal/framework/subproviders/morpheus/datasources/policy/datasource_test.go
```diff
@@ -0,0 +1,935 @@
...
+			},
+		},
+	})
+}
+
+func TestAccMorpheusPolicyDataSourceBothSearchAttrs(t *testing.T) {
+	defer testhelpers.RecordResult(t)
+	t.Parallel()
+
+	if testing.Short() {
```
ZoeySimone: `if testing.Short()` can be removed as this only tests the schema

RDavisHPE: Amended :+1:

ZoeySimone (approved):

PR #631, opened 2025-11-25
Add image datasource

ZoeySimone:


internal/framework/subproviders/morpheus/datasources/image/image_test.go
```diff
@@ -0,0 +1,271 @@
...
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config:      providerConfig + dataSourceConfig,
+				ExpectError: errMatch,
+			},
+		},
+	})
+}
```
RDavisHPE, line 296: I tend to include the equivalent of `TestAccMorpheusImageDataSourceBothSearchAttrs` to test `conflicts with` for id and name/type.

internal/framework/subproviders/morpheus/datasources/image/image.go
```diff
@@ -0,0 +1,373 @@
...
+// Schema defines the schema for the data source.
+func (d *DataSource) Schema(
+	ctx context.Context,
+	_ datasource.SchemaRequest,
+	resp *datasource.SchemaResponse,
+) {
+	resp.Schema = ImageDataSourceSchema(ctx)
+}
+
+func (d *DataSource) Read(
```
RDavisHPE, line 57: Up to you but looking around the codebase it seems datasources are usually structured with Read at the bottom of the file

internal/framework/subproviders/morpheus/datasources/image/image.go
```diff
@@ -0,0 +1,373 @@
...
+		)
+
+		return diags
+	}
+
+	image := images[0]
+
+	// okay, we need to get a bit silly here
+	// despite the structure of both requests being the same our SDK generates
+	// two different structs. To avoid code duplication we marshal the image from
```
RDavisHPE, line 191: Per code duplication - I suppose the alternative would be taking the ID from the returned image and using getImageById? That might be tidier even though it necessitates a second GET request (I do the same in the User datasource - but for different reasons).

Still this prevents two requests and seems unlikely to fail.

internal/framework/subproviders/morpheus/datasources/image/image.go
```diff
@@ -0,0 +1,373 @@
...
+	var diags diag.Diagnostics
+
+	// auto_join_domain
+	data.AutoJoinDomain = convert.BoolToType(image.IsAutoJoinDomain)
+
+	// cloud_init
+	data.CloudInit = convert.BoolToType(image.IsCloudInit)
+
+	// config_azure
+	if image.GetImageType() == "azure-reference" {
```
RDavisHPE, line 232: This only comes to mind after working with configs for policy - but is there a case (perhaps after an update) where the Azure config might be populated if the image type is not "azure-reference" (even though it presumably "shouldn't").

And if so would this be worth returning (i.e. use the `image.GetConfig().AzureReferenceVirtualImageConfiguration != nil` check directly instead of checking the image type first).

internal/framework/subproviders/morpheus/datasources/image/image_test.go
```diff
@@ -0,0 +1,271 @@
...
+	errMatch := regexp.MustCompile("Attribute \"name\" must be specified when \"image_type\" is specified")
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config:      providerConfig + dataSourceConfig,
+				ExpectError: errMatch,
+			},
+		},
+	})
```
RDavisHPE, line 295: Also I tend to include a `VerifyAttributes` test to make sure everything gets populated - but mileage of this may vary with your resource so up to you.

internal/framework/subproviders/morpheus/datasources/image/image_test.go
ZoeySimone: Added the test, and renamed the tests to be named a bit better.

internal/framework/subproviders/morpheus/datasources/image/image.go
ZoeySimone: I find this a bit nicer as it keeps all of the methods implemented for a datasource together, but I don't mind changing it around.

internal/framework/subproviders/morpheus/datasources/image/image.go
ZoeySimone: 🤦‍♀️ this feels like a very obvious approach in hindsight, I think I'll stick with what I have for now as it does have some benefits. 

Somewhat relevant note: I've found that while running tests the `Read` function may get executed several times per test (presumably once for each step: plan, apply, destroy) reducing number of requests is not required, but might save us some trouble in the future.

internal/framework/subproviders/morpheus/datasources/image/image.go
ZoeySimone: The provider doesn't support updating the azure config and the UI doesn't support it either. I might do a test through the API to see if it can be populated through the API. The `imageType` also cannot be changed from `azure-reference` to something else afaik.

Either way if the `imageType` is not "azure-reference" the config values are not going to be used which makes them pretty useless to have in the datasource

RDavisHPE: Sounds reasonable in that case :+1:

internal/framework/subproviders/morpheus/datasources/image/image.go
RDavisHPE: :+1:

RDavisHPE (approved):

internal/framework/subproviders/morpheus/datasources/image/image_test.go
ZoeySimone: I'll add this in a separate PR

PR #673, opened 2025-12-04
Parity: task_javascript tests

matthew-c-hpe:
Adds acceptance test(s) for the resource task_javascript.

Also adds template files for the example Terraform config with a go:generate statement.


matthew-c-hpe:
Erroneously created with changesets from other branches

PR #720, opened 2025-12-11
Porting installation scripts from hpegl to hpe 

RDavisHPE:
Porting installation scripts and documentation from hpegl to hpe.
- Copied scripts and INSTALL.md from hpegl repository
- Renamed and reimplemented to use the `hpe` provider rather than `hpegl`.
- Implemented tests in BATS/Pester


INSTALL.md
```diff
@@ -0,0 +1,198 @@
...
+
+```hcl
+terraform {
+  required_providers {
+    hpe = {
+      source  = "HPE/hpe"
+      version = ">= x.y.z"
+    }
+  }
+  required_version = ">= 0.13"
```
eamonnotoole: We should use `>= 0.4.0` or something like that here, i.e. an actual `hpe` version

eamonnotoole (commented):
Very nice Ryan.  Just one minor comment.

INSTALL.md
RDavisHPE: Changed `version` to `>= 0.4.0` and `required_version` to `1.11` (reading our docs, `Write-Only` is only supported by `1.11` or later)

eamonnotoole (approved):

INSTALL.md
eamonnotoole: Great, thanks.

PR #731, opened 2025-12-17
sdkv2: Sync generated examples

matthew-c-hpe:
Picks up the generated docs changes from some of the migration work, i.e. run `go generate ./...`.


matthew-c-hpe:
Closing as this isn't needed, I got this diff set from running `go generate ./...`, I should have used `make docs` as `make docs` also runs `terraform fmt`. However, this means that some of our templates are formatted in such a way that `terraform fmt` affects their output.

PR #735, opened 2025-12-17
Metal 5: add code to handle Metal RAID

eamonnotoole:
In this PR we make extensive changes to getVolumes() to handle the discrepancies that we seeing between HCL and API Volumes for Metal RAID. We hope to remove the need for some of the processing done here with changes to the Metal plugin.

The changes:
- we pass plan down to getVolumes
- we add a number of new functions that process the API list of volumes:
  - removeRaidDisks() removes RAID disk volumes from the list of API volumes, the trigger is repeated deviceName and also diskMode of nil and diskType not nil
  - markRootVolume() sets the rootVolume bool in an API volume if the volume deviceDisplayName is "BootVolume" or "Boot Volume", we hope to remove the need for this function
  - bootVolumeFirst() puts the boot (rootVolume=true) volume first in the list, we hope to remove this function when we get the boot volume name to be that set in the HCL
  - reorderVolumes() reorders the list of volumes to match the plan with boot volume first, we hope to remove the need to have the boot volume first here (see above)
  - fillVolumeFieldsFromPlan() we hope to remove the need for this function when we get a datastoreId and also the "correct" storage_type_id from the API
  - setDatastoreAutoSelectionAndSize() sets the API volume size to that in the plan and also puts the value of DatasoreAutoSelection into the AdditionalProperties map
- if the number of volumes from the API matches the number of volumes from the plan then we call setDatatstoreAutoSelectionAndSize()
- we call the following functions in the below order for imports and build the State volumes from the resulting list of API volumes:
  - removeRaidDisks
  - markRootVolume
  - bootVolumeFirst
- we call the following functions in the below order where the number of volumes from the API and plan differ and build the State volumes from the resulting list of API volumes:
  - removeRaidDisks
  - markRootVolume
  - bootVolumeFirst
  - reorderVolumes
  - fillVolumeFieldsFromPlan
  - setDatastoreAutoSelectionAndSize

### DIFF
--- a/internal/framework/subproviders/morpheus/resources/instance/instance_read.go
+++ b/internal/framework/subproviders/morpheus/resources/instance/instance_read.go
@@ -286,7 +286,7 @@ func getVolumes(
 	// Import
 	if plan.Name.IsNull() || plan.Name.IsUnknown() {
 		nonRaidVolumes := removeRaidDisks(apiVolumes)
-		markedRootVolumes := markRootVolume(nonRaidVolumes)
+		markedRootVolumes := markRootVolume(nonRaidVolumes, plan)
 		bootVolumeInFirst := bootVolumeFirst(markedRootVolumes)
 
 		return convertAPIVolumesToStateVolumes(ctx, bootVolumeInFirst)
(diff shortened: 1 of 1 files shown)


internal/framework/subproviders/morpheus/resources/instance/instance_read.go
```diff
@@ -282,6 +283,230 @@ func getVolumes(
...
+		// Only fill from plan if we have a corresponding plan volume
+		if i < maxIndex {
+			planVol := planVolumes[i].(VolumesValue)
+			if apiVol.DatastoreId == nil {
+				apiVol.DatastoreId = planVol.DatastoreId.ValueInt64Pointer()
+			}
+			// We set TypeId for all volume types since for some types (e.g. Metal RAID) the API TypeId is different
+			apiVol.TypeId = planVol.StorageTypeId.ValueInt64Pointer()
+		}
+		// If i >= maxIndex, just append the apiVol as-is (unmatched volumes)
```
jasonquigleyhpecom, line 383: These comments are anchored to the wrong code

jasonquigleyhpecom (approved):

PR #739, opened 2025-12-18
Adding TaintResourceState for when Create fails after POST

RDavisHPE:
This commit introduces utility functions for assisting in removing resources should there
be any errors on creation to avoid dangling resources.
- Added the following functions to utils/resource_cleanup.go
  ~~- CleanupResourceOnError: cleans up the resource~~ - sticking with the tainting appraoch
  - TaintResourceState: taints the resource

TaintResourceState has been applied for all resources so that the resource is tainted should Create fail post-creation.
Also updated some errors to make it clearer that a resource has been created.


eamonnotoole (approved):
Very nice Ryan

PR #741, opened 2025-12-18
Update render path of Instance Terraform template

matthew-c-hpe:
Updates the path where the instance example gets rendered to `examples/resources/morpheus_instance`.

Adds an `instance_example.go` file to hold the `go:generate` statement.

This is for uniformity with other resources.


matthew-c-hpe:
Converting this to draft for now until we complete the dev-0.5 and parity sync - I think some of these changes would cause merge conflicts.

matthew-c-hpe:
Closing this to tidy things up, not too important that this gets brought in and still will have a record of these changes if we need them in the future.

PR #744, opened 2025-12-19
Tidy formatting of RenderConfig functions

matthew-c-hpe:
Tidies the formatting of the Render{Resource}Config functions by doing the following:

- Adopts a pattern of passing a variable `args []string` to `RenderExample`. `args` is built up by iterating over the `defaults` map of each render function. This avoids hardcoding the arguments to `RenderExample` and avoids developer error in maintaining a correct list of args for each function.

- Standardise formatting of function signature to be `(t *testing.T, overrides map[string]string) (string, error)`. This prevents stuttering of overriding the `name` which we had previously. Overriding `Name` now uses an `override` like any other overridden parameter.

- Update function invocations with new formatting (i.e. remove `name` from parameter list, make it an override).

- Use `Example` and `example` as default values for `Name` and `Code` respectively for each rendered config.

- Make function signatures all one line long.

- General tidy of formatting within functions for consistency.

- Small fixes in certain tests. `Cypher` render invocations previously used `Name` instead of `Key`. `Key` is correct for these functions.

### DIFF
--- a/internal/sdkv2/morpheus/resources/automation/hpe_morpheus_execute_schedule_resource_test.go
+++ b/internal/sdkv2/morpheus/resources/automation/hpe_morpheus_execute_schedule_resource_test.go
@@ -47,11 +47,9 @@ func TestAccMorpheusExecuteScheduleExampleOk(t *testing.T) {
 
 	name := acctest.RandomWithPrefix(t.Name())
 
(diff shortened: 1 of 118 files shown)


RDavisHPE (approved):
Grug happy!

PR #749, opened 2025-12-19
Adding validation for VRO integration resource

RDavisHPE:
Adding validation for VRO integration resource to avoid 400 errors. The required-ness of attributes here depends on the `auth_type` (which was previously only set to basic). This PR adds the other auth_types and includes validation to ensure
- api_token is set when auth_type is 'aria'
- username, password, tenant and auth_id are set otherwise.

This avoids confusing 400 errors.

### DIFF
--- a/internal/sdkv2/morpheus/resources/integration/integration_vro.go
+++ b/internal/sdkv2/morpheus/resources/integration/integration_vro.go
@@ -70,7 +70,8 @@ func validateVROAuthConfig(ctx context.Context, d *schema.ResourceDiff, meta any
 		}
 
 		if username == "" || password == "" || tenant == "" || authId == "" {
-			return fmt.Errorf("the following fields are required when auth_type is not 'aria': username, password, tenant, auth_id")
+			return fmt.Errorf("the following fields are required when auth_type is not 'aria': " +
+				"username, password, tenant, auth_id")
 		}
 	}
 
--- a/internal/sdkv2/morpheus/resources/integration/integration_vro.go
+++ b/internal/sdkv2/morpheus/resources/integration/integration_vro.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/hex"
+	"fmt"
 	"log"
 	"strings"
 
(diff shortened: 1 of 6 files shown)


internal/sdkv2/morpheus/resources/integration/integration_vro.go
```diff
@@ -19,13 +20,52 @@ import (
...
+
+	if authType == "aria" {
+		// For aria auth type, api_token is required
+		apiToken := d.Get("api_token").(string)
+		if apiToken == "" {
+			return fmt.Errorf("api_token is required when auth_type is 'aria'")
+		}
+	} else {
+		// For non-aria auth types, username, password, tenant, and auth_id are required
+		username := d.Get("username").(string)
```
matthew-c-hpe: See the rest of the codebase for the `if, ok` pattern we've adopted for the migrated sdkv2 code to prevent possible panics from failed type assertions.

RDavisHPE: :+1:

internal/sdkv2/morpheus/resources/integration/integration_vro.go
```diff
@@ -19,13 +20,85 @@ import (
...
+		}
+		if tenant == "" {
+			missing = append(missing, "tenant")
+		}
+		if authId == "" {
+			missing = append(missing, "auth_id")
+		}
+
+		if len(missing) > 0 {
+			return fmt.Errorf("the following fields are required when auth_type is not 'aria': %s", strings.Join(missing, ", "))
```
matthew-c-hpe: I see what you're doing here but I think it'd be clearer to always explicitly list the four required fields - even if the user got some of them right. So I think this could be all squashed into an if statement, i.e.:

```
		if username == "" || password == "" || tenant == "" || authId == "" {
			return errors.New("the following fields are required when auth_type is not 'aria': username, password, tenant, auth_id")
		}
```

RDavisHPE: Fair :+1:

matthew-c-hpe (approved):

PR #751, opened 2026-01-08
Emit SDK error on datastore create retry loop read

eamonnotoole:
We've seen situations where a datastore create fails but the status in the error message is blank.  This suggests that the SDK call in the retry look to get the datastore failed.  If this happens we unwrap the error to obtain the underlying SDK error and add it to the returned error message.

### DIFF
--- a/internal/framework/subproviders/morpheus/resources/datastore/datastore_create.go
+++ b/internal/framework/subproviders/morpheus/resources/datastore/datastore_create.go
@@ -167,10 +167,23 @@ func (r *Resource) Create(
 			status = r.GetDatastore().Status
 		}
 
-		resp.Diagnostics.AddError(
-			"datastore provisioning failed",
-			fmt.Sprintf("Datastore %d failed to reach provisioned status. Current status: %v", id, status),
-		)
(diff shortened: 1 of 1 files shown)


internal/framework/subproviders/morpheus/resources/datastore/datastore_create.go
```diff
@@ -167,10 +167,23 @@ func (r *Resource) Create(
...
+
+		if errUnwrapped != nil {
+			resp.Diagnostics.AddError(
+				"datastore provisioning failed",
+				fmt.Sprintf("Datastore %d failed to reach provisioned status: %v", id, errUnwrapped),
+			)
+		} else {
+			resp.Diagnostics.AddError(
+				"datastore provisioning failed",
+				fmt.Sprintf("Datastore %d failed to reach provisioned status. Current status: %v", id, status),
```
jasonquigleyhpecom: `%s` instead of `%v`

matthew-c-hpe (dismissed):

jasonquigleyhpecom (dismissed):

jasonquigleyhpecom (approved):

PR #834, opened 2026-01-16
Fixing plan-level validation across ported resources

RDavisHPE:
Adding various requireds and plan validation across resources ported from `terraform-provider-morpheus`. This includes fixing any obvious errors when just running each configuration with just their required attributes.

This includes:
- For `hpe_morpheus_integration_servicenow` and `hpe_morpheus_integration_ansible_tower` - a username and password *or* a credential id must be provided or else the resource will 400. This has been added as validation.
- An internal logic error causing a TypeAssertion error whenever a `hpe_morpheus_form` is created without labels (which should be optional) has been resolved.
- A typo in the API structure for `hpe_morpheus_resource_pool_group` has been resolved.
- Various requireds have been added:
	- `code` for `hpe_morpheus_environment` has been marked as required as it 400s otherwise.
	- `spec_template_ids` for `hpe_morpheus_cluster_package` has been marked as it 400s otherwise.

### DIFF
--- a/docs/resources/morpheus_cluster_package.md
+++ b/docs/resources/morpheus_cluster_package.md
@@ -33,6 +33,7 @@ resource "hpe_morpheus_cluster_package" "tf_example_cluster_package" {
 - `name` (String) The name of the cluster package
 - `package_type` (String) A one word descriptor of package, such as calico, rook, prometheus, etc.
 - `package_version` (String) The version of the cluster package
+- `spec_template_ids` (List of Number) A list of spec template ids associated with the cluster package
 - `type` (String) The package category type (apps, custom, ingress, logging, monitoring, morpheus, network, serviceMesh, storage)
 
 ### Optional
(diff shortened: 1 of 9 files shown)


RDavisHPE:
Note: while `master_node_pool` under `hpe_morpheus_cluster_mks_vsphere` is currently not marked as required - currently the internal logic will throw `master_node_pool: slice is empty` if it is excluded. 

This could be because of an issue with our handling:
```go
if masterPoolValue := d.Get("master_node_pool"); masterPoolValue != nil {
	if masterPoolSlice, ok := masterPoolValue.([]interface{}); ok {
		if len(masterPoolSlice) == 0 {
			return diag.FromErr(helpers.EmptySliceError("master_node_pool"))
		}
		if masterPoolMap, ok := masterPoolSlice[0].(map[string]interface{}); ok {
			masterpool = masterPoolMap
		} else {
			return diag.FromErr(helpers.TypeAssertFailError("master_node_pool[0]", masterPoolSlice[0]))
		}
	} else {
		return diag.FromErr(helpers.TypeAssertFailError("master_node_pool", masterPoolValue))
	}
}
```
Since SDK v2.0 will always return []interface{} if `master_node_pool` is not included - it will always hit the `EmptySliceError`.

If we can be confident (i.e. via the UI) `master_node_pool` (and the same applies for `worker_node_pool`) are required this issue is circumvented - otherwise logic may need to be revisited here.

The PR leaves the implementation alone for now.

jasonquigleyhpecom (approved):

PR #858, opened 2026-01-23
Fix stack trace from instance backoff code

eamonnotoole:
QA testing has uncovered a situation where the instance Delete backoff code returned a stack-trace.  The issue seems to be that the response doesn't contain valid JSON so the code to extract the "status" stack-traces.  To fix this:
- we check that hresp is not nil before getting the status from it
- we add SDK GetInstanceOk() and GetStatusOk() to check that there is an instance record and within it a status to read
- if either function fails we raise a "Permanent" backoff error to terminate the backoff polling.

We've applied this logic to Create, Delete and Update.


RDavisHPE (dismissed):
Looks good - I noticed there's similar code for datastore CRD as well. Not certain if we can reproduce something similar there (in terms of the API throwing back non-JSON) but might be worth considering.

jasonquigleyhpecom (dismissed):

jasonquigleyhpecom (approved):

PR #877, opened 2026-01-29
Fix datastore and image nil hresp stack-trace

eamonnotoole:
A stack-trace due to a nil httpResponse object has been observed in instance.  The same could occur for datastore and image.  This PR adds code to fix the issue for datastore and image.

### DIFF
--- a/internal/framework/subproviders/morpheus/resources/datastore/datastore_create.go
+++ b/internal/framework/subproviders/morpheus/resources/datastore/datastore_create.go
@@ -145,7 +145,7 @@ func (r *Resource) Create(
 	waitForReady := func() (*sdk.GetDatastores200Response, error) {
 		response, hresp, err := client.DatastoresAPI.GetDatastores(ctx, id).Execute()
 		if err != nil {
-			if hresp == nil || hresp != nil && hresp.StatusCode != http.StatusOK {
+			if hresp == nil || hresp.StatusCode != http.StatusOK {
 				return nil, backoff.Permanent(err)
 			}
 		}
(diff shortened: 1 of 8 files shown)


internal/framework/subproviders/morpheus/resources/datastore/datastore_create.go
```diff
@@ -144,8 +144,10 @@ func (r *Resource) Create(
 	// Wait for the datastore to be ready
 	waitForReady := func() (*sdk.GetDatastores200Response, error) {
 		response, hresp, err := client.DatastoresAPI.GetDatastores(ctx, id).Execute()
-		if err != nil || hresp.StatusCode != http.StatusOK {
-			return nil, backoff.Permanent(err)
+		if err != nil {
+			if hresp == nil || hresp != nil && hresp.StatusCode != http.StatusOK {
```
jasonquigleyhpecom: `hresp != nil && ` is redundant

jasonquigleyhpecom: Same with all similar tests

jasonquigleyhpecom (changes requested):

internal/framework/subproviders/morpheus/resources/datastore/datastore_create.go
eamonnotoole: Yeah I figured as much

internal/framework/subproviders/morpheus/resources/image/delete.go
RDavisHPE: Nit: copyright year [I only remembered because you did it for the other ones :v) ]

eamonnotoole: Thanks Ryan.

jasonquigleyhpecom (approved):

PR #878, opened 2026-01-30
Add Morpheus task resource

ZoeySimone:
Adds a Morpheus task resource with support for Conditional Workflows and generic task config


internal/framework/subproviders/morpheus/resources/task/create.go
```diff
@@ -0,0 +1,175 @@
...
+	taskOptionsSet := false
+
+	// config
+	taskOptions := &sdk.AddTasksRequestTaskTaskOptions{}
+	if !plan.Config.IsNull() && !plan.Config.IsUnknown() {
+		configValue := plan.Config.UnderlyingValue()
+		configAny, err := convert.ValueToAny(ctx, configValue)
+		if err != nil {
+			resp.Diagnostics.AddError(
+				"create instance resource",
```
matthew-c-hpe: create task resource

internal/framework/subproviders/morpheus/resources/task/create.go
```diff
@@ -0,0 +1,175 @@
...
+				"instance: failed to convert config: "+
```
matthew-c-hpe: task

internal/framework/subproviders/morpheus/resources/task/create.go
```diff
@@ -0,0 +1,175 @@
...
+	if !plan.ExecuteTarget.IsNull() && !plan.ExecuteTarget.IsUnknown() {
+		addTaskReq.Task.SetExecuteTarget(plan.ExecuteTarget.ValueString())
+	}
+
+	// labels
+	if !plan.Labels.IsNull() && !plan.Labels.IsUnknown() {
+		labels, err := convert.SetToStrSlice(plan.Labels)
+		if err != nil {
+			resp.Diagnostics.AddError(
+				"create image resource",
```
matthew-c-hpe: task

internal/framework/subproviders/morpheus/resources/task/create.go
```diff
@@ -0,0 +1,175 @@
...
+				"image "+plan.Name.ValueString()+": failed to parse label: "+err.Error(),
```
matthew-c-hpe: task

internal/framework/subproviders/morpheus/resources/task/create.go
ZoeySimone: Fixed

internal/framework/subproviders/morpheus/resources/task/schema_gen.go
```diff
@@ -0,0 +1,742 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+)
+
+func TaskResourceSchema(ctx context.Context) schema.Schema {
+	return schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"account_id": schema.Int64Attribute{
+				Computed: true,
+			},
+			"allow_custom_config": schema.BoolAttribute{
```
matthew-c-hpe, line 32: Do we really need this attribute in the schema?

ZoeySimone: I've kept it as it's an option that's available both in the UI and the CLI. TBH not quite sure exactly how the field is actually used in Morpheus. Even though I don't understand what it's used for I feel like it is probably useful to have in a generalized task resource.

matthew-c-hpe: In that case, should it be changed to be both computed and optional?

matthew-c-hpe: marking as resolved, we'll add computed and optional in a follow up pr as this requires openapi spec changes.

matthew-c-hpe (approved):

PR #882, opened 2026-02-03
Move example rendering code

ZoeySimone:


matthew-c-hpe:
I don't think we're missing much with this at this point. Closing this.

PR #890, opened 2026-02-12
fix: Ansible Tower - Wrong inventory id and job_template id can be returned from data source

drokath:
When multiple Ansible tower are declared with the same inventory, the first found is returned.

During the creation of a job_tasks with the retrieve inventory or template id, the API does not return an error if they do not match ones supported by provided integration specified in the job_template resource

Retrieving the correct inventory id is a first step to fix this issue.


drokath:
Now duplicates #918

PR #901, opened 2026-02-18
Fix issue with task retry delay

eamonnotoole:
It looks like the API changes the value of task retry delay set on POST. To prevent errors on task READ we set the value in state to that in plan except on import.

### DIFF
--- a/morpheus/framework/resources/task/read.go
+++ b/morpheus/framework/resources/task/read.go
@@ -161,7 +161,12 @@ func getTaskAsState(
 	state.RetryCount = convert.Int64ToType(task.RetryCount)
 
 	// retry_delay_seconds
-	state.RetryDelaySeconds = convert.Int64ToType(task.RetryDelaySeconds)
+	// on import use the value from the API, otherwise use the value from the plan
+	if plan.Name.IsNull() || plan.Name.IsUnknown() {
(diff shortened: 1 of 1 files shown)


RDavisHPE (approved):

jasonquigleyhpecom (approved):

morpheus/framework/resources/task/read.go
```diff
@@ -161,7 +161,12 @@ func getTaskAsState(
 	state.RetryCount = convert.Int64ToType(task.RetryCount)
 
 	// retry_delay_seconds
-	state.RetryDelaySeconds = convert.Int64ToType(task.RetryDelaySeconds)
+	// on import use the value from the API, otherwise use the value from the plan
+	if plan.Name.IsNull() || plan.Name.IsUnknown() {
+		state.RetryDelaySeconds = convert.Int64ToType(task.RetryDelaySeconds)
+	} else {
+		state.RetryDelaySeconds = plan.RetryDelaySeconds
```
RDavisHPE, line 168: Actually, maybe this check should be for plan.RetryDelaySeconds? I'd imagine both name and the field are null for an import so shouldn't matter too much

eamonnotoole: A standard trick we use to detect an import is to see if the Name is set.  If it isn't, then we know that its an import.

PR #926, opened 2026-03-11
Add cloud_ids to the group resource

jasonquigleyhpecom:
This PR adds a cloud_ids field to the group resource to allow assigning
or removing clouds from the group.


morpheus/framework/resources/group/resource.go
```diff
@@ -77,9 +78,65 @@ func getGroupAsState(
...
+		clouds = append(clouds, cloud)
+	}
+
+	reqGrp := sdk.NewUpdateGroupsZonesRequestGroup(clouds)
+
+	req := sdk.NewUpdateGroupsZonesRequest(*reqGrp)
+
+	_, hresp, err := client.GroupsAPI.UpdateGroupsZones(ctx, id).UpdateGroupsZonesRequest(*req).Execute()
+	if err != nil || hresp.StatusCode != http.StatusOK {
+		return fmt.Errorf("group %d update clouds failed: ", id)
```
matthew-c-hpe: Update this error message, it looks like it's missing something following the `failed: `

Also I'm guessing the update operation must be relatively instantaneous, so the update call itself looks ok to me.

jasonquigleyhpecom: Fixed

matthew-c-hpe (approved):

PR #955, opened 2026-03-25
Add os_type_image resource

jasonquigleyhpecom:
This PR adds the `hpe_morpheus_os_type_image` resource.


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.24.1
 
 require (
 	github.com/HewlettPackard/hpe-morpheus-go-sdk/legacy v0.0.0-20260127104537-7d5e34cd02e9
-	github.com/HewlettPackard/hpe-morpheus-go-sdk/oapigen v0.29.0
+	github.com/HewlettPackard/hpe-morpheus-go-sdk/oapigen v0.31.0
```
matthew-c-hpe: Merge dev-1.3 back into this branch, we're on 0.32.0 now.

jasonquigleyhpecom: Already done locally - I'll be pushing updates later

morpheus/framework/resources/ostypeimage/resource.go
```diff
@@ -0,0 +1,235 @@
...
+		state.CloudId = types.Int64Null()
+	}
+
+	if img.ProvisionType.IsSet() {
+		state.ProvisionTypeId = types.Int64Value(img.GetProvisionType())
+	} else {
+		state.ProvisionTypeId = types.Int64Null()
+	}
+
+	// OsTypeId is not returned by the GET endpoint — caller must set it.
```
matthew-c-hpe: So import is broken? It can't import os_type_id as-is?

We could deduce os_type_id like this:

GET os_type_image -> contains virtual_image_id
GET virtual_images/{virtual_image_id} -> contains osType block, which contains `id` i.e. `os_type_id`.

Assuming that the os_type_image and its associated virtual_image_id map to the same underlying image resource in Morpheus.

jasonquigleyhpecom: Yes, we already discussed this as part of my plan for the data source

matthew-c-hpe (approved):

PR #960, opened 2026-03-27
Generalised Cluster Create, Read, Delete with HVM support

matthew-c-hpe:
Adds generalised cluster resource with support for create, read, and delete.

Basic update support will come in a follow-up PR.

HVM clusters are supported via the `config_hvm` attribute.

Generic clusters are supported via the `config` attribute.

Adds an acceptance test for a HVM cluster.

Adds an acceptance test for a generic cluster (HVM).

Adds documentation and example for HVM cluster.

Adds documentation and example for generic cluster (HVM example is used).

Drive-by: Bumps oapigen sdk version to release tag v0.32.0 - required for cluster features.


go.mod
jasonquigleyhpecom: As `mapstructure` is being bumped above, we should remove this

morpheus/framework/resources/cluster/read.go
```diff
@@ -0,0 +1,272 @@
...
+			dynamicPlacementModeVal = v
+		}
+
+		var powerPolicyVal string
+		if v, ok := cluster.Config["powerPolicy"].(string); ok {
+			powerPolicyVal = v
+		}
+
+		configHvmVal, diags := NewConfigHvmValue(ConfigHvmValue{}.AttributeTypes(ctx), map[string]attr.Value{
+			"create_user":            plan.ConfigHvm.CreateUser,
```
jasonquigleyhpecom: You need to be careful here when importing - plan will likely be an empty struct which will cause nil pointer panics.

Set these fields with the API values on import.

morpheus/framework/resources/cluster/read.go
```diff
@@ -0,0 +1,272 @@
...
+	return state, diags
+}
+
+func buildReadClusterServer(
+	ctx context.Context,
+	plan ClusterModel,
+) (ServerValue, diag.Diagnostics) {
+	return NewServerValue(ServerValue{}.AttributeTypes(ctx), map[string]attr.Value{
+		// the requires replace properties...
+		"data_device":              plan.Server.DataDevice,
```
jasonquigleyhpecom: plan could be an empty `ClusterModel` (importing) so make sure to check values/fields first

jasonquigleyhpecom (changes requested):

morpheus/framework/resources/cluster/read.go
matthew-c-hpe: Import complicates the server block a lot, so for import I've left a TODO note and returned an empty server block. It's still requires replace. so I think this would work in theory. But need to revisit after we do Update.

morpheus/framework/resources/cluster/read.go
matthew-c-hpe: Done, we populate what we can from the API values on import. Might need another look when we support and test import.

morpheus/framework/resources/cluster/read.go
```diff
@@ -0,0 +1,319 @@
...
+			"could not parse a slice of labels",
+		)
+
+		return state, diags
+	}
+
+	// Morpheus API may change the casing of the labels, to avoid Terraform
+	// throwing a gasket we convert the casing of labels to be as specified
+	// by the user.
+	for _, label := range labels {
```
jasonquigleyhpecom, line 246: Is this a general thing or only related to VMware?

jasonquigleyhpecom (approved):

morpheus/framework/resources/cluster/read.go
matthew-c-hpe: Not sure now that you mention it, I assumed it was general until it was suggested to be a VMWare thing in recent weeks. I haven't had to deal with labels before this.

PR #973, opened 2026-04-01
Add tenant_subdomain field

eamonnotoole:
Add tenant_subdomain field to allow for authentication in subtenants when using username/password.


morpheus/testhelpers/client.go
```diff
@@ -20,6 +20,7 @@ func newClient(ctx context.Context, t *testing.T) *sdk.APIClient {
 		os.Getenv("TF_VAR_testacc_morpheus_url"),
 		os.Getenv("TF_VAR_testacc_morpheus_username"),
 		os.Getenv("TF_VAR_testacc_morpheus_password"),
+		os.Getenv("TF_VAR_testacc_morpheus_subdomain"),
```
matthew-c-hpe: nit: let's use `TF_VAR_testacc_morpheus_tenant_subdomain` for this

jasonquigleyhpecom (dismissed):

matthew-c-hpe (changes requested):
Let's document what `tenant_subdomain` is for  by updating the `index.md.tmpl`.

morpheus/testhelpers/client.go
eamonnotoole: Done.

examples/provider/morpheus/provider-tenant.tf
```diff
@@ -0,0 +1,20 @@
+# Copyright 2025-2026 Hewlett Packard Enterprise Development LP
+
+terraform {
+  required_providers {
+    hpe = {
+      source  = "HPE/hpe"
+      version = "= 1.2.0"
```
matthew-c-hpe, line 7: should we document `>=` just in case someone copies the config verbatim if we have a newer version than 1.2.0 out

matthew-c-hpe (approved):

PR #977, opened 2026-04-02
Fix forms delete issue with subtenants

eamonnotoole:
Forms created in a subtenant could not be deleted due to a constraint error.  In this PR before forms delete we do an update to remove all option blocks and field groups, which fixes the issue.

### DIFF
--- a/morpheus/sdkv2/resources/form/form.go
+++ b/morpheus/sdkv2/resources/form/form.go
@@ -1673,7 +1673,11 @@ func resourceFormDelete(ctx context.Context, d *schema.ResourceData, meta any) d
 
 	diagsFromUpdate := resourceFormUpdate(ctx, d, meta)
 	if diagsFromUpdate.HasError() {
-		return diagsFromUpdate
+		logStr := fmt.Sprintf("Error during pre-destroy update of form to remove option types and field groups: %s",
+			diagsFromUpdate)
+		log.Printf(logStr)
+
+		return diag.Errorf(logStr)
 	}
 
 	resp, err := client.DeleteForm(convert.StringToInt64(id), req)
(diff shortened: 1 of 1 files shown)


morpheus/sdkv2/resources/form/form.go
```diff
@@ -1665,6 +1665,17 @@ func resourceFormDelete(ctx context.Context, d *schema.ResourceData, meta any) d
...
 	}
 
+	// to avoid constraint errors on destroy in some circumstances (for instance when running in a sub-tenant)
+	// we'll first do an update to the form to remove all option blocks and field groups before we
+	// attempt to delete the form
+	d.Set("option_type", []map[string]any{})
+	d.Set("field_group", []map[string]any{})
+
+	diagsFromUpdate := resourceFormUpdate(ctx, d, meta)
+	if diagsFromUpdate.HasError() {
```
RDavisHPE, line 1675: _Might_ be worth appending this onto an error along the lines of "Error during pre-destroy Update before deletion" to inform the user it was during an Update call that the deletion failed. However that could also just be TMI so up to you.

RDavisHPE (dismissed):

RDavisHPE (dismissed):
Thanks!

morpheus/sdkv2/resources/form/form.go
```diff
@@ -1665,6 +1665,21 @@ func resourceFormDelete(ctx context.Context, d *schema.ResourceData, meta any) d
...
+		logStr := fmt.Sprintf("Error during pre-destroy update of form to remove option types and field groups: %s",
+			diagsFromUpdate)
+		log.Printf(logStr)
```
RDavisHPE: Actually should we be logging this on top of returning the error?

morpheus/sdkv2/resources/form/form.go
eamonnotoole: Good idea.  See next commit

morpheus/sdkv2/resources/form/form.go
eamonnotoole: Removed, and fixed linter error.

RDavisHPE (dismissed):

RDavisHPE (dismissed):

RDavisHPE (approved):

PR #983, opened 2026-04-02
Register hpe_morpheus_os_type_image resource

matthew-c-hpe:
We forgot to register this in `resources.go` which is causing `make docs` to fail.


matthew-c-hpe:
It was a TF docs issue caused by dev overrides.

PR #986, opened 2026-04-03
Move option_type Schema for forms into a separate function

eamonnotoole:
We do this to avoid unnecessary repetition in the Schema.

### DIFF
--- a/morpheus/sdkv2/resources/form/form.go
+++ b/morpheus/sdkv2/resources/form/form.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"log"
 	"strconv"
+	"strings"
 
 	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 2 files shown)


morpheus/sdkv2/resources/form/form.go
```diff
@@ -589,255 +345,7 @@ func ResourceForm() *schema.Resource {
...
-										Type:        schema.TypeBool,
-										Description: "Whether to export the option type as a tag",
-										Optional:    true,
-										Default:     false,
-									},
-									"display_value_on_details": {
-										Type:        schema.TypeBool,
-										Description: "Display the selected value of the option type on the associated resource's details page",
-										Optional:    true,
-										Default:     false,
```
RDavisHPE, line 688: More of a note: these two seem to have `Default: false` under the field group - but not under the option type. I don't think this is significant but just the only thing I could find in terms of potential discrepencies

RDavisHPE (approved):

morpheus/sdkv2/resources/form/form.go
eamonnotoole: Well spotted Ryan - I'd missed that.  I agree, this doesn't seem to be significant.

PR #992, opened 2026-04-03
Add forms layout and pool config fields

eamonnotoole:
In this PR we add layout and pool config selection fields to the forms Schema.  We add these new fields to type networkManager, along with the existing cloud and group config selector fields.  We add plan-stage validation for the selector fields when using type networkManager.

We've updated the example, test and docs.


RDavisHPE (dismissed):
Nice one

RDavisHPE (approved):

PR #998, opened 2026-04-08
Implementing environment option type for form resource

RDavisHPE:
Implementing environment option type so that the option type is available to users.
- Implemented environment option type, acceptance test and example


morpheus/sdkv2/resources/form/form.go
```diff
@@ -431,6 +431,8 @@ func applyOptionTypeConfigByType(row map[string]any, optionTypeConfig map[string
 		config["customData"] = optionTypeConfig["custom_data"]
 		row["optionList"] = map[string]any{"id": optionTypeConfig["option_list_id"]}
 		row["config"] = config
+	case typeEnvironment:
```
RDavisHPE, line 434: Note that environment does just have a defaultValue field and nothing else unique to the optionType (confirmed through UI)

matthew-c-hpe (approved):

PR #1016, opened 2026-04-10
Implementing tag option type and fixing default_value

RDavisHPE:
Some form options types use `optionType` and others use `config.optionType` (and some use both). This PR sets the relevant fields according to each optionType and also implements the tag option type.

### DIFF
--- a/docs/resources/morpheus_form.md
+++ b/docs/resources/morpheus_form.md
@@ -776,6 +776,38 @@ resource "hpe_morpheus_form" "example" {
 }
 ```
 
+### Tag
+
+```terraform
+resource "hpe_morpheus_form" "example" {
+  name        = "demo"
+  code        = "demo"
+  description = "demo"
+  labels      = ["terraform", "demo"]
+
+  option_type {
+    name                     = "tf tag example"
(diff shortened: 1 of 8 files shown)


docs/resources/morpheus_form.md
```diff
@@ -776,6 +776,38 @@ resource "hpe_morpheus_form" "example" {
...
+  labels      = ["terraform", "demo"]
+
+  option_type {
+    name                     = "tf tag example"
+    code                     = "tag-input"
+    description              = "Terraform tag example"
+    type                     = "tag"
+    field_label              = "Tags"
+    field_name               = "tags"
+    default_value            = jsonencode([{ name = "Sample Name", value = "Sample Value" }])
```
RDavisHPE, line 795: Note - default_value turns out to be a pretty significant field. Each option type with a "code block" default_value seems to have its own JSON format for storing it. Will go back and update each of the examples to use a sensible/informative default_value.

RDavisHPE: image
example for disk

matthew-c-hpe (approved):

PR #1026, opened 2026-04-15
Add morpheus_load_balancer resource

jasonquigleyhpecom:


morpheus/framework/resources/loadbalancer/read.go
```diff
@@ -0,0 +1,290 @@
...
+	case isHAProxy:
+		state.TypeCode = types.StringNull()
+	default:
+		if data.Type != nil && data.Type.Code != nil {
+			state.TypeCode = types.StringValue(*data.Type.Code)
+		}
+	}
+
+	if data.Cloud != nil && data.Cloud.Id != nil {
+		state.CloudId = convert.Int64ToType(data.Cloud.Id)
```
matthew-c-hpe: redundant, convert function handles the nil case. we can get rid of the else.

morpheus/framework/resources/loadbalancer/read.go
```diff
@@ -0,0 +1,290 @@
...
+func isNotFound(hresp *http.Response) bool {
+	return hresp != nil && hresp.StatusCode == http.StatusNotFound
+}
+
+func getLoadBalancerAsState(
+	ctx context.Context,
+	id int64,
+	client *sdk.APIClient,
+	plan LoadBalancerModel,
+) (LoadBalancerModel, bool, diag.Diagnostics) {
```
matthew-c-hpe: why are we returning a bool?

morpheus/framework/resources/loadbalancer/read.go
```diff
@@ -0,0 +1,290 @@
...
+
+		state.ConfigHaproxy = haproxyCfg
+		state.Config = types.DynamicNull()
+	default:
+		state.Config = types.DynamicNull()
+		state.ConfigHaproxy = NewConfigHaproxyValueNull()
+	}
+
+	// Tenants
+	state.Tenants = types.SetNull(TenantsType{
```
matthew-c-hpe: redundant? the default value of state in constructing LoadBalancerModel{} should be null.

morpheus/framework/resources/loadbalancer/read.go
jasonquigleyhpecom: yeah, just this bit: `data.Cloud.Id != nil`

morpheus/framework/resources/loadbalancer/read.go
```diff
@@ -0,0 +1,290 @@
...
+		ObjectType: types.ObjectType{
+			AttrTypes: TenantsValue{}.AttributeTypes(ctx),
+		},
+	})
+	if len(data.Tenants) > 0 {
+		var tenantValues []attr.Value
+		for _, tenant := range data.Tenants {
```
matthew-c-hpe: we have a convert function for Set values. I think this whole if statement can be cleaned up by making use of it. convert.ToSetType.

morpheus/framework/resources/loadbalancer/read.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+func convertResourcePermissions(
+	ctx context.Context,
+	resourcePermission *sdk.GetLoadBalancer200ResponseLoadBalancerResourcePermission,
+) (PermissionsValue, diag.Diagnostics) {
+	var diags diag.Diagnostics
+
+	var groupValues []attr.Value
+	sites, ok := resourcePermission.GetSitesOk()
```
matthew-c-hpe: think we can reduce messy code with convert.ToSetType

morpheus/framework/resources/loadbalancer/resource_test.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+func TestMain(m *testing.M) {
+	systemoverride.ParseFlags()
+	code := m.Run()
+	testhelpers.WriteMergedResults()
+	os.Exit(code)
+}
+
+func newProviderWithError() (tfprotov6.ProviderServer, error) {
```
matthew-c-hpe: use the testhelpers provider constructor instead

morpheus/framework/resources/loadbalancer/read.go
```diff
@@ -0,0 +1,290 @@
...
+			"read load balancer resource",
+			"failed to create client: "+err.Error(),
+		)
+
+		return
+	}
+
+	id := plan.Id.ValueInt64()
+
+	state, notFound, diags := getLoadBalancerAsState(ctx, id, client, plan)
```
matthew-c-hpe: any reason we're checking notFound explicitly? Is this different to the other resources for some API detail?

morpheus/framework/resources/loadbalancer/schema_gen.go
```diff
@@ -0,0 +1,1380 @@
...
+							AttrTypes: TenantsValue{}.AttributeTypes(ctx),
+						},
+					},
+				},
+				Optional:            true,
+				Computed:            true,
+				Description:         "Array of tenant account ids that are allowed access",
+				MarkdownDescription: "Array of tenant account ids that are allowed access",
+			},
+			"type_code": schema.StringAttribute{
```
matthew-c-hpe, line 157: do we wanna rename this to load_balancer_type_code?

jasonquigleyhpecom: it's inside a `load_balancer` resource, so no

morpheus/framework/resources/loadbalancer/read.go
jasonquigleyhpecom: As we're only interested in creating a set of `attr.Value` we can't use `ToSetType` as it doesn't satisfy the `ToObjectValue` interface

morpheus/framework/resources/loadbalancer/resource_test.go
jasonquigleyhpecom: done

morpheus/framework/resources/loadbalancer/read.go
jasonquigleyhpecom: removed

morpheus/framework/resources/loadbalancer/read.go
jasonquigleyhpecom: fixed

morpheus/framework/resources/loadbalancer/read.go
jasonquigleyhpecom: done

morpheus/framework/resources/loadbalancer/read.go
```diff
@@ -0,0 +1,246 @@
...
+	case isHAProxy:
+		haproxyCfg, err := parseHAProxyConfig(ctx, data.GetConfig())
+		if err != nil {
+			return state, fmt.Errorf("failed to parse HAProxy config: %w", err)
+		}
+
+		state.ConfigHaproxy = haproxyCfg
+		state.Config = types.DynamicNull()
+	default:
+		state.Config = types.DynamicNull()
```
matthew-c-hpe: Won't this override `config` if `config` was previously set?
We should be reading all values from resp.Config into state.Config so we can have terraform track changes.

matthew-c-hpe (approved):

PR #1030, opened 2026-04-15
Add ModifyPlan method for instance and instance resize changes

eamonnotoole:
This method calls the Morpheus health API to get the build version of the targeted installation.  And depending on the version we will either trigger a force-replace on changes to:
- network_interfaces
- service_plan_options 
 
Or allow an update by calling the resize API.

We've updated the network_interfaces and child_networks Schema to remove use-state-for-unknown, require-replace and the custom plan modifiers, and add the interface id

We've reworked the update code for network_interfaces to work in a similar way to that for volumes.  It compares the plan and state network_interfaces and will build an resize request that includes information on existing interfaces as well as that for the new interfaces.


morpheus/framework/resources/instance/instance_read.go
```diff
@@ -1145,23 +1145,30 @@ func getStateInterfaces(
...
-	if compareServerInstanceIntfs(intfsFromServer, intfsFromInstance) {
+	// Get []NetworkInterfacesValue from the plan
+	var intfsFromPlan []NetworkInterfacesValue
+	pd := plan.NetworkInterfaces.ElementsAs(ctx, &intfsFromPlan, false)
+	if pd.HasError() {
+		return nil, pd
+	}
+
+	// Compare intfsFromServer against intfsFromPlan, to see if the "shapes" are the same
+	if compareServerPlanIntfs(intfsFromServer, intfsFromPlan) {
```
jasonquigleyhpecom, line 1156: what happens on import?

morpheus/framework/resources/instance/instance_update.go
```diff
@@ -297,56 +303,303 @@ func addNetworkInterfacesToResizeRequest(
...
+			different = true
+		}
+	}
+
+	return child, different
+}
+
+// listsMatch is a generic that compares lists from plan and state to see if they are the same.
+// Returns true if they are, false otherwise.
+func listsMatch[S attr.Value](
```
jasonquigleyhpecom: Move to a package?

eamonnotoole: Sure.

morpheus/framework/resources/instance/instance_read.go
eamonnotoole: It's dealt with earlier in that function

jasonquigleyhpecom (dismissed):

jasonquigleyhpecom (approved):

PR #1038, opened 2026-04-16
osType import error fix

rory-mcevoy-hpe:
This PR aims to fix a bug raised. It makes `code` `computed_optional` instead of `required`.


matthew-c-hpe:
Fixed in https://github.com/HPE/terraform-provider-hpe/pull/1043

PR #1059, opened 2026-04-20
Point instance acceptance tests at zodiac

jasonquigleyhpecom:
The tests that are being performed are checking small changes to the instance.
Those tests are indifferent to the underlying instance type so they can
safely be pointed at Zodiac.

### DIFF
--- a/morpheus/framework/resources/instance/example.tf.tmpl
+++ b/morpheus/framework/resources/instance/example.tf.tmpl
@@ -1,5 +1,5 @@
 data "hpe_morpheus_cloud" "vme_cloud" {
-  name = "HPE Alletra VME"
+  name = "{{.CloudName}}"
 }
 
 data "hpe_morpheus_service_plan" "vme_512mb" {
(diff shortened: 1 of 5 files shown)


morpheus/framework/resources/instance/example.tf.tmpl
```diff
@@ -10,16 +10,16 @@ data "hpe_morpheus_service_plan" "vme_512mb" {
 resource "hpe_morpheus_instance" "example" {
   name             = "{{.Name}}"
   cloud_id         = data.hpe_morpheus_cloud.vme_cloud.id # HPE Alletra VME
```
matthew-c-hpe: Remove the comment after cloud id

matthew-c-hpe (approved):

morpheus/framework/resources/instance/example.tf.tmpl
jasonquigleyhpecom: done

PR #1061, opened 2026-04-21
Refactoring Sweeping and Adding Cloud Sweeper

RDavisHPE:
Refactoring sweeper to allow for self-registration (i.e. to add a sweeper - one just has to create a sweep.go file without modifying the centralized client in `sweep_test.go`) and use generics so that sweepers just have to provide a means of listing their respective resource, checking if a given resource is a test resource and deleting said resource (along with options such as filters and allowed API responses to maintain previous functionalities/checks).

Also modified the Makefile to have an option to specify which resources one actually wants to sweep instead of having to append to the -sweep-run list (which was also a pain-point for adding new sweepers).
```
make sweep SWEEP=hpe_morpheus_cloud
```


morpheus/testhelpers/sweep/sweep.go
matthew-c-hpe, line 1: Can rename this to just `sweep.go`

morpheus/testhelpers/sweep.go
```diff
@@ -0,0 +1,207 @@
...
+				}
+
+				if !allowed {
+					log.Printf("[INFO] Skipping %s (%s)", resourceName, reason)
+
+					continue
+				}
+			}
+
+			log.Printf("[INFO] Sweeping %s", resourceName)
```
matthew-c-hpe, line 127: might as well use this refactor to switch to `slog` ? We can get rid of the hardcoded levels then in the messages

RDavisHPE: Switched there - but noticed that framework uses the standard `log` package so the output looks like this (i.e. [DEBUG] and WARN are mixed):
```
terraform-provider-hpe $ make sweep
go test -v ./morpheus/utils/test/sweep/... -sweep=all 
2026/04/21 11:53:56 [DEBUG] Running Sweepers for region (all):
2026/04/21 11:53:56 [DEBUG] Running Sweeper (hpe_morpheus_cloud) in region (all)
2026/04/21 11:53:56 WARN Cannot create sweep client resource=hpe_morpheus_cloud error="TF_VAR_testacc_morpheus_url not set"
(...)
```

morpheus/testhelpers/sweep/sweep.go
```diff
@@ -0,0 +1,207 @@
...
+
+			return fmt.Errorf("failed to list %s: %s", resourceName, errfmt.ErrMsg(err, hresp))
+		}
+
+		if hresp == nil || hresp.StatusCode != http.StatusOK {
+			return fmt.Errorf("failed to list %s: %s", resourceName, errfmt.ErrMsg(err, hresp))
+		}
+
+		var sweptCount int
+		var sweepErrors []string
```
matthew-c-hpe: Might be more idiomatic to have sweeperrors be of type error and use `errors.Join` to append to them? Eliminates the need for SweepErrorsToError then.

morpheus/testhelpers/sweep/sweep.go
```diff
@@ -0,0 +1,207 @@
...
+		"",
+		token,
+		opts...,
+	)
+
+	return client, nil
+}
+
+// RecoverSweepPanic recovers from panics during sweep execution and converts them to errors.
+func RecoverSweepPanic(resourceName string, retErr *error) {
```
matthew-c-hpe: Can be private.

morpheus/testhelpers/sweep.go
matthew-c-hpe: we'll go back to pkg `log`

morpheus/testhelpers/sweep/sweep.go
RDavisHPE: Switched to errors joining

morpheus/testhelpers/sweep/sweep.go
RDavisHPE: Privatised everything except RegisterTypedAPISweeper, callback functions and options

matthew-c-hpe (approved):

jasonquigleyhpecom (approved):

PR #1071, opened 2026-04-23
Simplifying sweeper code to just check for TestAccMorpheus prefix

RDavisHPE:
Removing additional checks to sweepers and removing resources based on the TestAccMorpheus prefix to avoid dangling resources if
- The resource doesn't have required tags
- The resource has a slightly different prefix (i.e. TestAccMorpheusPolicyInstance for an instance that is created for a policy test).
- The resource is a datastore (we previously skipped this entirely)


morpheus/framework/resources/cloud/sweep/sweep.go
```diff
@@ -13,7 +13,7 @@ import (
 	testsweep "github.com/HPE/terraform-provider-hpe/morpheus/testhelpers/sweep"
 )
 
-const testCloudPrefix = "TestAccMorpheusCloud"
+const testResourcePrefix = "TestAccMorpheus"
```
jasonquigleyhpecom: Would it be better to expose this via the testsweep package?

RDavisHPE: Personally - I think it might be better to keep the prefix with the implementation itself just to emphasize that this file is what makes the decision of "is this sweepable" - even if it is just a prefix constant. It might also reduce the amount of context switching Claude has to do - but that might be a bit of a stretch.

Still I'll switch over if you think it's better

RDavisHPE: Switched to the testsweep package

jasonquigleyhpecom (approved):

PR #1078, opened 2026-04-23
network firewall rule group resource

rory-mcevoy-hpe:
This PR aims to add the `networkfirewallrulegroup` resource.


docs/resources/morpheus_network_firewall_rule_group.md
matthew-c-hpe, line 1: Add a`.md.tmpl` and render this from it.

morpheus/framework/resources/networkfirewallrulegroup/schema_gen.go
```diff
@@ -0,0 +1,79 @@
...
+		Attributes: map[string]schema.Attribute{
+			"description": schema.StringAttribute{
+				Optional:            true,
+				Computed:            true,
+				Description:         "Network firewall rule group description",
+				MarkdownDescription: "Network firewall rule group description",
+			},
+			"external_type": schema.StringAttribute{
+				Required:            true,
+				WriteOnly:           true,
```
jasonquigleyhpecom: Remove the write only flag

eamonnotoole (commented):
## Code Review: Network firewall rule group resource

17 files, +1090 lines. CRUD resource for `hpe_morpheus_network_firewall_rule_group` with write-only `external_type`, import, sweeper, and tests.

### 🔴 Delete does not handle 404

Same issue as the firewall rule PR. If the resource is deleted out-of-band, Delete will error instead of succeeding silently:

```go
_, httpResp, err := client.NetworksAPI.
    DeleteNetworkFirewallRuleGroup(ctx, id, serverID).Execute()
if err != nil || httpResp.StatusCode != http.StatusOK {
    resp.Diagnostics.AddError(...)  // 404 hits this path
}
```

Add a 404 guard before the error.

### 🔴 Read does not handle 404

`getNetworkFirewallRuleGroupAsState` errors on any non-200 response. Should detect 404 and call `resp.State.RemoveResource(ctx)` so Terraform can recreate the resource.

### 🟡 Import uses `strconv.Atoi` (int) instead of `strconv.ParseInt` (int64)

```go
serverID, err := strconv.Atoi(parts[0])  // returns int, not int64
id, err := strconv.Atoi(parts[1])
```

On 32-bit systems `Atoi` could overflow for large IDs. Use `strconv.ParseInt(parts[0], 10, 64)` for safety and consistency with the firewall rule resource. The values are also set as raw `int` via `SetAttribute` — should use `types.Int64Value()`.

```markdown
description: |-

---
# hpe_morpheus_network_firewall_rule_group (Resource)
```

The description is blank in both the frontmatter and after the heading. Should describe what the resource manages.

### 🟡 No `CheckDestroy` in tests

Unlike the firewall rule tests, these tests don't have a `CheckDestroy` function to verify the resource was actually deleted after test teardown. This means leaked resources would go undetected (beyond the sweeper).

### 🟡 Sweep uses hardcoded test prefix

```go
const testFirewallRuleGroupPrefix = "TestAccMorpheusNetworkFirewallRuleGroup"
```

Consider using `testsweep.TestResourcePrefix` for consistency. If the prefix convention changes, this would be missed.

### ✅ Positives

- **`WriteOnly: true` for `external_type`** — good use of TF 1.11+ feature for a create-only parameter that shouldn't persist in state
- **`ReadFromConfig` pattern** — correctly reads write-only attributes from config (not plan) in Create
- **Description null handling in Update** — explicitly sends `nil` when removed, which is better than the firewall rule resource
- **`RequiresReplace`** on `external_type`, `group_layer`, and `network_integration_id` — correct immutability semantics
- **Taint-on-failure** in Create with good error messages
- **PlanOnly drift checks** in tests
- **Sweep uses `RegisterTypedAPISweeper`** — follows the shared pattern (improvement over the firewall rule sweep)
- **RequiresReplace test** — explicitly verifies that `group_layer` changes trigger replacement
- **Import test correctly ignores `external_type`** — since it's write-only

### Summary

The 404 handling gaps (Delete and Read) are the same issues flagged on the firewall rule PR and are the most important to fix. The `Atoi` vs `ParseInt` is a portability concern. The rest are consistency improvements.

morpheus/framework/resources/networkfirewallrulegroup/resource_test.go
```diff
@@ -0,0 +1,277 @@
...
+			"100",
+		),
+		resource.TestCheckResourceAttrSet(
+			"hpe_morpheus_network_firewall_rule_group.example",
+			"id",
+		),
+		resource.TestCheckResourceAttrSet(
+			"hpe_morpheus_network_firewall_rule_group.example",
+			"network_integration_id",
+		),
```
jasonquigleyhpecom, line 74: Missing test against `external_type`

jasonquigleyhpecom (dismissed):
## Code Review: Network firewall rule group resource (post-fixes)

17 files, ~1080 lines — new CRUD resource for \`hpe_morpheus_network_firewall_rule_group\`

### All Previously Raised Issues — Resolved

| Issue | Status |
|---|---|
| Delete 404 handling | Fixed (treats 404 as success) |
| Read 404 handling | Fixed (removes resource from state) |
| \`strconv.Atoi\` → \`ParseInt\` | Fixed (uses \`ParseInt(..., 10, 64)\`) |
| Missing \`.md.tmpl\` | Fixed (template present) |
| Empty docs description | Fixed — now reads "Manages a network firewall rule group resource in Morpheus." |
| \`WriteOnly: true\` on \`external_type\` | Fixed — removed; reads from plan, not config |
| WriteOnly config workaround in Create | Fixed — removed \`req.Config\` read; uses \`plan.ExternalType\` directly |
| Priority null-clearing inconsistency | Fixed — Update now sends \`nil\` to clear priority, matching description pattern |
| Missing \`external_type\` test | Fixed (check present in both Example and Update tests) |
| Spurious \`ostype\` import alias | Fixed — reverted to bare import |
| Copyright year 2025 | Fixed — all new files now use 2026 |
| WriteOnly references in docs | Fixed — NOTE block and \`[Write-only]\` annotation removed |

### Code Quality Assessment

| Area | Verdict |
|---|---|
| File structure | Follows convention (one file per CRUD op) |
| Error handling | Proper \`errfmt.ErrMsg()\`, 404 guards in Read and Delete |
| Taint on failure | \`cleanup.TaintResourceState\` in Create |
| Import | Composite \`network_integration_id:id\` with \`ParseInt\` + \`types.Int64Value\` |
| Update semantics | Consistent null-clearing for \`description\` and \`priority\`; immutable fields (\`external_type\`, \`group_layer\`, \`network_integration_id\`) use \`RequiresReplace\` |
| Sweeper | \`RegisterTypedAPISweeper\`, iterates all network servers, filters by \`testsweep.TestResourcePrefix\` |
| Tests | 3 acceptance tests (create+import, update+drift, requires-replace), all parallel with \`RecordResult\` |
| \`resources.go\` | Clean import, no unnecessary aliases |

### Remaining Non-blocking Observations

1. **No \`CheckDestroy\` in tests** — mitigated by sweepers, and consistent with the author's stated intent. Not a blocker.
2. **Import still ignores \`external_type\`** (\`ImportStateVerifyIgnore: []string{"external_type"}\`) — this is correct because the API doesn't return \`external_type\` in the GET response, so after import+read it will be null. The field is preserved from prior state during normal operations.

### Verdict

All previously identified issues have been addressed. The resource is clean, well-structured, and follows established codebase patterns. Ready to merge.

jasonquigleyhpecom (dismissed):
## Re-Review (post make docs)

Confirmed `make docs` generates cleanly with no diff. The schema-level `Description` field now drives the doc template correctly.

### Final State

- All 17 files are clean
- `go build ./...` passes
- `go vet ./...` passes
- `make docs` produces no uncommitted changes
- No blocking issues remain

LGTM. Ready to merge.

eamonnotoole (approved):

PR #1081, opened 2026-04-23
Add consts for TF_VAR_testacc* env vars

matthew-c-hpe:
We may want to have these as consts we can access from a package to avoid mistakes.

Might also be useful to have these for certain tests.


eamonnotoole (dismissed):

matthew-c-hpe:
Moving to draft for now. I didn't consider our test redirection.

PR #1083, opened 2026-04-24
network firewall rule group datasource

rory-mcevoy-hpe:
This PR aims to add the `networkfirewallrulegroup` datasource.


docs/data-sources/morpheus_network_firewall_rule_group.md
matthew-c-hpe, line 1: Make a .md.tmpl and render the examples correctly.

morpheus/framework/datasources/networkfirewallrulegroup/schema_gen.go
```diff
@@ -0,0 +1,4560 @@
...
+					},
+					CustomType: RulesType{
+						ObjectType: types.ObjectType{
+							AttrTypes: RulesValue{}.AttributeTypes(ctx),
+						},
+					},
+				},
+				Computed: true,
+			},
+			"server_id": schema.Int64Attribute{
```
matthew-c-hpe: The UI and documentation both call these Network Integrations, we should use the same naming convention. Either 'network_integration_id' or 'integration_id'.

morpheus/framework/datasources/networkfirewallrulegroup/datasource.go
```diff
@@ -0,0 +1,397 @@
...
+			"GET failed for firewall rule groups on server %d: %s",
+			serverID, errfmt.ErrMsg(err, hresp),
+		)
+	}
+
+	// The list endpoint returns interface{} because the OpenAPI spec does
+	// not define a typed array for ruleGroups. We must assert manually.
+	ruleGroups := r.GetRuleGroups()
+
+	ruleGroupsList, ok := ruleGroups.([]interface{})
```
matthew-c-hpe: Update the OpenAPI spec for this endpoint, I think this has generated without a type. GetRuleGroups should return a strongly typed struct.

It looks like the `ruleGroups` array is documented without type information about the array itself.

Add this:
`morpheus-openapi/paths/api@networks@servers@id@firewall-rule-groups.yaml`
```
              properties:
                ruleGroups:
                  type: array -- THIS
```

eamonnotoole (commented):
## Code Review: Network firewall rule group data source

13 files, +5372 lines. Data source for `hpe_morpheus_network_firewall_rule_group` with lookup by ID or name, returning rules with nested applications, destinations, sources, scopes, and profiles.

### 🟡 SDK legacy version downgrade

`go.mod` changes the legacy SDK from `20260506` (May 6) to `20260410` (April 10) — this is a **downgrade**. The oapigen SDK goes from `0.43.0` to `0.44.0` (correct upgrade). The legacy downgrade could cause merge conflicts or regressions if other branches depend on the newer version. Is this intentional?

### 🟡 Multiple-found error doesn't include IDs

In `getRuleGroupByName`, when multiple rule groups match:
```go
if matchCount > 1 {
    return nil, errors.New(ErrorMultipleFound)
}
```

The load balancer monitor data source (PR #1118) includes matching IDs in the error to help users resolve ambiguity:
```
"multiple load balancer monitors found with name %s. IDs: %s. Please specify an ID instead"
```

Consider doing the same here.

Same pattern as the other firewall PRs — the description is blank in the frontmatter and after the heading.

### 🟡 `AtLeastOneOf` on both `id` and `name`

`AtLeastOneOf` is declared on both attributes. It only needs to be on one (Terraform evaluates it globally). Harmless but redundant — same pattern we flagged on PR #1118 and was subsequently fixed.

### 🟡 Hardcoded test data

Tests use hardcoded values (`NetworkIntegrationId: "128"`, `Id: "233264"`, `Name: "dg-test-policy"`) that are environment-specific. These will fail in any environment without those exact resources. Consider either using `systemoverride` consistently or documenting the test data requirements.

### ✅ Positives

- **Generic helper functions** — `mapIDNameSet` and `mapEmptyObjectSet` are clean, reusable, and well-typed with Go generics. The `idNameExtractor` type alias keeps extract function signatures readable.
- **Rich nested schema** — `rules` set includes applications, destinations, sources, scopes, profiles, applied_targets, config, and rule_group. Comprehensive coverage.
- **Schema validators** — `AtLeastOneOf` + `ConflictsWith` on `id`/`name` ensures exactly one is provided at plan time.
- **Thorough test coverage** — positive paths (by ID, by name), negative paths (not found, no search attrs, both search attrs). The plan-time validation test uses an offline provider config, which is a nice touch.
- **Name lookup re-fetches by ID** — after finding the match in the list, it re-fetches with the single-object GET to get full details. Good pattern.
- **`convert.ToSetType` integration** — cleanly converts SDK response slices to TF set types.

### Summary

The SDK version downgrade is the most concerning item and needs clarification. The rest are quality-of-life improvements. The data source implementation is well-structured with good use of generics and comprehensive test coverage.

eamonnotoole (approved):

PR #1089, opened 2026-04-27
Add network_domain_id to instance

eamonnotoole:
In this PR we add the network_domain_id attribute to instance to allow the specification of the network_domain when creating an instance. Update of this domain isn't supported for instance, so a change will result in a recreate.

We've added support for network_domain_id on import and added the setting to one of the examples.  We've updated the docs.


morpheus/framework/resources/instance/instance_read.go
```diff
@@ -254,6 +254,14 @@ func getInstanceAsState(
 
 	state.NetworkInterfaces = networkInterfacesList
 
+	// network_domain_id
+	networkDomainId, ndiags := getInstanceNetworkId(instance, plan)
```
jasonquigleyhpecom: Should `getInstanceNetworkId` be `getInstanceNetworkDomainId`?

eamonnotoole: Yeah, I'll change that.

jasonquigleyhpecom (approved):

PR #1094, opened 2026-04-29
Firewall rule resource

rory-mcevoy-hpe:
This PR aims to add the `networkfirewallrule` resource.


morpheus/framework/resources/networkfirewallrule/schema_gen.go
```diff
@@ -0,0 +1,2095 @@
...
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+)
+
+func NetworkFirewallRuleResourceSchema(ctx context.Context) schema.Schema {
+	return schema.Schema{
+		Attributes: map[string]schema.Attribute{
+			"config": schema.SingleNestedAttribute{
```
matthew-c-hpe: Double check the API that this is a correct shape for config. I would be surprised if it is as it goes against the pattern for every other `config` block behaviour in Morpheus. Remember that what may have been generated for the provider code spec could have been as a result of what somebody documented in the OpenAPI spec for a GET on a Firewall Rule that was already created.

morpheus/framework/resources/networkfirewallrule/delete.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil || (httpResp != nil && httpResp.StatusCode != http.StatusOK) {
+		resp.Diagnostics.AddError(
+			"error deleting network firewall rule",
+			fmt.Sprintf("network firewall rule %d DELETE failed: ", id)+errfmt.ErrMsg(err, httpResp),
+		)
+
+		return
+	}
+
+	waitForDeleted := func() (struct{}, error) {
```
matthew-c-hpe, line 55: Regarding the function signature and return values - this is fine, but just be mindful that you understand the language details when doing tricks like this.

eamonnotoole (commented):
## Code Review: Firewall rule resource

16 files, +3155 lines. CRUD resource for `hpe_morpheus_network_firewall_rule` with import, sweeper, and comprehensive tests.

### 🔴 Delete does not handle 404 (already-deleted resource)

In `delete.go`, if the resource was already deleted out-of-band, the DELETE call returns 404 and the code errors instead of succeeding silently:

```go
_, httpResp, err := client.NetworksAPI.
    DeleteNetworkFirewallRule(ctx, id, serverId).Execute()
if err != nil || (httpResp != nil && httpResp.StatusCode != http.StatusOK) {
    resp.Diagnostics.AddError(...)  // 404 hits this path
    return
}
```

Add a 404 guard before the error (same pattern as the virtual server resource):
```go
if httpResp != nil && httpResp.StatusCode == http.StatusNotFound {
    return
}
```

### 🔴 Read does not handle 404 (resource deleted out-of-band)

In `read.go`, `getNetworkFirewallRuleAsState` errors on any non-200 response. For proper Terraform behavior, Read should detect 404 and call `resp.State.RemoveResource(ctx)` so Terraform knows the resource was deleted externally and can recreate it on next apply.

### 🟡 Documentation shows `config` nested attribute that does not exist in schema

The generated docs at `docs/resources/morpheus_network_firewall_rule.md` show:
```
- `config` (Attributes) Additional configuration...
  - `application` (Set of String)
  - `profile` (Set of String)
```

But the actual schema in `schema_gen.go` has `application` and `profile` as **top-level** attributes with no `config` wrapper. The docs appear to have been hand-edited. They need to be regenerated to match the actual schema, otherwise users will write incorrect HCL.

```markdown
description: |-

---
# hpe_morpheus_network_firewall_rule (Resource)
```

The description is blank in the frontmatter and after the heading. Should describe what the resource manages.

### 🟡 Delete timeout of 45 minutes is excessive

```go
backoff.WithMaxElapsedTime(45 * time.Minute)
```

Firewall rule deletions are typically fast (seconds to low minutes). 45 minutes will leave users stuck if something goes wrong. Consider 5-10 minutes.

### 🟡 Import uses raw int64 instead of types.Int64Value

In `import.go`:
```go
resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("network_integration_id"), serverId)...)
resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("id"), ruleId)...)
```

Should use `types.Int64Value(serverId)` and `types.Int64Value(ruleId)` for consistency with other resources.

### 🟡 Update uses untyped `map[string]interface{}` instead of SDK types

`create.go` correctly uses SDK typed structs (`sdk.NewCreateNetworkFirewallRuleRequestRule`, etc.), but `update.go` builds a raw `map[string]interface{}`. This loses compile-time type safety. If the SDK has typed update request objects, prefer those.

### 🟡 Sweep diverges from the shared pattern

The sweep creates a custom `newSweepClient` and uses a JSON marshal/unmarshal round-trip to extract rule data. Other resources (e.g., virtual server) use `testsweep.RegisterTypedAPISweeper`. Also uses a hardcoded `testPrefix` instead of `testsweep.TestResourcePrefix`.

### ✅ Positives

- **Generic `extractStringIDs`** — clean use of Go generics for mapping response collections
- **Taint-on-failure in Create** — resource state is tainted on post-create read failure for clean recovery
- **Delete confirmation polling** — properly waits for the resource to actually be gone
- **Thorough tests** — CRUD, import, nested attributes, bad-ID error paths, and PlanOnly drift checks
- **Import error handling** — clear messages for invalid format and non-numeric IDs

### Summary

The 404 handling gaps in Delete and Read are the most important items — they'll cause issues when resources are deleted out-of-band. The doc/schema mismatch will confuse users. The rest are consistency improvements.

eamonnotoole (approved):

PR #1097, opened 2026-05-01
Add hpe_morpheus_load_balancer_virtual_server resource

eamonnotoole:
This will need extensive testing and probably modification before release.  We've added a static config_nsxt in addition to a generic config.


eamonnotoole (commented):
## Code Review: Add hpe_morpheus_load_balancer_virtual_server resource

19 files, +3274 lines. Full CRUD resource with NSX-T typed config, generic dynamic config fallback, import support, sweeper, and comprehensive tests.

### 🟡 Bug-risk: Inconsistent type in import.go

In `import.go`, `load_balancer_id` is set with `types.Int64Value(lbID)` but `id` is set as a raw `int64`:

```go
resp.Diagnostics.Append(resp.State.SetAttribute(
    ctx, path.Root("load_balancer_id"),
    types.Int64Value(lbID),  // ✅ types.Int64Value
)...)
resp.Diagnostics.Append(resp.State.SetAttribute(
    ctx, path.Root("id"), vsID,  // ⚠️ raw int64
)...)
```

The TF framework accepts both, so this works, but it should be `types.Int64Value(vsID)` for consistency.

### 🟡 Missing `systemoverride.ParseFlags()` in TestMain

`resource_test.go` `TestMain` does not call `systemoverride.ParseFlags()`:

```go
func TestMain(m *testing.M) {
    code := m.Run()  // missing systemoverride.ParseFlags() before this
    testhelpers.WriteMergedResults()
    os.Exit(code)
}
```

Other test files in the repo (e.g., the load balancer monitor data-source) include it. This could affect test configuration when system overrides are needed.

### 🟡 Inconsistent LB name truncation in tests

`resource_test.go` uses prefix+suffix to stay unique within 32 chars:
```go
lbName = lbName[0:16] + lbName[len(lbName)-16:]
```

`resource_nsxt_test.go` simply truncates:
```go
lbName = lbName[0:min(32, len(lbName))]
```

The first approach is better for uniqueness — consider using it consistently.

### 🟡 Import detection heuristic

Using `current.VipName.IsUnknown()` to detect import vs normal read is a pragmatic approach. Just be aware this relies on VipName being Unknown during import (because only `id` and `load_balancer_id` are populated). If the schema changes (e.g., a default value is added), this could silently break. A comment documenting this assumption would help.

### ✅ Positives

- **Dual config strategy** — typed `config_nsxt` for known LB types + dynamic `config` escape hatch is a great pattern. `ConflictsWith` prevents using both.
- **RequiresReplace on config and load_balancer_id** — correct semantics; config changes are fundamental to the VS identity
- **Delete handles 404 gracefully** — treats already-deleted resources as success
- **Sweeper** — properly iterates all LBs to find test virtual servers, uses `TestResourcePrefix` filtering
- **Excellent test coverage** — basic CRUD, NSX-T CRUD, in-place update, replace-on-config-change, and import tests with ID verification
- **vip_protocol validator** — `OneOf("http", "https", "udp", "tcp")` catches invalid values at plan time
- **State preservation for write-only fields** — `SslServerCert`, `Config`, `ConfigNsxt` correctly preserved from plan since GET doesn't return them
- **Import doc + format** — clear `1.42` format documented with working example

### Summary

Well-structured resource implementation. The PR description correctly notes this needs extensive testing before release. The issues above are minor — the import type inconsistency and missing `ParseFlags()` are the most actionable.

eamonnotoole (commented):
## Re-review after "Fix review comments"

All four items addressed:

1. ✅ **Import type consistency** — `vsID` now wrapped in `types.Int64Value()`
2. ✅ **Missing `systemoverride.ParseFlags()`** — added to `resource_test.go` `TestMain`
3. ✅ **LB name truncation** — NSX-T tests now use the same prefix+suffix strategy
4. ✅ **Import detection heuristic** — thorough comment added documenting the assumption and breakage risk

No further comments. **LGTM.**

jasonquigleyhpecom (approved):

PR #1099, opened 2026-05-01
Fix load balancer sweeper to remove underlying HAProxy containers

jasonquigleyhpecom:
This PR works around a bug in the API that leaves dangling container instances when a HAProxy load balancer is deleted. As there are no direct endpoints in the API for listing Docker containers, we use the search API and match returned items with a precise regular expression to avoid deleting unrelated resources.


morpheus/framework/resources/loadbalancer/sweep/sweep.go
```diff
@@ -55,3 +69,106 @@ func init() {
...
+			sweptCnt += swept
+			errCnt += errored
+		}
+
+		meta, ok := resp.GetMetaOk()
+		if !ok || meta == nil {
+			break
+		}
+
+		offset += int64(len(hits))
```
RDavisHPE, line 152: One thing - if we're doing pagination and offset - is there a possibility that resources could get skipped?
i.e.
- Search page size is 3 (for example)
- Initial results (by index): [A, B, C, D, E, F].
- First request uses offset 0, gets [A, B, C].
- Sweeper deletes A, B, C.
- Remaining list is now [D, E, F].
- Sweeper increases offset to 3 and requests next page.
- But there are only 3 items left, so offset 3 returns empty.
- D, E, F were never processed.

If so, might be worth gathering the IDs first and then deleting afterwards (or perhaps the pageSize is big enough for this not to be too bad of a problem)

morpheus/framework/resources/loadbalancer/sweep/sweep.go
```diff
@@ -55,3 +69,106 @@ func init() {
...
+
+	const pageSize int64 = 100
+
+	for {
+		resp, hresp, err := client.SearchAPI.Search(ctx).
+			Phrase(testsweep.TestResourcePrefix).
+			Max(pageSize).
+			Offset(offset).
+			Execute()
+		if err != nil {
```
RDavisHPE: Should we check resp != nil here?

morpheus/framework/resources/loadbalancer/sweep/sweep.go
jasonquigleyhpecom: yes, you're absolutely right - I'll fix that (see what I did there?)

morpheus/framework/resources/loadbalancer/sweep/sweep.go
jasonquigleyhpecom: yep

jasonquigleyhpecom: Fixed

morpheus/framework/resources/loadbalancer/sweep/sweep.go
jasonquigleyhpecom: Fixed

RDavisHPE (approved):

PR #1115, opened 2026-05-12
Add load balancer monitor resource

eamonnotoole:
We've added examples and information on NSX-T and NSX-V.  This PR
updates the loadbalancer example Render stub functions.


eamonnotoole (commented):
## Code Review — PR #1115: Add load balancer monitor resource

Overall this is a well-structured, well-documented new resource that follows the existing provider patterns. The CRUD operations, import, sweep, tests, and docs are all present and consistent. A few items worth discussing:

---

### Bugs / Potential Issues

**1. Delete polling never terminates on success (`resource_delete.go`)**
The `waitForDeleted` function returns `nil` error on *every* path except permanent errors, meaning it will keep polling for 45 minutes even after the resource is confirmed deleted (404). The `nil` error tells backoff "keep retrying". You need to return a sentinel/permanent success when you get 404, and return a retryable error while the resource still exists (non-404):

```go
func() (struct{}, error) {
    _, httpResp, err := client.LoadBalancersAPI.
        GetLoadBalancerMonitor(ctx, loadBalancerID, id).Execute()
    if httpResp != nil && httpResp.StatusCode == http.StatusNotFound {
        return struct{}{}, nil // deleted — stop polling
    }
    if err != nil {
        return struct{}{}, backoff.Permanent(err) // unexpected error — abort
    }
    return struct{}{}, fmt.Errorf("monitor %d still exists", id) // still alive — retry
}
```

**2. Update does not validate/map `monitor_type` (`resource_update.go`)**
Create validates `monitor_type` against the LB type and maps canonical → NSX-T profile names. Update sends `plan.MonitorType` verbatim without any mapping. If a user changes `monitor_type` on an NSX-T LB, the update sends the canonical name (e.g. `"http"`) instead of the NSX-T profile (`"LBHttpMonitorProfile"`). Consider extracting the type-mapping logic and reusing it in Update.

---

### Design Questions

**3. NSX-T example uses raw profile name, docs say canonical**
The schema docs say `monitor_type` accepts canonical values (`http`, `https`, etc.) and the provider maps them. But the NSX-T example/template hardcodes `LBHttpMonitorProfile`, and the test checks for `LBHttpMonitorProfile` in state. This is confusing — should the example show the canonical `"http"` and let the provider map it?

**4. `monitor_username` has `RequiresReplace` — intentional?**
The schema marks `monitor_username` with `stringplanmodifier.RequiresReplace()`, meaning any change to the username destroys and recreates the monitor. This seems surprising for a mutable field. Is this an API limitation, or should it be updatable in-place?

---

### Minor / Nits

**5. `partition` attribute has no description**
The schema defines `partition` as Optional+Computed but with no description or markdown. The TODO comment in create explains why it is not sent, but the attribute is still exposed. Consider either adding a description or removing it from the schema if it is truly not user-settable.

**6. Redundant `fmt` import in `loadbalancer_example.go`**
Both `RenderLoadBalancerNsxtConfig` and `RenderLoadBalancerNsxvConfig` use `fmt.Sprintf` but `fmt` was presumably already imported. Just confirming it compiles clean — no action needed if so.

**7. Template variable naming: `MonitorPassword` vs `MonitorPasswordWo`**
The NSX-V template uses `{{.MonitorPassword}}` as the variable name but the TF attribute is `monitor_password_wo`. The go:generate line also uses `MonitorPassword ""`. This works but could confuse someone reading the template vs the schema. Consider aligning to `MonitorPasswordWo` for consistency.

---

Items 1 and 2 look like real bugs that need fixing before merge. The rest are discussable. Nice work overall — comprehensive tests and good documentation.

eamonnotoole (commented):
## Re-review after "Fix review comments" (1a77658)

Both bugs from my prior review are fixed:

✅ **Delete polling** — `waitForDeleted` now correctly returns a retryable error (`"monitor %d still exists"`) when the resource is still present, and `nil` on 404. Looks good.

✅ **Update monitor_type mapping** — Update now performs the full LB type lookup and canonical→profile mapping, matching Create. The validation/error messages are consistent too.

The remaining items from the prior review (NSX-T example using raw profile name, `monitor_username` RequiresReplace, `partition` lacking a description, template variable naming) were design questions — fine to address separately or leave as-is if intentional.

LGTM 👍

eamonnotoole (commented):
## Re-review after "Fix review comments 2" (638627d)

The latest commit addresses two of the remaining design items nicely:

✅ **`partition` removed from docs** — no longer appears in the schema documentation  
✅ **Template variable naming aligned** — `MonitorPassword` → `MonitorPasswordWo` in example.go defaults, go:generate line, and NSX-V template  

---

### New bug: perpetual drift for NSX-T `monitor_type`

The NSX-T example now correctly uses the canonical `"http"` (changed from `"LBHttpMonitorProfile"` in a previous fixup). However, **Create and Update don't preserve the canonical `monitor_type` from the plan in state**. They store whatever `getLoadBalancerMonitorAsState` reads from the API, which is the NSX-T profile name `"LBHttpMonitorProfile"`.

This creates a perpetual diff:

1. User writes `monitor_type = "http"` in config
2. Create maps `"http"` → `"LBHttpMonitorProfile"` and sends to API  
3. `getLoadBalancerMonitorAsState` reads back `"LBHttpMonitorProfile"` into state
4. Next plan: config `"http"` ≠ state `"LBHttpMonitorProfile"` → drift detected every time

The `Read` function already handles this correctly by preserving `data.MonitorType` from prior state, but Create and Update don't do the same.

**Fix**: Add `state.MonitorType = plan.MonitorType` after `getLoadBalancerMonitorAsState` in both `resource_create.go` and `resource_update.go`, alongside the existing preservation of `ExtraConfig` and `Config`:

```go
// Preserve canonical monitor_type from plan to avoid drift.
state.MonitorType = plan.MonitorType
```

The test assertion on line 72 of `resource_test.go` also needs updating to match:

```go
// Before:
resource.TestCheckResourceAttr(resourceName, "monitor_type", "LBHttpMonitorProfile"),
// After:
resource.TestCheckResourceAttr(resourceName, "monitor_type", "http"),
```

Without this fix, `ExpectNonEmptyPlan: false` in the NSX-T acceptance tests will fail due to the config/state mismatch.

eamonnotoole (commented):
## Re-review after "Fix review comments 3" (9553783)

✅ **NSX-T `monitor_type` drift fixed** — Create and Update now preserve the canonical `monitor_type` from the plan, and the test assertion updated to expect `"http"`.

All previously raised issues are resolved. LGTM 👍

jasonquigleyhpecom (approved):

PR #1118, opened 2026-05-13
Add load balancer monitor data-source

eamonnotoole:


eamonnotoole (commented):
## Code Review

### 🟡 Improvement: Add `AtLeastOneOf` schema validator
**`datasource.go`** — `getLoadBalancerMonitor` returns a runtime error `"either id or name must be specified"`, but there is no schema-level `AtLeastOneOf` validator on `id`/`name`. Adding one would surface the error at `terraform plan` time rather than during read, giving a better UX. The `ConflictsWith` validators are already in place — just missing the complementary `AtLeastOneOf`.

### 🟡 Nit: Redundant import alias
**`datasources.go`** — The alias `loadbalancermonitor "github.com/...loadbalancermonitor"` is redundant since the package is already named `loadbalancermonitor`. The adjacent `loadbalancer` import does not use an alias. Minor consistency nit.

### ✅ Positives
- **Name lookup** correctly does an exact match filter client-side and errors clearly on 0 or >1 matches with IDs listed
- **Re-fetches by ID** after name lookup to get the full single-object response — good pattern
- **Config nesting** properly handles null at every level (config, monitor, monitor_config)
- **StatusDate** uses RFC3339 formatting
- **Tests** cover both lookup-by-ID and lookup-by-name paths with real infrastructure
- **Docs & examples** are thorough, including monitor type tables for NSX-T and NSX-V

### ✅ No bugs found
The implementation looks correct and complete.

eamonnotoole (commented):
## Re-review after "Fix review comments 1"

The `AtLeastOneOf` validator was added to both `id` and `name` — this addresses the schema-level validation feedback. Users will now get a clear error at plan time if neither is specified. ✅

### One minor note
`AtLeastOneOf` only needs to be declared on **one** of the two attributes (Terraform evaluates it globally). Having it on both is harmless but redundant. Not a blocker.

### Remaining nit (not a blocker)
The redundant import alias in `datasources.go` is still there — purely cosmetic.

**LGTM** — Clean, well-structured data source with good docs and test coverage.

eamonnotoole (commented):
## Re-review after "Fix review comments 2"

Both remaining notes addressed:

1. ✅ Redundant import alias removed in `datasources.go`
2. ✅ Duplicate `AtLeastOneOf` removed from `name` — now only declared on `id`

No further comments. **LGTM.**

jasonquigleyhpecom (approved):

PR #1119, opened 2026-05-13
Adding NSXt static configuration to load balancer resource

RDavisHPE:
Adding NSXt static configuration to load balancer resource.
- Added NSXt static configuration including attributes found in the hpegl equivalent including: Required:
- `tier1_gateways` (String) Provider ID of the Tier1 Gateway. Use hpegl_vmaas_router datasource to obtain the provider_id here. Optional:
- `admin_state` (Boolean) If `true` then admin State rule will be active/enabled.
- `log_level` (String) In Filter. Supported Values are "DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL", "ALERT", "EMERGENCY"
- `size` (String) In Filter. Supported Values are "SMALL", "MEDIUM", "LARGE"


eamonnotoole (commented):
## Review Summary

This PR adds typed `config_nsxt` support to the load balancer resource, mirroring the existing `config_haproxy` pattern. The schema design, validation, and create/read logic are solid. Some items to address:

---

### 🐛 Potential Issue: Import will not populate `config_nsxt`

The refactored `getLoadBalancerAsState` in `read.go` dispatches based on which config attribute exists in the current state/plan:

```go
switch {
case !plan.ConfigHaproxy.IsNull() && !plan.ConfigHaproxy.IsUnknown():
    // parse haproxy
case !plan.ConfigNsxt.IsNull() && !plan.ConfigNsxt.IsUnknown():
    // parse nsxt
default:
    // parse as generic dynamic config
}
```

During **import**, the plan/state will have all config attributes as null (only `id` is set by `ImportState`). This means imported NSX-T LBs will always fall into the `default` branch and populate `state.Config` (the generic dynamic attribute), not `state.ConfigNsxt`. The next `terraform plan` will then show a diff as Terraform tries to reconcile the state (`config` populated) with the user's HCL (`config_nsxt` set).

**Suggestion**: Consider detecting the LB type from the API response (like the old `isHAProxy` check) as a fallback when all typed config attributes are null — this would make import work correctly with the typed attributes.

---

### ⚠️ NSX-V test removal without explanation

Two NSX-V monitor tests are deleted (`TestAccMorpheusLoadBalancerMonitorNsxvExampleOk`, `TestAccMorpheusLoadBalancerMonitorNsxvUpdateOk`) along with `RenderLoadBalancerNsxvConfig`. If NSX-V is being deprecated/removed, that's fine — but the PR description only mentions adding NSX-T config. Consider:
- Adding a note in the PR body explaining why NSX-V tests are removed
- Verifying no other file references `RenderLoadBalancerNsxvConfig` (would cause compilation failure)

---

### 💡 Minor: `tier1_gateway` initialization in `parseNsxtConfig`

```go
var tier1 string  // defaults to ""
...
"tier1_gateway": types.StringValue(tier1),  // always non-null, even if API omits it
```

The other optional fields (`admin_state`, `log_level`, `size`) correctly initialize as `types.XxxNull()` and only become non-null when the API returns them. `tier1_gateway` is always set to a `StringValue` (even if empty string). Since it's Required this likely won't cause issues in practice, but for consistency consider:

```go
tier1Value := types.StringNull()
if tier1Raw, ok := configMap["tier1"]; ok {
    if v, ok := tier1Raw.(string); ok {
        tier1Value = types.StringValue(v)
    }
}
```

---

### 💡 Missing acceptance test for `config_nsxt` flow

The `RenderLoadBalancerNsxtConfig` helper is updated to use the new template (with `config_nsxt` block), and other resources' tests (e.g. virtual server) call it to create a prerequisite LB. However, there's no dedicated test that verifies the NSX-T LB `config_nsxt` attributes are read back correctly and survive import. The downstream tests exercise create but may not assert on `config_nsxt.tier1_gateway`, `config_nsxt.admin_state`, etc.

---

### 📝 Observations (non-blocking)

- **`RequiresReplace` on both individual attributes and the whole object**: This is redundant but harmless — the object-level modifier acts as a safety net.
- **`ExactlyOneOf` validator**: Good change from `AtLeastOneOf` — ensures mutual exclusivity properly.
- **`ValidateConfig` for `network_server_id`**: Nice addition — catches the missing required field at plan time with a clear error message rather than a cryptic API error.
- **Whitespace-only changes in `schema_gen.go`**: Many blank lines added after `err :=` in unrelated types (ConfigHaproxy, Permissions, Tenants). These appear to be auto-formatter artifacts — harmless but noisy.

---

### ✅ What looks good

- Clean schema with proper validators (`OneOf` for log_level/size)
- Create logic correctly maps plan attributes to SDK objects
- Read logic correctly parses API config map back into typed attributes
- `ValidateConfig` provides early feedback for NSX-T requirements
- Example and docs are clear and complete

eamonnotoole (commented):
## Review of recent commits (`0953a38` → `713d765`)

These three commits address the import issue I flagged in my earlier review. LGTM ✅

### What changed

The read config dispatch in `getLoadBalancerAsState` now uses the **API type code** (`isHAProxy`, `isNSXT`) rather than checking which plan config attribute is non-null:

```go
// Before (plan-based — broken on import):
case !plan.ConfigHaproxy.IsNull() && !plan.ConfigHaproxy.IsUnknown():
case !plan.ConfigNsxt.IsNull() && !plan.ConfigNsxt.IsUnknown():

// After (type-code-based — works on import):
case isHAProxy:
case isNSXT:
```

This correctly handles import: the API response always includes the type code, so the typed config attributes (`config_haproxy` / `config_nsxt`) will be populated regardless of whether the plan has them set.

### TypeCode handling

The `TypeCode` logic also improved:
- HAProxy without an explicit `type_code` in plan → null (preserving existing behavior)
- Explicit `type_code` in plan → preserved from plan
- Otherwise → set from API response (covers import)

This is correct and clean.

### No concerns

All three commits are focused refactors of the same function with no new issues introduced.

eamonnotoole (approved):

PR #1127, opened 2026-05-15
Add the hpe_morpheus_network_router_route resource

matthew-c-hpe:


matthew-c-hpe (commented):
## Code Review: PR #1127 — Add `hpe_morpheus_network_router_route` resource

Overall this is a well-structured resource following the project conventions. A few issues to address:

---

### 🐛 Bug: Copy-paste error in `create.go` (line ~65)

```go
if !plan.NextHop.IsNull() && !plan.Network.IsUnknown() {
    route.SetDestination(plan.NextHop.ValueString())
}
```

The second guard checks `plan.Network` but should check `plan.NextHop`:

```go
if !plan.NextHop.IsNull() && !plan.NextHop.IsUnknown() {
```

This means if `Network` is unknown but `NextHop` is set, the destination will still be set. Conversely, if `NextHop` is set but `Network` is unknown, the destination could be skipped incorrectly.

---

### ⚠️ `router_id` should be Required, not Optional+Computed

In `schema_gen.go`, `router_id` is declared as `Optional: true, Computed: true`. A route cannot exist without a parent router — this should be `Required: true`. As-is, a user could omit `router_id` and the create would call the API with a zero-value router ID, which would either fail confusingly or hit the wrong endpoint.

*(I understand this is a generated file — the fix belongs in the generator input/config.)*

---

### 📝 Misleading variable name in `read.go`

In the `Read()` method, the variable loaded from `req.State` is named `plan`:

```go
var plan NetworkRouterRouteModel
resp.Diagnostics.Append(req.State.Get(ctx, &plan)...)
```

This is reading from **state**, not from a plan. Renaming to `state` (or `current`) would improve clarity and be consistent with `delete.go` which correctly uses `var state`.

---

### 💡 Consider: `enabled` defaults to `false`

The schema sets `Default: booldefault.StaticBool(false)` for `enabled`. This means a route is **disabled by default**, which may surprise users. Most networking resources default to enabled. Worth confirming this is the intended UX — if not, flip the default to `true`.

---

### 🧹 Minor: Empty resource description in docs

The generated doc page (`morpheus_network_router_route.md`) has an empty description block. Adding a one-line description to the schema (e.g., *"Manages a static route on a Morpheus network router."*) would improve the docs and Terraform Registry listing.

---

### ✅ What looks good

- Clean separation of CRUD into individual files
- Proper use of `RequiresReplace()` plan modifiers since updates aren't supported
- Import logic with `router_id.id` composite format is well implemented
- Sweeper correctly scopes cleanup to test-prefixed resources
- Error handling with `errfmt.ErrMsg` and `TaintResourceState` follows project patterns
- Test covers create, plan stability, and import in one flow

matthew-c-hpe (commented):
## Re-review: PR #1127 (after `4d5fb23`)

Previous issues addressed:

- ✅ **Bug fixed**: `create.go:54` — `plan.NextHop.IsUnknown()` now correctly checks `NextHop` instead of `Network`
- ✅ **`router_id` is now Required** — schema updated correctly
- ✅ **`enabled` defaults to `true`** — sensible default for a networking resource

---

### One remaining nit: docs description

The "Used to create and delete Router routes." text was added as hardcoded text in the template (line 11), but the `description` frontmatter (line 4-5) and the `{{ .Description }}` placeholder (line 9) are still empty because the schema itself has no top-level `MarkdownDescription`. This means the Terraform Registry metadata will have a blank description.

If you want the description to flow through properly, add it to the schema in `schema_gen.go` (or its generator input):

```go
return schema.Schema{
    MarkdownDescription: "Manages a static route on a Morpheus network router.",
    Attributes: map[string]schema.Attribute{
```

That said, this matches the pattern of every other resource in the repo (none set a top-level schema description), so it's fine to leave as-is and address later across all resources.

---

### ✅ Everything else looks good

No other issues found. LGTM.

eamonnotoole (approved):

PR #1130, opened 2026-05-15
Load balancer monitor and virtual server fixes

eamonnotoole:
More validation for the resources and added examples for virtual server
resource.  Virtual server data-source added.


eamonnotoole (commented):
## Review Summary

Good PR overall — the new data source is well-structured, the validation improvements are valuable, and the improved examples/docs are a clear win. A few items to address:

---

### 🐛 Likely Bug: Data source tests assert wrong `vip_port`

In both `TestAccMorpheusLoadBalancerVirtualServerDataSourceByIdOk` and `TestAccMorpheusLoadBalancerVirtualServerDataSourceByNameOk` (`datasource_test.go`), the resource is created with `"VipPort": "443"`, but the assertion checks for `"80"`:

```go
"VipPort": "443",   // created with 443
...
resource.TestCheckResourceAttr(dsName, "vip_port", "80"),  // asserts 80?
```

If NSX-T normalizes the port on the server side, this deserves a comment explaining the behavior. Otherwise this looks like a test bug that would fail on execution.

---

### ⚠️ Test coverage regression: HAProxy/generic virtual server tests removed

`resource_test.go` is deleted entirely, which removes:
- `TestAccMorpheusLoadBalancerVirtualServerExampleOk` (HAProxy)
- `TestAccMorpheusLoadBalancerVirtualServerUpdateOk` (HAProxy update)

These tested non-NSX-T virtual server CRUD. Was the removal intentional (e.g., HAProxy is no longer supported/tested)? If so, fine — but if not, this is a coverage gap.

---

### 💡 Minor: Validator error message uses root path

In `requires_non_empty_string_also_requires_int64_at.go`, the error message uses `path.Root(v.SiblingName)`:

```go
fmt.Sprintf(
    "Attribute %q must be set when %q is %q.",
    path.Root(v.SiblingName),  // shows as top-level "persistence_profile"
    request.Path,
    request.ConfigValue.ValueString(),
)
```

Since this validator runs inside a nested object (`config_nsxt`), the error would display the sibling as a root-level attribute rather than `config_nsxt.persistence_profile`. Consider using `siblingPath` (already constructed above) for a more accurate message.

---

### 📝 Docs: No generic/non-NSX-T example remaining

The old `example.tf` (generic virtual server without `config_nsxt`) is deleted and replaced with NSX-T-only examples. If non-NSX-T load balancers are still supported, users of those types lose their reference example. Consider keeping a minimal generic example or noting "NSX-T is the only supported type" if that is the case.

---

### ✅ Things that look good

- Data source `Read` logic: clean separation of by-ID vs. by-name with proper multiple-match detection
- `populateVirtualServerState`: thorough mapping of all fields with proper null handling
- Pool ID preservation across Create/Update/Read for the NSX-T config quirk
- Read refactoring: the null-based config detection (replacing the `isImport` flag) is cleaner and handles more edge cases
- `RequiresNonEmptyStringAlsoRequiresInt64At` validator — nice reusable pattern
- Schema validators on `persistence_profile`, `ssl_client_profile`, `ssl_server_profile` enforce correct usage at plan time
- `receive_code` description improvement is accurate

eamonnotoole (commented):
## Review of latest commit: `09caf85` — "Review fixes"

LGTM ✅ — All three issues addressed cleanly:

1. **`vip_port` test assertions fixed** — corrected from `"80"` to `"443"` in both data source tests, with a clear comment explaining the intentional protocol/port mismatch is testing that the API does not normalize.

2. **Validator error message fixed** — now uses `siblingPath` (the actual resolved path) instead of `path.Root(v.SiblingName)`, so users will see the correct nested path like `config_nsxt.persistence_profile` rather than a misleading root reference. Unused `path` import also cleaned up.

3. **Resource test comments added** — the same explanatory comment was added to the NSX-T resource tests for consistency.

No issues with this commit.

jasonquigleyhpecom (approved):

PR #1145, opened 2026-05-25
Add tainting to a selection of resources

eamonnotoole:
These resources already do a GET after a POST.


eamonnotoole (commented):
## Code Review :white_check_mark:

The implementation of resource tainting here looks excellent and fully aligns with the planned pattern for our synchronous APIs that require a GET-after-POST to read the complete state.

### Key Highlights:
- **Tainting Pattern:** The introduction of `cleanup.TaintResourceState` cleanly handles scenarios where the resource is successfully created in Morpheus, but a network blip or timing issue causes the subsequent `Read` to fail. Setting the ID only and tainting the state ensures Terraform can appropriately destroy and recreate (or allow for a manual import), preventing orphaned resources.
- **Improved Error Messaging:** For resources relying on a post-POST List where an ID isn't directly extractable upon failure (`library_instance_type`, `library_option_type_list`), the updated diagnostics correctly guide the user towards manual verification and import.
- **VDI Pool Handling:** The two-stage read in `vdi_pool` (List then Get) is handled precisely. The List failure provides improved messaging, while the GET failure leverages the acquired ID to taint the state.
- **Consistency:** The `errfmt.DiagError` references in the `Read` failure blocks correctly utilize `plan.Name.ValueString()` now instead of empty strings, enhancing debuggability.

Builds and lints correctly pass.

jasonquigleyhpecom (approved):

PR #1150, opened 2026-05-27
Updates to security_group and subnet

eamonnotoole:
In this PR we:
- use SDK v0.61.0
- update security_group and subnet to support more API fields (for example permissions and tenants)
- add docs templates
- add examples


eamonnotoole (commented):
# 🔍 terraform-provider-hpe PR Review

I have reviewed the recent updates to the **Security Group** and **Subnet** resources in PR #1150. Overall, the updates align well with provider standards and the migration to `v0.61.0` of the SDK.

Here is my assessment of the recent changes:

### ✅ Gaps & Completeness
- `type_id` mapping has been successfully restored in `mapCreateResponseToModel` and `mapResponseToModel` for Subnets. This resolves the potential state drift.
- Full CRUD operations correctly map the new `visibility`, `active`, and permissions fields using the `convert` package helpers.

### ⚠️ Warnings
- **Missing Spec Field:** As noted in the code with a `TODO`, `network_server_id` is supported by the API as a create-only field for HVM/Standard clouds with multiple network integrations. Since it is not returned in the GET response, it should eventually be implemented as a `WriteOnly: true` schema attribute. It is safe to defer this as it is not strictly required for Azure, GCP, or NSX-T clouds.
- **Tenant Removal:** In `Update()` for both resources, the code currently skips assigning tenants to the request body if `plan.TenantIds.IsNull()` is true. 
  ```go
  if !plan.TenantIds.IsNull() && !plan.TenantIds.IsUnknown() {
      // ... sets body.Tenants
  }
  ```
  If a user removes all tenants from their Terraform configuration, `plan.TenantIds` may evaluate to Null, meaning the API will not receive an empty list to clear the existing tenants. **Verify if the Morpheus API requires an explicit empty array to remove all tenants.**

### 🏆 Pattern Compliance
- **Perfect Delete Idempotency:** The 404 check (`errfmt.IsNotFound(httpResp)`) before `errfmt.CheckResponse` is correctly implemented for idempotent destroys.
- **SDK Constructors:** Excellent use of `sdk.NewCreateSubnetRequestSubnetWithDefaults()` and `sdk.NewUpdateSubnetRequestSubnetWithDefaults()`.
- **Safe Type Conversions:** Nullable SDK response fields are correctly safeguarded using `convert.Int64ToType()`, `convert.StrToType()`, etc.

### 📝 Summary
**Pass.** The structure, error formatting, and code quality are solid. No blocking violations were found.

eamonnotoole (commented):
# 🔍 terraform-provider-hpe PR Review Update

Following up on the previous review, I have verified the recent commits addressing the **Tenant Removal** perpetual diff issue in PR #1150.

### ✅ Tenant Associations Fix
- The perpetual diff bug for `tenant_ids` removal has been correctly addressed in both **Security Group** and **Subnet** `Update` methods.
- By providing an `else` clause when the plan value is `null`, the provider now explicitly sends an empty array (`[]int64{}` and `[]sdk.UpdateSubnetRequestSubnetTenantsInner{}`) to the Morpheus API.
- This correctly informs the Morpheus `permissionService.updateTenantPermissions` logic to delete any existing tenant associations, behaving exactly as expected and matching the proven pattern used for `labels` in the `cloud` and `group` resources.

### 📝 Summary
**Pass.** The warning from the initial review has been successfully resolved. Code quality remains high and the correct handling of `null` HCL optional attributes prevents state drifting issues. Excellent work!

matthew-c-hpe (approved):

PR #1180, opened 2026-06-10
feat: add server, network_pool, network_type, security_group datasources and migrate to shrunk SDK

eoghanino:


docs/data-sources/morpheus_network_pool.md
```diff
@@ -0,0 +1,32 @@
+---
+page_title: "hpe_morpheus_network_pool Data Source - terraform-provider-hpe"
+subcategory: "morpheus"
+description: |-
+  
+---
+# hpe_morpheus_network_pool (Data Source)
```
matthew-c-hpe, line 7: We should have rendered example code here. If you insert a `.md.tmpl` in `docs/datasources` we can get `tfplugin-docs` to render this markdown file from the template, rather than hard coding this.

matthew-c-hpe: Inside the `.md.tmpl` we can include references to files. We typically have references to rendered `.tf` example files which live in the `examples` directory.

We have a pattern of including `.tf.tmpl` files in the relevant data source package, which we generate `example.tf` files in that `examples` directory using `go:generate` statements in `example.go` in the data source package.

morpheus/framework/datasources/networkpool/datasource_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/HPE/terraform-provider-hpe/provider"
+)
+
+func TestMain(m *testing.M) {
+	code := m.Run()
+	testhelpers.WriteMergedResults()
+	os.Exit(code)
+}
+
+func newProviderWithError() (tfprotov6.ProviderServer, error) {
```
matthew-c-hpe: We've got a helper function in the `testhelpers` package which we can use as a provider constructor, so we don't need this any more for constructing the provdier. I'll leave a comment below showing where to put it.

morpheus/framework/datasources/networkpool/datasource_test.go
```diff
@@ -0,0 +1,70 @@
...
+	providerConfig := testhelpers.ProviderBlock()
+
+	datasourceConfig := `
+data "hpe_morpheus_network_pool" "test" {
+  name = "Morpheus"
+}
+`
+
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
```
matthew-c-hpe: We can replace this with:

```
ProtoV6ProviderFactories: testhelpers.GetAccTestFactories(t, morpheus.New(), nil),
```
That third argument in the call to GetAccTestFactories is used if we need to construct an sdkv2 provider for the acceptance test - not needed for these data source tests as they purely use the Terraform Plugin Framework provider. But it's useful for certain resource acceptance tests if we need to mix `framework` and `sdkv2`, e.g. if we have `sdkv2` resources as dependencies in testing.

PR #1183, opened 2026-06-11
Add subnet_id to hpe_morpheus_instance network_interfaces (MORPH-12632)

eamonnotoole:
## What

Adds a `subnet_id` input to `hpe_morpheus_instance` `network_interfaces` **and** its nested `child_virtual_networks`, enabling provisioning onto a specific subnet (required by some clouds, notably **Azure**). Resolves **MORPH-12632**.

## How it works

The Morpheus API encodes the network selection as a single `network.id` string dispatched by prefix (verified in morpheus-ui `InstanceService.parseNetworks`, which parses child interfaces through the same recursive path):

| Input | `network.id` sent |
|---|---|
| `network_id` | `"{id}"` |
| `network_group_id` | `"networkGroup-{id}"` |
| `subnet_id` (new) | `"subnet-{id}"` |

`subnet_id` is `Optional`, `ConflictsWith` `network_id` and `network_group_id`, and is available on **both** `network_interfaces` and `child_virtual_networks`.

## Read / import behaviour

The API resolves a subnet to its parent network (returned as `network_id`) **and also returns the subnet** on `containerDetails.server.interfaces[].subnet`. That field was missing from the OpenAPI spec/SDK; it is added in morpheus-openapi and shipped in **oapigen v0.64.0**, which this PR bumps to.

- **Read:** `subnet_id` is read back directly from the server interface's `subnet`, so it round-trips on refresh (no plan-preservation workaround). `network_id` is `Optional+Computed`, so the resolved parent network is stored without a diff.
- **Import:** `subnet_id` is null — the `instance.interfaces` list used on import does not carry the subnet (only the server-interface read path does); the resolved `network_id` is populated instead.

## Changes

- `go.mod` — bump `oapigen` to **v0.64.0**
- `schema_gen.go` — `subnet_id` on `network_interfaces` and `child_virtual_networks`
- `mappings.go` — `"subnet-{id}"` branch in the top-level and child create/update mappers
- `instance_read.go` — read `subnet_id` back from `iface.Subnet` (server + child paths); explicit null on import
- `mappings_internal_test.go`, `instance_read_internal_test.go` — unit tests for the mappers (all three prefixes, both levels) and the read-back
- Azure example + doc section; regenerated docs

## Testing

- `go build ./...`, `go vet`, `gofumpt`, `golangci-lint` (0 issues) — pass
- Unit tests pass (mapper prefixes + subnet read-back)
- `make docs` regenerated and idempotent

## Dependencies

- morpheus-openapi #266 (server-interface `subnet` read field) → SDK **oapigen v0.64.0**


morpheus/framework/resources/instance/schema_gen.go
```diff
@@ -546,6 +546,18 @@ func InstanceResourceSchema(ctx context.Context) schema.Schema {
 							Description:         "Is this interface the 'primary interface'?",
 							MarkdownDescription: "Is this interface the 'primary interface'?",
 						},
+						"subnet_id": schema.Int64Attribute{
+							Optional:            true,
+							Computed:            true,
+							Description:         "id of the subnet to be used (required by some clouds, e.g. Azure). Cannot be used with 'network_id' or 'network_group_id'.",
+							MarkdownDescription: "id of the subnet to be used (required by some clouds, e.g. Azure). Cannot be used with 'network_id' or 'network_group_id'.",
+							Validators: []validator.Int64{
```
matthew-c-hpe, line 580: How about add a RequiredWith validator to this to require it when using Azure?

jasonquigleyhpecom (approved):

PR #1186, opened 2026-06-11
Add field_context to hpe_morpheus_form option types (MORPH-12635)

eamonnotoole:
## What

Adds an optional `field_context` attribute to `hpe_morpheus_form` option types. By default a form field's value is stored under `config.customOptions`; setting `field_context = "instance"` lets a custom field target `instance.*` properties instead. Resolves **MORPH-12635**.

## How it works

- **Schema:** `field_context` is Optional + Computed (free-form string). The value set isn't a fixed enum in Morpheus (`config`, `instance`, `config.n`, …), so no `StringInSlice` validator is applied; the description documents the common values. Default is `config`.
- **Request:** `applyOptionTypeConfigByType` sends `fieldContext` **only when set**, so the Morpheus default (`config`) is preserved when the attribute is omitted. This single insertion point covers all call sites (top-level + field-group, create + update).
- **Read-back:** `buildOptionTypeRow` populates `field_context` from the option type's `fieldContext`.

Omitting `field_context` preserves the previous behavior exactly (backward-compatible).

## Stacking

> [!NOTE]
> This PR is **stacked on #1182** (base branch `fix-form-cascade-bugs`). The `field_context` read-back lands in the `buildOptionTypeRow` helper introduced by that PR, which avoids conflicts in the form read path. Please merge #1182 first; GitHub will retarget this PR's base to `dev-1.5` automatically once #1182 merges.

## Changes

- `form.go` — `field_context` schema attribute, request-side handling, and read-back
- `field_context_internal_test.go` — unit tests
- regenerated docs

## Testing

- `go build ./...`, `go vet`, `gofumpt`, `golangci-lint` (0 issues) — pass
- Unit tests pass (`TestBuildOptionTypeRowFieldContext`; `TestApplyOptionTypeConfigFieldContext` — sent when set, omitted when empty)
- `make docs` regenerated and idempotent
- Note: an end-to-end acceptance test would require a live environment; the request/read logic is covered by the unit tests.

### DIFF
--- a/docs/resources/morpheus_form.md
+++ b/docs/resources/morpheus_form.md
@@ -454,6 +454,11 @@ resource "hpe_morpheus_form" "example" {
 
 ### Cloud (cloud)
 
+~> **Note:** For the `cloud` option type, set `group_field` (with `group_field_type = "field"`)
+to the code of the upstream group form field that drives the group→cloud cascade. Morpheus's cloud
+input uses the plain `group` config key for this dependency; the provider maps `group_field` to it
+automatically.
+
 ```terraform
 resource "hpe_morpheus_form" "example" {
   name        = "demo"
@@ -483,6 +488,35 @@ resource "hpe_morpheus_form" "example" {
     instance_type_field_type = "value"
     instance_type_code       = "apache"
     cloud_type               = "4"
+  }
+}
+```
+
+#### Cloud with Group Cascade
+
+```terraform
+resource "hpe_morpheus_form" "example" {
+  name        = "demo-cascade"
+  code        = "demo-cascade"
+  description = "Demonstrates group→cloud cascade"
+  labels      = ["terraform", "demo"]
+
+  option_type {
+    name        = "Group Selector"
+    code        = "group-selector"
+    type        = "group"
+    field_label = "Group"
+    field_name  = "fGroups"
+  }
+
+  option_type {
+    name             = "Cloud Selector"
+    code             = "cloud-selector"
+    type             = "cloud"
+    field_label      = "Cloud"
+    field_name       = "fClouds"
+    group_field_type = "field"
+    group_field      = "fGroups"
   }
 }
 ```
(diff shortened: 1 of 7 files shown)


docs/resources/morpheus_form.md
```diff
@@ -1201,6 +1201,7 @@ Optional:
 - `enable_storage_type_selection` (Boolean) Whether to allow users to select a storage type for a diskManager option type
 - `exclude_from_search` (Boolean) Whether the option type should be excluded from search or not
 - `export_meta` (Boolean) Whether to export the option type as a tag
+- `field_context` (String) The context the field value is stored in. Defaults to 'config' (stored in config.customOptions). Set to 'instance' to target instance.* properties.
```
RDavisHPE, line 1204: To be more accurate - since we're not setting a default on our end - we could say "If omitted, Morpheus applies its default"

(I also found that my examples tend to populate back with `config.customOptions` and not `config` so perhaps that phrasing would reduce potential confusion.

morpheus/sdkv2/resources/form/form.go
```diff
@@ -364,6 +364,12 @@ func validateFormOptionTypes(_ context.Context, d *schema.ResourceDiff, _ any) e
 }
 
 func applyOptionTypeConfigByType(row map[string]any, optionTypeConfig map[string]any) diag.Diagnostics {
+	// field_context applies to any option type. Only send it when set so the
+	// Morpheus default ("config") is preserved when it is omitted.
+	if fieldContext, ok := optionTypeConfig["field_context"].(string); ok && fieldContext != "" {
```
RDavisHPE, line 369: One thing I found here - _if_ the user explicitly sets `field_context = "" `  terraform will ignore any changes to `field_context` after that fact (and if the user sets it to "" after setting it to a value, it will ignore that too). 

The omit check is still good to have as fieldContext is generally populated to the default on the read - so maybe a DiffSuppress could work here. Still - that could complicate things so I'm not against the current solution.

PR #1191, opened 2026-06-15
Add instance power state and snapshot resources

eamonnotoole:
## Summary

Adds two resources for managing existing Morpheus instances (hpegl VMaaS parity):

- **`hpe_morpheus_instance_power_state`** — declarative power management of an existing instance (`running` / `stopped` / `suspended`). Idempotent; detects drift and converges (re-applies the power transition if the instance is changed out of band); `delete` is a no-op (removing the resource does not change power).
- **`hpe_morpheus_instance_snapshot`** — lifecycle-managed instance snapshots. `name`/`description`/`memory_snapshot` are create-time (ForceNew); `retain_on_delete` keeps the snapshot in Morpheus when the resource is destroyed (removed from state only). Create is asynchronous (no id returned), so it polls the instance's snapshot list by name to resolve the new snapshot. Import uses the composite id `<instance_id>/<snapshot_id>`.

Both are standard declarative resources (OpenTofu-compatible) rather than framework actions.

## SDK

- Bumps `hpe-morpheus-go-sdk` oapigen to **v0.67.0** (snapshot ids are now `int64`, so the prior `int32` casts are removed).
- Updates `network_router_nat` for the v0.67.0 NAT update model (the `Name` field is now a pointer).

## Testing

- `go build ./...` and `go build -tags sweep ./...`: pass
- `golangci-lint` on the new packages: 0 issues
- `make docs`: clean tree
- Acceptance tests are authored but not run here (require a live appliance and an existing instance).

Resolves MORPH-7625, MORPH-7626.


morpheus/framework/datasources/instancesnapshot/datasource_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/HPE/terraform-provider-hpe/provider"
+)
+
+func TestMain(m *testing.M) {
+	code := testhelpers.TestMain(m)
+	testhelpers.WriteMergedResults()
+	os.Exit(code)
+}
+
+func newProviderWithError() (tfprotov6.ProviderServer, error) {
```
matthew-c-hpe: Prefer `testhelpers.GetAccTestFactories` as the constructor

morpheus/framework/resources/instancesnapshot/resource_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/HPE/terraform-provider-hpe/provider"
+)
+
+func TestMain(m *testing.M) {
+	code := testhelpers.TestMain(m)
+	testhelpers.WriteMergedResults()
+	os.Exit(code)
+}
+
+func newProviderWithError() (tfprotov6.ProviderServer, error) {
```
matthew-c-hpe: Prefer `testhelpers.GetAccTestFactories` as the constructor

jasonquigleyhpecom (dismissed):

matthew-c-hpe (approved):

PR #1192, opened 2026-06-15
Fix AMV networking acceptance tests + bump oapigen SDK to v0.67.0

jasonquigleyhpecom:
## Summary

Fixes failing AMV networking acceptance tests across the NSX‑T / network‑router family and the subnet resource, and bumps the generated SDK to pick up the supporting OpenAPI changes. Took the suite from ~75 failures down to the paused/deferred/out‑of‑scope baseline.

## SDK bump
- `go.mod`: `oapigen v0.63.0 → v0.67.0` (NAT `config.action` request models + network‑router `networkServer` shape, regenerated from merged `morpheus-openapi@tf-dev`). Verified `go build`/`go vet`/`golangci-lint` clean in CI mode (`GOWORK=off`).

## Resource / data source fixes
- **Subnet**: read write‑only `config` from `req.Config` (Azure `subnetName`/`subnetCidr`); `name` optional+computed; map empty `labels` to null (fixes inconsistent‑result‑after‑apply); tests run serially to avoid Azure per‑VNet collision; use Azure network + deletable subnet type.
- **Network Router NAT**: send `action` via `config.action` (create required / update optional); per‑test tier‑1 connected to tier‑0.
- **Network Router**: send `enable_bgp` on create; data source interfaces‑set fix (`SetValueFrom` over typed values).
- **BGP neighbor**: EBGP with valid tier‑0 source address.
- **DHCP server**: `config_nsxt` round‑trip + import ignore.
- **Firewall rule / rule‑group**: self‑contained fixtures + DS fixes.
- **Task (vRO)**: panic fix + body assertion/DiffSuppress.

## Notes
- Branch is 15 commits ahead of `dev-1.5` (and ~6 behind — may need a rebase before merge).
- Remaining acceptance failures are paused (BGP/LB/NAT — opaque NSX‑T realization errors), deferred (router firewall feature), or out‑of‑scope/env.


morpheus/framework/resources/network_router_nat/resource.go
```diff
@@ -169,6 +194,12 @@ func getNatAsState(
 		state.Name = types.StringValue(*nat.Name)
 	}
 
+	if nat.Action != nil {
+		state.Action = types.StringValue(*nat.Action)
+	} else {
+		state.Action = plan.Action
```
matthew-c-hpe, line 200: We want to store action to state as null if it's not there, no?

The provider won't correctly report a diff if Action's nil here.

matthew-c-hpe (approved):

PR #1201, opened 2026-06-17
Reject self-referential dependent_field on option types and forms

eamonnotoole:
## Summary

A form field / option type configured with `dependent_field` equal to its own `field_name` makes the field depend on itself, producing a circular `dependsOnCode` and an unstable form reload. Morpheus stores a submitted `dependsOnCode` verbatim and provides no guard against self-reference, so this adds a plan-time validator.

## Changes

- **`helpers.ValidateDependentFieldNotSelf`** — a `CustomizeDiff` that rejects `dependent_field == field_name`. It reads the raw config, so it never fires on a computed read-back of `dependent_field`. Wired into the nine standalone option type resources: `checkbox`, `hidden`, `number`, `password`, `radio_list`, `select_list`, `text`, `textarea`, `typeahead`.
- **`hpe_morpheus_form`** — the same check added to `validateOptionTypeConfig`, so inline form fields (root and field-group) are covered.
- Plan-time `ExpectError` acceptance tests for both the standalone-resource and form paths.

## Behaviour

- **Rejected** at plan time: `dependent_field` equal to the field's own `field_name`.
- **Still allowed**: `dependent_field` pointing at a *different* field (the normal cascade/reload case).

## Testing

- `go build ./...`, `go vet`, `golangci-lint run` — all clean.
- New `ExpectError` tests assert the plan-time rejection; no schema change, so docs are unchanged.

Refs: MORPH-12958


morpheus/sdkv2/resources/form/dependent_field_test.go
```diff
@@ -0,0 +1,55 @@
...
+// guard that rejects an option type whose dependent_field equals its own
+// field_name (a field cannot depend on itself, which would create a circular
+// dependsOnCode). This is a pure plan-time validation, so it needs no live
+// appliance.
+func TestAccMorpheusFormRejectsSelfReferentialDependentField(t *testing.T) {
+	t.Parallel()
+
+	if testing.Short() {
+		t.Skip("Skipping slow test in short mode")
+	}
```
RDavisHPE: Since this is a pure plan time test - it shouldn't need to skip in short mode and could be ran as a unit test

morpheus/sdkv2/resources/optiontype/validators.go
RDavisHPE, line 1: This function is only used in the option_type implementation - it would probably be better off in the optiontype package as a private function (like the form implementation)/

RDavisHPE (dismissed):

morpheus/sdkv2/resources/optiontype/option_type_text_resource_test.go
```diff
@@ -175,3 +176,31 @@ func TestAccMorpheusOptionTypeTextExampleOk(t *testing.T) {
...
+	invalidConfig := `
+resource "hpe_morpheus_option_type_text" "bad_self_ref" {
+  name            = "bad-self-ref-test"
+  field_name      = "myField"
+  field_label     = "My Field"
+  dependent_field = "myField"
+}
+`
+	resource.Test(t, resource.TestCase{
+		ProtoV6ProviderFactories: testhelpers.GetAccTestFactories(t, morpheus.New(), sdkv2morpheus.Provider()),
```
RDavisHPE, line 199: We typically have `IsUnitTest: true` for any unit tests (same goes for the form test)

RDavisHPE (approved):

PR #1213, opened 2026-06-23
Add EfficientIP (SOLIDserver) support to network pool server

eamonnotoole:
## Summary

Adds support for **EfficientIP (SOLIDserver)** to `hpe_morpheus_network_pool_server`, and makes the pool server type selectable by a stable code (in addition to the existing numeric ID).

## Changes

- **New `type_code` attribute** (string) — selects the pool server type by its stable code (e.g. `infoblox`, `bluecat`, `phpipam`, `solarwindsipam`, `solidserver`). Mutually exclusive with the existing numeric `type_id`; exactly one is required. The resolved code is now sent to the API, which also corrects type selection for non-Infoblox types.
- **`type_id` relaxed** from required to optional/computed (non-breaking). When set, its type code is looked up and used.
- **New `inventory_existing` attribute** (bool) — controls whether existing networks are inventoried from the pool server. Applies to all pool server types.
- **Examples** switched to `type_code`; added an **EfficientIP** example.
- **Docs**: added an EfficientIP section, an applicability-table column, and a reference list of known pool server type codes.

## Known type codes

| IPAM System | `type_code` |
|---|---|
| Infoblox | `infoblox` |
| BlueCat | `bluecat` |
| phpIPAM | `phpipam` |
| SolarWinds | `solarwindsipam` |
| EfficientIP (SOLIDserver) | `solidserver` |

EfficientIP requires its integration/plugin to be installed. The authoritative, per-environment list is available from `GET /api/networks/pool-server-types`.

## Testing

- `go build`, `go vet`, `golangci-lint` (no new issues), `go test -short`, and `make docs` all pass.
- Acceptance test updated to exercise the `type_code` path (Infoblox). Live acceptance tests require a configured Morpheus environment.

## Notes

- `inventory_existing` is written as `on`/`off` and read tolerantly; the exact persisted representation should be confirmed against a live environment.


morpheus/framework/resources/network_pool_server/resource.go
```diff
@@ -413,4 +445,88 @@ func mapReadResponseToModel(
...
+	if b {
+		v = "on"
+	}
+
+	return &v
+}
+
+// networkPoolServerConfigBool coerces a generic config map value (string or bool) into a
+// bool, accepting the common truthy representations Morpheus may persist for a checkbox.
+func networkPoolServerConfigBool(v interface{}) bool {
```
matthew-c-hpe: Use convert.StringToBool, it looks like inventoryExisting is an on/off checkbox, the `convert` function handles this so we can remove these two helper functions.

morpheus/framework/resources/network_pool_server/resource.go
```diff
@@ -387,12 +388,18 @@ func mapReadResponseToModel(
 	if server.Name != nil {
 		model.Name = types.StringValue(*server.Name)
 	}
+	// type_id and type_code are mutually exclusive Optional+Computed attributes, so
```
matthew-c-hpe: We shouldn't compute what wasn't set to null, we should compute it to its actual value - it's there in `t / server.Type` - at least I think that computing to actual value is more reasonable than null.

eamonnotoole: Done in 754281b7 — both `type_id` and `type_code` now reflect their actual values from `server.Type` (nil-safe via `convert.Int64ToType` / `convert.StrToType`), no nulling.

I also re-checked the earlier "breaks ConflictsWith" concern against the framework source: `schemavalidator.ConflictsWithValidator.Validate` reads only `req.Config`/`req.ConfigValue` — it returns early when the attribute is null in *config* and errors only if a conflicting attribute is non-null in *config*. It never inspects state or plan, so reflecting both actual values in state is safe. Re-added the `type_id` set assertion to the acceptance test too.

matthew-c-hpe (approved):

PR #1253, opened 2026-07-03
Add six lookup data sources to resolve resource IDs by name

jasonquigleyhpecom:
## Summary

Adds six read-only data sources so configurations can look up Morpheus entities
by `name` (or `id`) instead of hard-coding numeric IDs when wiring up resources.
Each fills a gap where a resource attribute references an entity that previously
had no corresponding data source in the provider.

| Data source | Resolves | Used by (resource attribute) |
|---|---|---|
| `hpe_morpheus_monitoring_check_type` | monitoring check type | `hpe_morpheus_monitoring_check.check_type_id` |
| `hpe_morpheus_backup_job` | backup job | `hpe_morpheus_backup_host.job_id`, `hpe_morpheus_backup_instance.job_id` |
| `hpe_morpheus_cluster_layout` | cluster layout | `hpe_morpheus_cluster.layout_id` |
| `hpe_morpheus_network_router_type` | network router type | `hpe_morpheus_network_router.type_id` |
| `hpe_morpheus_network_pool_server_type` | network pool server type | `hpe_morpheus_network_pool_server.type_id` |
| `hpe_morpheus_subnet_type` | subnet type | `hpe_morpheus_subnet.type_id` |

Each data source supports lookup by `id` or by `name` (mutually exclusive),
following the existing `hpe_morpheus_network_type` pattern.

## Implementation notes

- Generated schemas, hand-written id/name read logic, examples, docs, and
  acceptance tests for each.
- `network_router_type` matches by name client-side (its list endpoint has no
  name filter); `cluster_layout` uses the phrase filter plus a client-side exact
  match.
- `subnet_type` adds a `GetSubnetType` SDK method and its response models,
  pending the corresponding OpenAPI change (`GET /api/subnet-types/{id}`). All
  other data sources use existing SDK methods.

## Verification

- `go build ./...` — passes
- `make lint` — 0 issues
- `make docs` — regenerated and idempotent (clean tree)
- data source test packages compile; `go vet` clean
- Live acceptance tests require `TF_ACC` and a Morpheus endpoint; not run in this environment

Opening as a **draft** for early review. The `subnet_type` OpenAPI change is
tracked separately; once it lands and the SDK is regenerated, the hand-added
`GetSubnetType` code can be dropped in favor of the generated version.


morpheus/framework/datasources/backupjob/datasource.go
```diff
@@ -0,0 +1,231 @@
...
+		return getBackupJobByID(ctx, config.Id.ValueInt64(), apiClient)
+	} else if !config.Name.IsNull() {
+		return getBackupJobByName(ctx, config.Name.ValueString(), apiClient)
+	}
+
+	return nil, errors.New(ErrorNoValidSearchTerms)
+}
+
+// Read refreshes the Terraform state with the latest data.
+func (d *DataSource) Read(
```
eamonnotoole, line 184: 🔴 **Gap — split `Read()` into `read.go`.** All six new packages bundle `Read()` and the state-mapping helpers (`...AsState`, `get...ByID`, `get...ByName`) inside `datasource.go`. Per provider conventions, `datasource.go` should hold only `NewDataSource()`, `Metadata()` and `Schema()`, with the read logic in a dedicated `read.go`. This comment is representative — the same applies to `clusterlayout`, `monitoringchecktype`, `networkpoolservertype`, `networkroutertype` and `subnettype`.

morpheus/framework/datasources/clusterlayout/datasource.go
```diff
@@ -0,0 +1,236 @@
...
+		HasSettings:             convert.BoolToType(l.HasSettings),
+		InstallContainerRuntime: convert.BoolToType(l.InstallContainerRuntime),
+		MemoryRequirement:       convert.Int64ToType(l.MemoryRequirement),
+		ServerCount:             convert.Int64ToType(l.ServerCount),
+		SortOrder:               convert.Int64ToType(l.SortOrder),
+	}
+
+	// DateCreated is *time.Time
+	if l.DateCreated != nil {
+		state.DateCreated = types.StringValue(l.DateCreated.String())
```
eamonnotoole: 🟠 **Violation — non-RFC3339 timestamp.** `l.DateCreated.String()` (and `l.LastUpdated.String()` on line 97) emit Go's internal time format (`2006-01-02 15:04:05.999999999 -0700 MST`). Terraform state conventionally uses RFC3339. Use `.Format(time.RFC3339)`, or reuse the `timeToType` helper as the `backup_job` data source does.

internal/sdk/oapigen/api_networks.go
```diff
@@ -17414,3 +17414,127 @@ func (a *NetworksAPIService) UpdateSubnetExecute(r ApiUpdateSubnetRequest) (*Upd
 
 	return localVarReturnValue, localVarHTTPResponse, nil
 }
+
+type ApiGetSubnetTypeRequest struct {
```
eamonnotoole: 🟠 **Violation — hand-edit in generated SDK.** This file (and the new `model_get_subnet_type_*.go` files) carry a `Code generated by OpenAPI Generator ... DO NOT EDIT` header. Manually adding `GetSubnetType` / `GetSubnetTypeExecute` here means it will be overwritten the next time the vendored SDK is regenerated. Hand-written SDK additions belong in `internal/sdk/legacy/` (which is preserved across regeneration), not in the generated `oapigen` package.

morpheus/framework/datasources/backupjob/datasource_test.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+
+	"github.com/HPE/terraform-provider-hpe/morpheus/framework/datasources/backupjob"
+	"github.com/HPE/terraform-provider-hpe/morpheus/testhelpers"
+	"github.com/HPE/terraform-provider-hpe/morpheus/testhelpers/capabilities"
+	"github.com/HPE/terraform-provider-hpe/provider/adapter"
+)
+
+func TestMain(m *testing.M) {
+	code := m.Run()
```
eamonnotoole: 🟠 **Violation — `TestMain` bypasses sweep interception.** `code := m.Run()` skips `testhelpers.TestMain(m)`, which intercepts the `-sweep` flag. Use `code := testhelpers.TestMain(m)` as the `framework/resources/*` tests do. Same on line 19 of the other five new `datasource_test.go` files (`clusterlayout`, `monitoringchecktype`, `networkpoolservertype`, `networkroutertype`, `subnettype`).

morpheus/framework/datasources/clusterlayout/datasource.go
```diff
@@ -0,0 +1,236 @@
...
+	_ datasource.SchemaRequest,
+	resp *datasource.SchemaResponse,
+) {
+	resp.Schema = ClusterLayoutDataSourceSchema(ctx)
+}
+
+func clusterLayoutAsState(
+	_ context.Context,
+	l *sdk.GetClusterLayout200ResponseLayout,
+) (ClusterLayoutModel, diag.Diagnostics) {
```
eamonnotoole: 🔵 **Recommendation — drop the unused `diag.Diagnostics`.** `clusterLayoutAsState` declares, threads and returns a `diag.Diagnostics` that is never populated. The signature can be simplified to return just `ClusterLayoutModel`, dropping the always-empty append at the call site.

eamonnotoole (dismissed):
## Automated code review — six lookup data sources

Reviewed the PR diff against `dev-1.6` (68 files). Overall the data sources are well structured: the config overrides correctly unnest the response envelopes and drop nested blocks, and client-side exact-match filtering compensates for endpoints that lack a `name` query parameter. A few items should be addressed before merge.

| Severity | Count |
|---|---|
| 🔴 Gap | 1 |
| 🟠 Violation | 3 |
| 🔵 Recommendation | 1 |

**Blocking**
- 🔴 `Read()` and its state helpers live in `datasource.go` in all six packages; they should be split into a dedicated `read.go`.
- 🟠 `clusterlayout` timestamps use `.String()` (Go's internal format) instead of RFC3339.
- 🟠 Hand-edits to the generated `internal/sdk/oapigen` package will be lost on the next SDK regeneration.
- 🟠 The new `datasource_test.go` files call `m.Run()` directly, bypassing `testhelpers.TestMain(m)`.

**Non-blocking**
- 🔵 `clusterLayoutAsState` returns an always-empty `diag.Diagnostics`.

**Pre-existing:** the timestamp `.String()` and `m.Run()` patterns also appear in many older data sources (e.g. `loadbalancerpool`, `role`, `cloud`) — out of scope for this PR, worth a follow-up.

morpheus/framework/datasources/clusterlayout/datasource.go
jasonquigleyhpecom: Fixed — `cluster_layout` now formats `date_created`/`last_updated` as RFC3339 (via the shared `convert.TimeToType` helper) instead of `.String()`.

internal/sdk/oapigen/api_networks.go
jasonquigleyhpecom: Reverted — `subnet_type` now uses the generated `ListSubnetTypes` (with client-side id/name matching), and the hand-edits to the generated `oapigen` package (`api_networks.go` plus the `model_get_subnet_type_*` files) have been removed, so nothing is lost on the next SDK regeneration.

morpheus/framework/datasources/backupjob/datasource_test.go
jasonquigleyhpecom: Fixed — all six `datasource_test.go` files now use `code := testhelpers.TestMain(m)` so the `-sweep` flag is honoured.

morpheus/framework/datasources/clusterlayout/datasource.go
jasonquigleyhpecom: Done — `clusterLayoutAsState` now returns just `ClusterLayoutModel`; the always-empty `diag.Diagnostics` and its append at the call site are gone.

matthew-c-hpe (approved):

PR #1256, opened 2026-07-03
MORPH-13145/13146/13147/13209: Fix network router NAT, firewall, BGP

eamonnotoole:
## Summary

Fixes four issues on the NSX-T network router sub-resources and adds a supporting data source.

## Fixes

**NAT** (hpe_morpheus_network_router_nat) — MORPH-13147, MORPH-13145
- Send the required firewall match and service fields on create/update (previously rejected with a 400 for NSX-T routers).
- Default priority.
- Guard nil optional fields on read (previously could panic).

**Firewall rule** (hpe_morpheus_network_router_firewall_rule) — MORPH-13146
- Send the parent rule-group id (parent_id) plus policy, direction, protocol and port_range (previously rejected with a 400 / server-side nil pointer).

**BGP neighbor** (hpe_morpheus_network_router_bgp_neighbor) — MORPH-13209
- Default the graceful-restart mode and route filtering, and correct the graceful-restart enum value (previously rejected with a 400).

## New data source

**hpe_morpheus_network_router_firewall_rule_groups** — lists a router's firewall rule groups (id, name, description, external_id, status, priority, group_layer) so a rule group id can be looked up for a firewall rule's parent_id.

## SDK

Updates the bundled Morpheus SDK to expose the new typed fields above.

## Testing

- go build ./..., make lint, and make docs pass; package unit tests pass.
- Acceptance tests for these resources require a live NSX-T network router and are gated accordingly (not run in CI without that infrastructure).


morpheus/framework/datasources/networkrouterfirewallrulegroups/data-source.tf.tmpl
RDavisHPE, line 1: We conventionally call this file example.tf.tmpl - though a few recent examples have been merged with data-source.tf.tmpl
```
$ find . -name data-source.tf.tmpl
./morpheus/sdkv2/datasources/credential/data-source.tf.tmpl
./morpheus/framework/datasources/securitygroups/data-source.tf.tmpl
./morpheus/framework/datasources/networkrouterfirewallrulegroups/data-source.tf.tmpl
./morpheus/framework/datasources/storageservers/data-source.tf.tmpl
```

morpheus/framework/datasources/networkrouterfirewallrulegroups/read.go
```diff
@@ -0,0 +1,222 @@
...
+		rg := rs.RuleGroups[i]
+		if !ruleGroupMatchesFilters(&rg, filters) {
+			continue
+		}
+
+		// Description is NullableString — guard with IsSet() and nil-check Get().
+		description := types.StringNull()
+		if rg.Description.IsSet() && rg.Description.Get() != nil {
+			description = types.StringValue(*rg.Description.Get())
+		}
```
RDavisHPE: Can probably use the convert package here

specs/resources/network_router_nat/config.yaml
```diff
@@ -19,12 +27,8 @@ network_router_nat:
       default: true
     priority:
       int64:
-        computed_optional_required: optional
+        computed_optional_required: computed_optional
         description: Priority of the NAT rule.
-    protocol:
-      string:
-        computed_optional_required: optional
-        description: Protocol for the NAT rule.
```
RDavisHPE, line 27: protocol has been removed here - was this intentional?

morpheus/framework/resources/networkrouterbgpneighbor/schema_gen.go
```diff
@@ -193,8 +194,9 @@ func NetworkRouterBgpNeighborResourceSchema(ctx context.Context) schema.Schema {
 				Description:         "Graceful restart mode (e.g. HELPER_ONLY, GRACEFUL_RESTART, DISABLE)",
```
RDavisHPE: This description should be updated to have GR_AND_HELPER instead of GRACEFUL_RESTART

specs/resources/network_router_nat/config.yaml
eamonnotoole: That was a mistake, which is being fixed.  The field will be reinserted and marked as Deprecated.

eamonnotoole: Fixed in 3ea4a424 (restored in f14024eb): protocol is reinstated and marked deprecated. It stays functional and is still sent to the API for backward compatibility, with a deprecation message pointing to the service attribute.

morpheus/framework/resources/networkrouterbgpneighbor/schema_gen.go
eamonnotoole: Fixed in 3ea4a424: the restart_mode description now reads GR_AND_HELPER (via a description override in the config), and the same correction was made in the API spec.

morpheus/framework/datasources/networkrouterfirewallrulegroups/read.go
eamonnotoole: Done in 3ea4a424: this now uses convert.StrToType(rg.Description.Get()).

morpheus/framework/datasources/networkrouterfirewallrulegroups/data-source.tf.tmpl
eamonnotoole: This data source mirrors hpe_morpheus_storage_servers (the plural pattern I followed), which uses data-source.tf.tmpl, as does securitygroups in your own list above. I kept it consistent with those - happy to rename to example.tf.tmpl if you prefer the older convention.

RDavisHPE: All good

jasonquigleyhpecom (approved):

PR #1263, opened 2026-07-07
MORPH-13817: Fix instance Read panic on 404 from env-vars endpoint

eamonnotoole:
## Summary
`hpe_morpheus_instance` Read panicked with a nil-pointer dereference whenever the environment-variables endpoint returned a non-200. On platforms that don't expose it (e.g. HPE VME, which returns 404), every `apply`/`refresh`/`import` crashed the provider — and on create the VM was provisioned but the post-create Read crashed, leaving it orphaned (not in state).

## Root cause
`getInstanceEnvVars` (`morpheus/framework/resources/instance/instance_read.go`) discarded both the HTTP response and error from `GetEnvVariables(...).Execute()` and then returned `resp.Envs`. On a non-200 the SDK yields a nil `resp`, so `resp.Envs` dereferenced a nil pointer. (Regression: the pre-1.1 code path was nil-safe.)

## Fix
Capture the HTTP response and gate on status — return no env vars unless the call is `200 OK` with a non-nil body:

```go
resp, hresp, _ := client.InstancesAPI.GetEnvVariables(ctx, id).Execute()
if hresp == nil || hresp.StatusCode != http.StatusOK || resp == nil {
    return nil, diags
}
return resp.Envs, diags
```

The SDK's decode error is intentionally ignored — it's unreliable on valid 200 responses (this endpoint carries polymorphic fields the SDK can't model), so HTTP status is the correct signal. This matches the expected behavior: an instance with no env vars, or a platform lacking the endpoint, returns no env vars.

## Testing
- `go build ./...` and `golangci-lint` (instance package) pass.
- The failure path is a 404 from a live VME appliance, not reproducible in unit tests without that platform; the guard is defensive and the happy path (200 + envs) is unchanged.

### DIFF
--- a/morpheus/framework/resources/instance/instance_read.go
+++ b/morpheus/framework/resources/instance/instance_read.go
@@ -1108,13 +1108,28 @@ func getInstanceEnvVars(
 	client *sdk.APIClient,
 ) ([]sdk.GetEnvVariables200ResponseEnvsInner, diag.Diagnostics) {
 	var diags diag.Diagnostics
-	resp, hresp, _ := client.InstancesAPI.GetEnvVariables(ctx, id).Execute()
-	// The SDK can return a decode error on an otherwise-valid 200 response
-	// (polymorphic fields it cannot model), so the error is intentionally ignored
-	// and we gate on the HTTP status instead. Some platforms (e.g. HPE VME) do not
-	// expose this endpoint and return 404; treat a nil response or any non-200
-	// status as "no env vars" rather than dereferencing a nil resp and panicking.
+	resp, hresp, err := client.InstancesAPI.GetEnvVariables(ctx, id).Execute()
+
+	// Some platforms (e.g. HPE VME) do not expose this endpoint and return 404.
+	// That is expected — there simply are no env vars — so return quietly without
+	// a diagnostic (erroring here would fail the whole instance Read, and reading
+	// resp.Envs would panic on the nil resp).
+	if hresp != nil && hresp.StatusCode == http.StatusNotFound {
+		return nil, diags
+	}
+
+	// Any other failure (nil response or a non-200 status) is unexpected. Surface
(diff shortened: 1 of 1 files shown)


morpheus/framework/resources/instance/instance_read.go
```diff
@@ -1108,18 +1108,15 @@ func getInstanceEnvVars(
...
-	// return nil, diags
-	// }
+	resp, hresp, _ := client.InstancesAPI.GetEnvVariables(ctx, id).Execute()
+	// The SDK can return a decode error on an otherwise-valid 200 response
+	// (polymorphic fields it cannot model), so the error is intentionally ignored
+	// and we gate on the HTTP status instead. Some platforms (e.g. HPE VME) do not
+	// expose this endpoint and return 404; treat a nil response or any non-200
+	// status as "no env vars" rather than dereferencing a nil resp and panicking.
+	if hresp == nil || hresp.StatusCode != http.StatusOK || resp == nil {
+		return nil, diags
```
matthew-c-hpe, line 1131: Shouldn't we still build up a diagnostic message for this?

eamonnotoole: Good call — added in af218355. I split the failure handling:

- **404** (endpoint absent, e.g. HPE VME) → return quietly. That's the expected "no env vars" case, and erroring there would fail the whole Read (the crash this PR fixes), so it stays silent.
- **Any other failure** (nil response / non-200) → now surfaces a `Warning` diagnostic with the status + error context, instead of silently dropping env vars.

It's a `Warning` (not `Error`) so the Read still succeeds — env vars are supplementary. `err` is now captured for the message but deliberately isn't part of the gate: the SDK returns a decode error even on valid 200s (polymorphic fields it can't model), so HTTP status stays the primary signal.

matthew-c-hpe (approved):

PR #1267, opened 2026-07-08
fix - storage_volume_type data source - Support looking up by code

matthew-c-hpe:
## Summary

The `hpe_morpheus_storage_volume_type` data source advertised `code` and
`category` as inputs but never used them during read, so anything other than an
`id`/`name` lookup silently ignored them. This change makes `code` and
`category` functional as filters that refine a `name` lookup.

Storage volume type codes are **not** guaranteed unique (there is no unique
constraint on `code`, and the platform's sync identity is `code` + `name`), so
`code` is treated as a filter rather than a standalone lookup key.

## Changes

- **Lookup keys** — a storage volume type is looked up by `id` or `name`. `id`
  and `name` remain mutually exclusive via `ConflictsWith`.
- **`code` / `category` as filters** — when looking up by `name`, an optional
  `code` and/or `category` are passed through to the search as query parameters
  so that a name which is only unique within a code or category resolves to a
  single result. Both stay optional/computed and are populated on read.
- **Name lookup helper** — the by-name lookup logic lives in the data source
  package as a private `getStorageVolumeTypeByName` helper (rather than on the
  SDK client), keeping the data-source-specific search behaviour co-located with
  the data source.
- **Nil guard** — the by-name lookup guards against a nil results list before
  dereferencing it, returning a descriptive error instead of panicking when the
  API returns a response without the `storageVolumeTypes` key. The response
  result is also type-asserted safely.
- **Docs** — regenerated; the data source page notes lookup by `id` or `name`,
  with `code`/`category` available to narrow a `name` lookup.

## Example

```hcl
# Look up by name
data "hpe_morpheus_storage_volume_type" "disk" {
  name = "Kubernetes Volume"
}

# Look up by name, narrowed by code and/or category
data "hpe_morpheus_storage_volume_type" "narrowed" {
  name     = "Standard"
  code     = "standard"
  category = "disk"
}
```

## Verification

- `go build ./...`
- `golangci-lint run` (0 issues)
- `make docs` (idempotent — no drift on re-run)


internal/sdk/legacy/storage_volume_types.go
```diff
@@ -83,3 +83,23 @@ func (client *Client) FindStorageVolumeTypeByName(name string) (*Response, error
...
+	resp, err := client.ListStorageVolumeTypes(&Request{
+		QueryParams: map[string]string{
+			"code": code,
+		},
+	})
+	if err != nil {
+		return resp, err
+	}
+	listResult := resp.Result.(*ListStorageVolumeTypesResult)
+	StorageVolumeTypeCount := len(*listResult.StorageVolumeTypes)
```
jasonquigleyhpecom: 🟠 **Violation** — Missing nil guard. `listResult.StorageVolumeTypes` is dereferenced here without checking if it is `nil`. If the API returns an empty response without the `storageVolumeTypes` key, this will panic.

```go
if listResult.StorageVolumeTypes == nil {
    return resp, fmt.Errorf("found 0 storage volume types with code %v", code)
}
```

morpheus/sdkv2/datasources/storage/storage_volume_type.go
```diff
@@ -22,27 +22,28 @@ func DataSourceStorageVolumeType() *schema.Resource {
...
+				Description:   "The code of the storage volume type",
+				Optional:      true,
+				Computed:      true,
+				ConflictsWith: []string{"id", "name"},
 			},
 			"category": {
 				Type:        schema.TypeString,
 				Description: "The category of the storage volume type",
-				Optional:    true,
 				Computed:    true,
```
jasonquigleyhpecom, line 48: 🔵 **Recommendation** — Removing `Optional: true` from `category` is a backward-incompatible change for any user who previously had it in their configuration. The same applies to adding `ConflictsWith` for `code` above. Consider adding a `BREAKING CHANGE:` block to the PR description so this is captured in the release notes.

jasonquigleyhpecom (changes requested):
---

## 📋 Review: `hpe_morpheus_storage_volume_type`

**PR:** [#1267 — fix - storage_volume_type data source - Support looking up by code](https://github.com/HPE/terraform-provider-hpe/pull/1267)

**Scope:** 🎯 PR diff vs `origin/dev-1.6` — 5 files · agents: patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 1 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 1 |

---

### 🟠 Violations

- **`internal/sdk/legacy/storage_volume_types.go:98`**: Missing nil guard. `listResult.StorageVolumeTypes` is dereferenced to calculate `len()` without checking if it is `nil`. If the API returns an empty response without the `storageVolumeTypes` key, this will cause a panic.

### 🔵 Recommendations

- **`morpheus/sdkv2/datasources/storage/storage_volume_type.go:47`**: The PR correctly makes `category` read-only and adds `ConflictsWith` to `code`, matching actual behavior. However, these are backward-incompatible changes: any users who previously specified `category` (or both `name` and `code`) in their Terraform configurations will now see `terraform plan` failures. Consider adding a `BREAKING CHANGE:` block to the PR description so it is properly captured in the release notes.

- **`internal/sdk/legacy/storage_volume_types.go:78`**: The pre-existing `FindStorageVolumeTypeByName` method has the exact same missing nil guard on `listResult.StorageVolumeTypes`.

---

internal/sdk/legacy/storage_volume_types.go
```diff
@@ -64,17 +64,29 @@ func (client *Client) GetStorageVolumeType(id int64, req *Request) (*Response, e
...
-		QueryParams: map[string]string{
-			"name": name,
-		},
+		QueryParams: queryParams,
 	})
 	if err != nil {
 		return resp, err
 	}
 	listResult := resp.Result.(*ListStorageVolumeTypesResult)
+	if listResult.StorageVolumeTypes == nil {
```
matthew-c-hpe: ✅ **Resolved** (was 🟠 Violation + 🗂️ Pre-existing) — `listResult.StorageVolumeTypes` is now nil-guarded before the `len()` dereference. If the API returns a response without the `storageVolumeTypes` key, this returns a descriptive `found 0 storage volume types named …` error instead of panicking.

The separate `FindStorageVolumeTypeByCode` that carried the original violation was removed in the code-as-filter refactor, so both instances of the missing guard are gone.

morpheus/sdkv2/datasources/storage/storage_volume_type.go
```diff
@@ -22,26 +22,27 @@ func DataSourceStorageVolumeType() *schema.Resource {
...
 				ConflictsWith: []string{"id"},
 			},
 			"code": {
 				Type:        schema.TypeString,
-				Description: "The code of the storage volume type",
+				Description: "The code of the storage volume type. When set alongside name, the lookup is filtered to this code.",
 				Optional:    true,
 				Computed:    true,
 			},
 			"category": {
```
matthew-c-hpe, line 44: ✅ **Resolved** (was 🔵 Recommendation) — the backward-incompatible schema changes were reverted rather than shipped with a `BREAKING CHANGE:` note. `category` is `Optional + Computed` again, and `code` no longer carries `ConflictsWith`. Only `id` ↔ `name` are mutually exclusive, matching pre-PR behavior — existing configs that set `category`, or `name` alongside `code`, are no longer rejected by schema validators.

matthew-c-hpe (commented):
## 🔁 Re-review: prior comments addressed

All findings from the previous review are resolved. Summary of what changed since:

- 🟠 **Nil guard (SDK)** — `FindStorageVolumeTypeByName` now guards `StorageVolumeTypes` against nil before `len()`. The `FindStorageVolumeTypeByCode` method that held the original violation was removed entirely.
- 🔵 **Backward-incompatible schema** — reverted. `category` is Optional+Computed again and `ConflictsWith` is scoped to `id`/`name` only (pre-PR behavior), so no breaking change ships.
- 🎯 **Design** — `code` is no longer a standalone lookup path (codes aren't unique; sync identity is `code`+`name`). Lookup is by `id` or `name`; `code` and `category` are optional query-param filters that narrow a `name` lookup. A `code`-only config correctly errors with `cannot be read without name or id`.

Verification: `go build ./...` ✓ · `golangci-lint run` 0 issues ✓ · `make docs` idempotent ✓

_Informational review (COMMENT). No blocking issues found in the diff._

jasonquigleyhpecom (approved):

PR #1271, opened 2026-07-09
Fix acceptance-test regressions across several resources

jasonquigleyhpecom:
## Summary

Fixes a set of provider bugs that caused acceptance-test failures across several resources. These are the code fixes (idempotency / post-apply consistency / response decoding). Remaining failures are test-data drift in the shared test environment and are addressed separately.

## Fixes

- **Security group** — `resource_permission_groups_all` carried a static `false` default while the API reports `true` when no specific groups are granted access, causing *"Provider produced inconsistent result after apply"* on create. The attribute is now computed from the API response. This also clears the cascading failures in the security group rule resource and the security group / security groups data sources, which build on the security group resource.

- **Provisioning license** — `tenants` could be left unknown after apply when omitted from configuration (*"Provider returned invalid result object after apply"*). Create/Update now keep the user's value when supplied and otherwise use the value returned by the API, so the attribute is always known after apply.

- **Network group** — import produced state that diverged from create/refresh for `tenant_ids` / `resource_permissions` (they were populated only on import), so `ImportStateVerify` failed on the API-assigned owner tenant. Create, refresh and import now behave symmetrically.

- **Write-only secrets** — the API returns only a non-reproducible salted hash for secret fields, so reading it back into state produced a permanent diff (*"non-refresh plan was not empty"*). The configured value is now preserved in state instead of the hash, for: credential (password / api_key / secret_key / client_secret), provisioning setting (cloudinit / windows / pxe_root passwords), docker registry integration (password), and instance-type masked environment variables.

- **Cluster layout data source** — the compute-server virtual image is now modeled as an object (the API returns an object, not a string), so the data source decodes the response instead of failing with a type-conversion error. The bundled Morpheus SDK is updated accordingly, and a subnet-type get-by-id lookup is added.

## Testing

All fixes verified with `TF_ACC` acceptance tests against a live environment; `make lint` (0 issues) and `make docs` (clean) pass. 23 previously-failing acceptance tests now pass.


morpheus/framework/resources/networkgroup/resource.go
```diff
@@ -209,19 +203,13 @@ func (r *networkGroupResource) Read(ctx context.Context, req resource.ReadReques
...
-	if isImport {
-		tenantSet, tenantDiags := networkGroupTenantIdsFromAPI(group.Tenants)
-		resp.Diagnostics.Append(tenantDiags...)
-		state.TenantIds = tenantSet
-
-		rpVal, rpDiags := networkGroupResourcePermissionsFromAPI(group.ResourcePermission)
-		resp.Diagnostics.Append(rpDiags...)
-		state.ResourcePermissions = rpVal
-	}
+	// tenant_ids and resource_permissions are config-managed (Optional) and are
```
rory-mcevoy-hpe, line 206: 🔵 **Recommendation** — The PR drops API population on import for `tenant_ids` and `resource_permissions` to dodge `ImportStateVerify` drift. While this stops the error, it deviates from the standard provider pattern and leaves imported state incomplete. The canonical fix is to make these attributes `computed_optional` in `specs/resources/network_group/config.yaml` so the provider natively tolerates the API-assigned defaults on import.

rory-mcevoy-hpe (commented):
---

## 📋 Review: `hpe_morpheus_various`

**PR:** #1271

**Scope:** 🎯 PR diff vs `origin/dev-1.6` — 8 files · agents: patterns, spec-coverage · UI val: config-check

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 1 |

---

### ⚙️ 4. Pattern Compliance

*   **UI Validation (`security_group`)**: The API implementation returns several fields (like `accounts`, `all`, `allPlans`) that are missing from the API contract.
*   **SDKv2 Masked Values**: The fix applied to write-only hashes across `morpheus/sdkv2/` (preserving `d.Get()` values to prevent permanent diffs) is correct and aligns perfectly with the standard legacy SDK patterns.

---

specs/resources/security_group/config.yaml
```diff
@@ -64,7 +64,6 @@ security_group:
       description: The visibility of the security group.
       default: "private"
     resource_permission_groups_all:
-      default: false
       state_for_unknown: true
```
jasonquigleyhpecom, line 67: 🟡 **Warning (Spec Gap):** The configuration for `resource_permission_groups_all` was updated, but UI validation reveals that the `all` attribute (along with other tenant permission fields like `accounts`) is missing from the OpenAPI specification for `security-groups` creates and updates. It is only present in the API implementation.

jasonquigleyhpecom (commented):
---

## 📋 Review: `terraform-provider-hpe` resources

**PR:** [#1271 — Fix acceptance-test regressions across several resources](https://github.com/HPE/terraform-provider-hpe/pull/1271)

**Scope:** 🎯 PR diff vs `origin/dev-1.6` — 15 files · agents: patterns, spec-coverage · UI val: config-check · connected: [#302](https://github.com/HewlettPackard/morpheus-openapi/pull/302)

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 1 |
| 🔵 Recommendations | 0 |

**Verdict:** 🟡 **PASS WITH WARNINGS** — 1 issues found (0 critical)

---

### 🔗 3. UI Validation

* 🟡 **Warning (Spec Gap):** The configuration for `resource_permission_groups_all` in `security_group` was updated, but UI validation reveals that the `all` attribute (along with other tenant permission fields like `accounts` and `plans`) is missing from the OpenAPI specification for `security-groups` creates and updates. It is only present in the API implementation.

The following issues were observed in the touched `security_group` resource, but were **not** introduced by this PR:

* **Missing Create Attributes:** `custom_options` (which encapsulates `vpc`, `resource_group`, `resource_pool_id`) and `network_server_id` are defined in the OpenAPI create model but are explicitly omitted via `removes` in `config.yaml`.
* **Missing `requires_replace` (`cloud_id`):** Mapped to `ZoneId`, it is sent during create but absent from the update model. It should force recreation but lacks `requires_replace: true`.
* **Erroneous `requires_replace` (`tenant_ids`):** Forces recreation when tenants change, despite `TenantPermissions` being supported in the update model and valid update code existing in `resource.go`.
* **Redundant Null Guard:** In `mapResponseToModel`, `sg.Description` is checked with `.IsSet()` before `.Get()`, which is unnecessary as `convert.StrToType` safely handles `nil`.

---

morpheus/framework/resources/networkgroup/resource.go
jasonquigleyhpecom: Thanks — this is the deliberate choice of the config-managed approach over making these attributes `computed_optional`.

Making `tenant_ids` / `resource_permissions` `computed_optional` risks a "Provider produced inconsistent result after apply" whenever a config *sets* `tenant_ids` and the API returns a superset (e.g. the owner tenant auto-added) — the same class of bug this PR fixes for the `security_group` `all` attribute. Dropping the import-only population keeps create/refresh/import symmetric and config-authoritative; the accepted trade-off is the incomplete imported state noted here.

Worth revisiting the `computed_optional` approach once we can confirm the API does not mutate an explicit `tenant_ids` list. Resolving as intentional.

specs/resources/security_group/config.yaml
jasonquigleyhpecom: Thanks — this looks like a false positive. The generated SDK create/update request models (`AddSecurityGroupsRequestSecurityGroupResourcePermissions` / `UpdateSecurityGroupsRequestSecurityGroupResourcePermissions`) include `all`, `sites`, `allPlans`, and `plans`, and the tenant-permissions model includes `accounts`; `resource.go` sends `all` on both create and update. Since the SDK is generated from the API contract, those fields are present in it — the UI validation appears to be mismatching on envelope/naming rather than flagging a genuine gap.

No change is needed for this PR, which only removes an incorrect default, and the acceptance tests confirm `all` round-trips. Resolving.

RDavisHPE (approved):

PR #1273, opened 2026-07-09
Phase 2: update acceptance test data to the current environment

jasonquigleyhpecom:
## Summary

Test-data fixes for acceptance tests whose hardcoded IDs/names drifted out of sync with the current test environment. Complements the code fixes in #1271.

## Verified green (13 tests)

- **Look-up data sources** — point searches at values that exist in the environment:
  - cluster layout: id `7 → 449`, name → `HVM 1.3 Cluster on HVM/Ubuntu 24.04`
  - network router type: id `4 → 7`, name → `OVS Bridge Domain`
  - subnet type: name → `Esxi Subnet`
  - backup job: name → `Morpheus Appliance`
- **Task operational workflows** — use valid workflow IDs and their canonical names (`797`/`796` and the sdkv2 nested-workflow example).
- **Datastore validation** — remove the obsolete `datastore_type.id` required-argument step (that field is now optional, resolved from `code`).

## Included but not fully verified in-session (4 tests)

- **HVM cluster tests** — layout `219 → 449` (219 is disabled; 449 is the current enabled equivalent). This clears the "Layout not enabled" error, but the tests provision a real cluster and exceed a 25-minute test timeout, so full provisioning wasn't confirmed here.

## Not addressed — environment-blocked

- **cluster namespace ×2** — the environment has no Kubernetes cluster (existing clusters return `ok:false` on the namespaces endpoint).
- **HKS ×1 / instance ×1** — real-infra provisioning (slow / needs specific user-group + storage-profile). These need environment provisioning or CI timeout changes, not test-data edits.

## Notes

Stacked on #1271 — the cluster layout data source tests depend on that PR's SDK decode fix. **Merge #1271 first**; this PR's diff will then collapse to just the two Phase-2 commits.


morpheus/sdkv2/resources/task/task_nested_workflow_resource_example.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/HPE/terraform-provider-hpe/morpheus/testhelpers"
 )
 
-//go:generate sh -c "../../../../bin/render -out examples/resources/morpheus_task_nested_workflow/resource.tf task_nested_workflow_resource.tf.tmpl Code 'tfexample_nested_workflow' Labels '[\"demo\", \"terraform\"]' Name 'tfexample_nested_workflow' OperationalWorkflowId '1' OperationalWorkflowName 'Example workflow'"
+//go:generate sh -c "../../../../bin/render -out examples/resources/morpheus_task_nested_workflow/resource.tf task_nested_workflow_resource.tf.tmpl Code 'tfexample_nested_workflow' Labels '[\"demo\", \"terraform\"]' Name 'tfexample_nested_workflow' OperationalWorkflowId '797' OperationalWorkflowName 'qatf-ttestacc-full-operational-workflow-1'"
```
jasonquigleyhpecom: 🟡 **Warning** — Updating the `go:generate` default values causes test-environment-specific fixture names (`qatf-ttestacc-full-operational-workflow-1`, ID `797`) to leak into the user-facing documentation and examples.

Instead of changing the defaults here, consider keeping the generic `Example workflow` and `1` for documentation, and pass the specific fixture values via the `overrides` map when calling `RenderTaskNestedWorkflowConfig(t, ...)` in `task_nested_workflow_resource_test.go`.

jasonquigleyhpecom (commented):
---

## 📋 Review: `hpe_morpheus_*` Test Fixes

**PR:** [#1273 — Phase 2: update acceptance test data to the current environment](https://github.com/HPE/terraform-provider-hpe/pull/1273)

**Scope:** 🎯 PR diff vs `origin/dev-1.6` — 11 files · agents: patterns, structure · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 1 |
| 🔵 Recommendations | 0 |

---

---

morpheus/sdkv2/resources/task/task_nested_workflow_resource_example.go
jasonquigleyhpecom: Good catch — fixed in 31ca155a. Restored the generic example defaults (`id 1` / `Example workflow` in the `go:generate` line and the defaults map) and moved the environment-specific operational-workflow fixture (`797` / `qatf-ttestacc-full-operational-workflow-1`) into the test's `overrides` map, exactly as suggested. Regenerated the example and docs so the fixture no longer leaks into user-facing output.

morpheus/framework/resources/task/task_test.go
```diff
@@ -17,10 +17,10 @@ import (
...
 
 const (
-	testWorkflowName              = "A Test"
-	testFailureWorkflowName       = "A Failure"
-	testIfOperationalWorkflowId   = "140"
-	testElseOperationalWorkflowId = "102"
+	testWorkflowName              = "qatf-ttestacc-full-operational-workflow-2"
+	testFailureWorkflowName       = "qatf-ttestacc-full-operational-workflow-1"
+	testIfOperationalWorkflowId   = "797"
+	testElseOperationalWorkflowId = "796"
```
RDavisHPE: It might be worth making sure these workflows won't get deleted or are renamed to something without "testacc" in case they are misconstrued as a leftover from a different run and removed. The previous workflows were somewhat more explicit in their naming.

RDavisHPE (dismissed):

RDavisHPE (approved):

PR #1274, opened 2026-07-09
MORPH-12346/12347: Validate hex color format in setting_whitelabel

eamonnotoole:
## Summary

`hpe_morpheus_setting_whitelabel` did not validate the format of `primary_color` and `secondary_color`, so an invalid value such as `"not-a-color"` was silently accepted. This adds plan-time hex-color validation to both attributes.

Fixes MORPH-12346 (`primary_color`) and MORPH-12347 (`secondary_color`).

## Change

Both color attributes now carry a hex-color validator (added via the resource config and regenerated into the schema):

```
^#(?:[0-9a-fA-F]{3}|[0-9a-fA-F]{4}|[0-9a-fA-F]{6}|[0-9a-fA-F]{8})$
```

- Accepts 3/4/6/8-digit hex, for example `#fff`, `#1a73e8`, `#1a73e8ff`
- Rejects non-hex values (`not-a-color`, `red`, `rgb(...)`, missing `#`, and so on) at plan time, with a clear message

## Tests

- Unit test exercising both validators directly (no live backend), covering valid and invalid forms
- Acceptance tests `TestAccMorpheusSettingWhitelabel_validationInvalidPrimaryColor` / `...InvalidSecondaryColor` asserting the plan-time error

## Verification

- `go build ./...`, `go vet`, `gofmt`, `golangci-lint` — all clean
- Unit tests pass; `make docs` produces no diff

### DIFF
--- a/specs/resources/setting_whitelabel/config.yaml
+++ b/specs/resources/setting_whitelabel/config.yaml
@@ -76,26 +76,16 @@ setting_whitelabel:
       validators:
         - custom:
             imports:
-              - path: regexp
-              - path: github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator
-            schema_definition: |-
-              stringvalidator.RegexMatches(
-              regexp.MustCompile(`^#(?:[0-9a-fA-F]{3}|[0-9a-fA-F]{4}|[0-9a-fA-F]{6}|[0-9a-fA-F]{8})$`),
-              "must be a valid hex color code, for example #1a73e8, #fff, or #1a73e8ff",
-              )
+              - path: github.com/HPE/terraform-provider-hpe/utils/validators
+            schema_definition: validators.HexColor()
     secondary_color:
       computed_optional_required: optional
       description: The header background color.
       validators:
         - custom:
             imports:
-              - path: regexp
-              - path: github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator
-            schema_definition: |-
-              stringvalidator.RegexMatches(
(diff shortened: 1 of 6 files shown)


specs/resources/setting_whitelabel/config.yaml
```diff
@@ -73,9 +73,29 @@ setting_whitelabel:
...
       computed_optional_required: optional
       description: The primary button background color.
+      validators:
+        - custom:
+            imports:
+              - path: regexp
+              - path: github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator
+            schema_definition: |-
+              stringvalidator.RegexMatches(
+              regexp.MustCompile(`^#(?:[0-9a-fA-F]{3}|[0-9a-fA-F]{4}|[0-9a-fA-F]{6}|[0-9a-fA-F]{8})$`),
```
matthew-c-hpe: Could we implement a custom plan modifier for this?

eamonnotoole: Good call on pulling this out of the inline block. I extracted it into a named custom validator, `validators.HexColor()` (in `utils/validators`, mirroring the existing `validators.Lowercase()`), referenced from the config as `schema_definition: validators.HexColor()` — so the schema no longer carries the inline regex.

I kept it as a validator rather than a plan modifier: the goal here is to reject an invalid color, which is what validators are for (they run at `ValidateResourceConfig`, so they also fire during `terraform validate`, before a plan). The role of a plan modifier is to mutate the planned value, so one that only raised an error would be doing validation work in the wrong place. Happy to reconsider if you had a specific normalization behaviour in mind.

Covered by `TestUnitHexColor` (validator logic) and `TestUnitSettingWhitelabelColorValidators` (schema wiring). Updated in 1d55b2bd.

matthew-c-hpe (approved):

PR #1280, opened 2026-07-10
Validate storage_volume max_storage for type_id at plan time

eamonnotoole:
## Summary

Two `hpe_morpheus_storage_volume` improvements surfaced while investigating MORPH-13021 (which itself turned out to be a stale build plus incomplete config, not a provider defect — this PR hardens two adjacent gaps).

## 1. Plan-time `max_storage` validation for `type_id`

The static `max_storage` attribute validator enforces the HPE Alletra MP / 9000 1-65536 GiB bound only when the type is given as `type_code` — a schema validator has no API client to resolve a numeric `type_id`. This adds a `ModifyPlan` that, when only `type_id` is set, resolves it to its type code and applies the same bound, so an out-of-range size is reported at **plan** time instead of as a 400 from the API at apply.

- Reads only the planned config (never state), so it runs on create and update.
- Best effort: if the client is unavailable or the type lookup fails, the plan is **not** blocked — the backend still enforces the limit on apply.
- Reuses the existing `maxStorageSizeError` decision helper.

## 2. Complete example/test now includes the required datastore

An Alletra MP LUN requires a data store (supplied via the write-only `config_alletramp_bmaas` block). The `example_complete` example and `TestAccMorpheusStorageVolumeResourceCompleteOk` omitted it, so they could not actually create a volume on real Alletra hardware (they only pass because they are `capabilities.Alletra`-gated and skipped). This adds the datastore block; the import step ignores the write-only version trigger, which the API does not return.

## Tests

- New `ModifyPlan` unit test (httptest-backed client): `type_id` oversize -> plan error, `type_id` in range -> ok, `type_code` -> no API call, no `max_storage` -> no API call.
- Existing `maxStorageSizeError` table unchanged.

## Verification

- `go build ./...`, `go vet`, `gofmt`, `golangci-lint` — all clean
- Unit tests pass; `make docs` produces no diff


morpheus/framework/resources/storagevolume/modify_plan_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/tfsdk"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
+
+	"github.com/HPE/terraform-provider-hpe/morpheus/model"
+	"github.com/HPE/terraform-provider-hpe/morpheus/utils/clientfactory"
+)
+
+// number builds a tftypes numeric value for an int64.
+func number(n int64) tftypes.Value {
```
matthew-c-hpe: Use `types.Int64Value()`
```
	"github.com/hashicorp/terraform-plugin-framework/types"
```

jasonquigleyhpecom: Good call — done. Dropped the raw `number()`/`tftypes`/`math/big` helper and now build the plan from a typed `StorageVolumeModel` via `plan.Set(...)`, populating values with `types.Int64Value()` / `types.StringValue()` (matching the pattern used elsewhere in the package). The write-only `config_alletramp_bmaas` block and dynamic `config` are set to explicit nulls so the plan still matches the schema.

gofmt/vet/build/golangci-lint clean; all four sub-cases still pass.

matthew-c-hpe (approved):

PR #1283, opened 2026-07-13
MORPH-12625: Upload whitelabel logos via the images endpoint

rory-mcevoy-hpe:
## Summary

`hpe_morpheus_setting_whitelabel` accepted `header_logo`, `footer_logo`, `login_logo` and `favicon` on apply, but the settings API silently ignored those fields, so they were read back as `null` on the next refresh — the "produced an unexpected new value ... but now null" error (MORPH-12625).

Those four fields are upload-only: Morpheus stores them as uploaded image attachments (served back as a generated storage URL) and they can only be set through the multipart whitelabel *images* endpoint, not the JSON settings body.

## What changed (non-breaking)

The four existing attributes are **kept as-is** (optional strings) — no schema attributes are removed, renamed, or made write-only. Only their behavior is fixed:

- Each value is now treated as a **local image file path** and uploaded via the multipart images endpoint on create/update; clearing a path resets that image.
- The provider no longer overwrites the configured value with the API's storage URL on read, so the value round-trips consistently (this is what fixes the error).

Logos accept `png`/`jpg`/`svg`; the favicon accepts `ico`/`png`. Remote URLs are not supported (the API accepts uploaded file content only).

## Testing

- `go build ./...`, lint (package clean), unit tests, and `make docs` (idempotent) all pass.
- Added acceptance test `TestAccMorpheusSettingWhitelabelResourceImagesOk`: uploads all four images, asserts a clean re-plan (the regression), and resets them on removal. Generates real PNG fixtures at runtime.

## Limitations

- Local file paths only (no remote URLs).
- A file-content change at the same path is not auto-detected; change the path or use `terraform apply -replace=...` to re-upload.
- Server-side image changes are not reconciled back into state.

Resolves MORPH-12625.


morpheus/framework/resources/settingwhitelabel/resource_images_test.go
```diff
@@ -0,0 +1,130 @@
...
+			{Config: withoutImages, Check: resetChecks},
+			{Config: withoutImages, ExpectNonEmptyPlan: false, PlanOnly: true},
+		},
+	})
+}
+
+// writeTestPNG writes a minimal valid 1x1 PNG image to dir/name and returns its
+// absolute path. A real PNG is used (rather than a placeholder) so the Morpheus
+// server accepts it against the whitelabel image contentType constraints.
+func writeTestPNG(t *testing.T, dir, name string) string {
```
matthew-c-hpe: Move this to testhelpers.WritePNG

morpheus/framework/resources/settingwhitelabel/resource_images_test.go
```diff
@@ -0,0 +1,130 @@
...
+	loginLogo := writeTestPNG(t, dir, "login.png")
+	// The favicon field also accepts image/png per the Morpheus domain
+	// contentType constraint, so a PNG fixture is valid here too.
+	favicon := writeTestPNG(t, dir, "favicon.ico")
+
+	resourceName := "hpe_morpheus_setting_whitelabel.images"
+
+	withImages := providerConfig + fmt.Sprintf(`
+resource "hpe_morpheus_setting_whitelabel" "images" {
+  enabled        = true
```
matthew-c-hpe: Don't enable it in this test

morpheus/framework/resources/settingwhitelabel/resource_images_test.go
```diff
@@ -0,0 +1,130 @@
...
+  appliance_name = %q
```
matthew-c-hpe: Don't set appliance name in this test.

morpheus/framework/resources/settingwhitelabel/resource_images_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+
+	"github.com/HPE/terraform-provider-hpe/morpheus/testhelpers"
+	"github.com/HPE/terraform-provider-hpe/morpheus/testhelpers/capabilities"
+	"github.com/HPE/terraform-provider-hpe/provider/adapter"
+)
+
+// TestAccMorpheusSettingWhitelabelResourceImagesOk is the regression test for
+// MORPH-12625. header_logo, footer_logo, login_logo and favicon are local image
```
matthew-c-hpe: Remove this comment

morpheus/framework/resources/settingwhitelabel/resource_images_test.go
```diff
@@ -0,0 +1,130 @@
...
+// file paths that the provider uploads via the multipart images endpoint.
+// Previously they were sent (and silently dropped) via the JSON settings
+// endpoint and read back as null, producing a "provider produced an unexpected
+// new value ... but now null" error on apply.
+//
+// Step 1 applies a config that sets all four images (this alone reproduced the
+// original failure). Step 2 is a plan-only step asserting the follow-up plan is
+// empty (the configured paths round-trip consistently). Steps 3-4 remove the
+// images and assert they are reset.
+func TestAccMorpheusSettingWhitelabelResourceImagesOk(t *testing.T) {
```
matthew-c-hpe: Move this test to resource_test.go

morpheus/framework/resources/settingwhitelabel/resource.go
```diff
@@ -219,6 +237,108 @@ func (r *settingWhitelabelResource) readIntoModel(
...
+	plan *SettingWhitelabelModel,
+	prior *SettingWhitelabelModel,
+	diagnostics *diag.Diagnostics,
+) {
+	imgReq := client.WhitelabelSettingsAPI.UpdateWhitelabelImages(ctx)
+
+	var openFiles []*os.File
+	defer func() {
+		for _, f := range openFiles {
+			_ = f.Close()
```
matthew-c-hpe: Just f.Close()

morpheus/framework/resources/settingwhitelabel/resource.go
```diff
@@ -219,6 +237,108 @@ func (r *settingWhitelabelResource) readIntoModel(
...
+		}
+	}()
+
+	upload := false
+	addFile := func(pathVal types.String, attach func(*os.File)) bool {
```
matthew-c-hpe: Simplify this, let's try break it out into a separate helper function that returns `err` instead of the bool. There's too much indirection to follow this clearly.

We can swap the func(*os.File) argument with an `ApiUpdateWhitelabelImagesRequest` I think.

morpheus/framework/resources/settingwhitelabel/resource.go
```diff
@@ -219,6 +237,108 @@ func (r *settingWhitelabelResource) readIntoModel(
...
+// server-generated storage URL that never matches the supplied path, so these
+// values are carried through from the plan and are never reconciled from the
+// read response.
+func (r *settingWhitelabelResource) uploadImages(
+	ctx context.Context,
+	client *sdk.APIClient,
+	plan *SettingWhitelabelModel,
+	prior *SettingWhitelabelModel,
+	diagnostics *diag.Diagnostics,
+) {
```
matthew-c-hpe: This should return `err`, and set diagnostics in the caller.

morpheus/framework/resources/settingwhitelabel/resource.go
```diff
@@ -219,6 +237,108 @@ func (r *settingWhitelabelResource) readIntoModel(
...
+		if err != nil {
+			diagnostics.AddError("could not read whitelabel image file", err.Error())
+
+			return false
+		}
+		openFiles = append(openFiles, f)
+		attach(f)
+		upload = true
+
+		return true
```
matthew-c-hpe: setting upload to true and returning true  is pointless,

we could as easily do:

```
	if upload = addFile(plan.HeaderLogo, func(f *os.File) { imgReq = imgReq.HeaderLogoFile(f) }); !upload {
		return
	}
```

But this should be adapted for addFile to use `err`

morpheus/framework/resources/settingwhitelabel/resource.go
```diff
@@ -219,6 +237,108 @@ func (r *settingWhitelabelResource) readIntoModel(
...
+
+			return
+		}
+	}
+
+	if prior == nil {
+		return
+	}
+
+	// Reset images that were set previously but cleared in the new plan. The JSON
```
matthew-c-hpe: Can we break this out into a reset function?

If we detect a difference in plan and state, do reset then upload.

It's getting a bit too complicated.

morpheus/framework/resources/settingwhitelabel/resource.go
```diff
@@ -219,6 +237,108 @@ func (r *settingWhitelabelResource) readIntoModel(
...
+	// settings endpoint honours the reset flags (the multipart endpoint's favicon
+	// reset field is misnamed in the spec), so route resets there.
+	resets := sdk.UpdateWhitelabelSettingsRequestWhitelabelSettings{}
+	needReset := false
+	clearImage := func(planVal, priorVal types.String, setReset func()) {
```
matthew-c-hpe: Again, too much indirection.

matthew-c-hpe (changes requested):

matthew-c-hpe (commented):
Let's add plan-time validation for the correct file extensions. Add a new validator to `utils/validators`. For checking file extensions which accepts a []string slice of valid file extensions for that string value.

This is so we can validate the user has set a file with extension e.g. `.png`.

We can probably use `https://pkg.go.dev/path/filepath#Ext`

templates/resources/morpheus_setting_whitelabel.md.tmpl
```diff
@@ -12,8 +12,47 @@ description: |-
...
+  uploaded file content only (there is no server-side URL fetch).
+- **File content changes are not detected.** If a file's contents change but its
+  path stays the same, Terraform will not detect a change and will not re-upload.
+  Force a re-upload by changing the path or running
+  `terraform apply -replace=hpe_morpheus_setting_whitelabel.<name>`.
+- **Removing a path does not clear the image.** Clearing or removing one of these
+  attributes leaves the previously uploaded image in place on the appliance.
+  Destroy the resource (which resets all images) or remove the image in the
+  Morpheus UI to take one down.
+- **Server-side changes are not reconciled.** Logos changed or removed directly in
```
matthew-c-hpe: Remove this point, this is a Terraform issue / quirk

morpheus/framework/resources/settingwhitelabel/resource.go
```diff
@@ -219,9 +235,95 @@ func (r *settingWhitelabelResource) readIntoModel(
...
+	}
+
+	imgReq := client.WhitelabelSettingsAPI.UpdateWhitelabelImages(ctx)
+
+	if !plan.HeaderLogo.IsNull() && !plan.HeaderLogo.IsUnknown() {
+		headerLogo, err := openImageFile(plan.HeaderLogo)
+		if err != nil {
+			return err
+		}
+		if headerLogo != nil {
```
matthew-c-hpe: We don't need to check for nil on any of these, I believe os.Open guarantees non-nil file pointer in a non-error scenario. It's only nil if err != nil from calling os.Open.

templates/resources/morpheus_setting_whitelabel.md.tmpl
```diff
@@ -12,8 +12,47 @@ description: |-
...
+
+- **Local file paths only.** `header_logo`, `footer_logo`, `login_logo` and
+  `favicon` must reference image files on the machine running Terraform. Remote
+  URLs are not supported because the Morpheus whitelabel images API accepts
+  uploaded file content only (there is no server-side URL fetch).
+- **File content changes are not detected.** If a file's contents change but its
+  path stays the same, Terraform will not detect a change and will not re-upload.
+  Force a re-upload by changing the path or running
+  `terraform apply -replace=hpe_morpheus_setting_whitelabel.<name>`.
+- **Removing a path does not clear the image.** Clearing or removing one of these
```
matthew-c-hpe: Removing a path should clear the image. 

In update, after we run `uploadImages`:

```
	if plan.FooterLogo.IsNull() {
		body.WhitelabelSettings.ResetFooterLogo = boolPtr(true)
	}
```

templates/resources/morpheus_setting_whitelabel.md.tmpl
```diff
@@ -12,8 +12,47 @@ description: |-
...
+  attributes leaves the previously uploaded image in place on the appliance.
+  Destroy the resource (which resets all images) or remove the image in the
+  Morpheus UI to take one down.
+- **Server-side changes are not reconciled.** Logos changed or removed directly in
+  the Morpheus UI/API are not read back into Terraform state.
+- **Uploaded values are not round-tripped.** The value you set is a local path; the
+  API stores the file and returns a different server-generated storage URL, which
+  is intentionally not surfaced as the attribute value.
+- **Import cannot populate image paths.** The API only exposes the stored image
```
matthew-c-hpe: Remove this, import is not supported on this resource

matthew-c-hpe (dismissed):

matthew-c-hpe (approved):

PR #1284, opened 2026-07-14
Fix acceptance-test regressions and gate infra-dependent tests

jasonquigleyhpecom:
## Summary

Fixes a set of acceptance-test regressions surfaced by a full `TF_ACC` run, and gates tests that depend on infrastructure not yet available on the test appliance so they **skip** cleanly instead of failing.

### Fixes (now passing)
- **Image data source** — guard an empty virtual-image `config` before the marshal/unmarshal round-trip. The generated oneOf wrapper marshals to nothing when no variant is set, which `encoding/json` rejects as *"unexpected end of JSON input"*; images with an empty config hit this. (MORPH-14029)
- **Instance `user_group`** — refresh the stale hard-coded user-group ID in `TestAccMorpheusInstanceResourceUserGroupStorageProfile`. (MORPH-14033)
- **Template label race** — the file- and script-template data source tests ran in parallel with identical labels `["demo","template","terraform"]`; concurrent creation of the same new label hit the account-wide uniqueness constraint (`labels[N].name: must be unique`). The two data source tests and the file-template resource test now use unique per-test labels. (MORPH-14166)

### Gated (skip until infra is available)
- **`kubernetes_cluster` capability (new)** — the `cluster_namespace` resource/data source tests and the HKS-HVM cluster test require a live, namespace-capable Kubernetes cluster, which the current test environment doesn't provide. They are re-gated on this capability, and the cluster ID is now sourced from `TF_VAR_testacc_morpheus_cluster_id`. This keeps the ~15 Kubernetes *library* tests (spec templates, blueprints, cluster layouts/types) running. (MORPH-14164, MORPH-14165)
- **`network_loadbalancer_haproxy` capability (new)** — the HAProxy load balancer tests provision an HAProxy *container*, which needs the `load-balancer-haproxy-1.7` instance-type/layout seeded and a cloud able to provision the container; neither is present, so provisioning fails server-side. The two tests are gated on this capability. The stale resource-pool value was also refreshed (`pool-574` → `pool-1`) so the tests are correct once the layout exists.

## Verification
- `make lint` clean (0 issues); `go build ./...` and `go vet` clean
- `make docs` regenerated (only the load balancer pool value changed) and idempotent
- Acceptance reruns against the test appliance:
  - **PASS:** image data source, instance `user_group`, file + script template data sources (run in parallel)
  - **SKIP (intended):** 5× `cluster_namespace`, HKS-HVM, 2× HAProxy load balancer

## Related
- MORPH-14033 — refresh stale acceptance-test environment IDs
- MORPH-14029 — image / virtual-image config decode failure
- MORPH-14164 — `cluster_namespace` tests need a namespace-capable cluster
- MORPH-14165 — HKS HVM cluster provisions with no workers
- MORPH-14166 — template tests label-uniqueness race

### DIFF
--- a/morpheus/testhelpers/config.go
+++ b/morpheus/testhelpers/config.go
@@ -5,7 +5,7 @@ import "os"
 // EnvKubernetesClusterID is the environment variable that supplies the ID of a
 // namespace-capable Kubernetes cluster for tests gated on the
 // kubernetes_cluster capability (cluster namespaces, HKS provisioning).
-const EnvKubernetesClusterID = "TF_VAR_testacc_morpheus_cluster_id"
+const EnvKubernetesClusterID = "TF_VAR_testacc_morpheus_k8s_cluster_id"
 
 // KubernetesClusterID returns the Kubernetes cluster ID to use for
 // namespace/HKS tests, taking TF_VAR_testacc_morpheus_cluster_id when set and
--- a/morpheus/testhelpers/config.go
+++ b/morpheus/testhelpers/config.go
@@ -1,5 +1,25 @@
 package testhelpers
 
+import "os"
+
+// EnvKubernetesClusterID is the environment variable that supplies the ID of a
+// namespace-capable Kubernetes cluster for tests gated on the
+// kubernetes_cluster capability (cluster namespaces, HKS provisioning).
+const EnvKubernetesClusterID = "TF_VAR_testacc_morpheus_cluster_id"
+
+// KubernetesClusterID returns the Kubernetes cluster ID to use for
+// namespace/HKS tests, taking TF_VAR_testacc_morpheus_cluster_id when set and
+// otherwise falling back to the supplied value. These tests only execute when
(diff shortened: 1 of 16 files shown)


morpheus/testhelpers/config.go
```diff
@@ -1,5 +1,25 @@
 package testhelpers
 
+import "os"
+
+// EnvKubernetesClusterID is the environment variable that supplies the ID of a
+// namespace-capable Kubernetes cluster for tests gated on the
+// kubernetes_cluster capability (cluster namespaces, HKS provisioning).
+const EnvKubernetesClusterID = "TF_VAR_testacc_morpheus_cluster_id"
```
matthew-c-hpe: Suggestion - `TF_VAR_testacc_morpheus_k8s_cluster_id`

Could feasibly see us wanting this sort of env var for HVM clusters too.

RDavisHPE (dismissed):

eoghanino (dismissed):
LGTM

matthew-c-hpe (approved):

PR #1285, opened 2026-07-14
Adding network_router_firewall_rule_group resource

RDavisHPE:
Implements full CRUD, import, and acceptance tests for NSX-T gateway firewall policy rule groups on a Morpheus network router.
**What's included**
- Resource: hpe_morpheus_network_router_firewall_rule_group (package networkrouterfirewallrulegroup)
- Import format: router_id.group_id (e.g. 1755.995)
- Sweeper, example config, and .md.tmpl docs
- specs/resources/network_router_firewall_rule_group/config.yaml codegen spec

**Key design decisions**
- external_type removed from schema — always "GatewayPolicy" for gateway firewall; hardcoded in Create/Update. ("SecurityPolicy" belongs to the distributed firewall resource hpe_morpheus_network_firewall_rule_group.) This is what is present in the hpegl provider.
- group_layer is immutable — validated against OneOf("Emergency", "SharedPreRules", "LocalGatewayRules") (per NsxtNetworkService.groovy); carries RequiresReplace + UseStateForUnknown so omitting it in an update config does not spuriously trigger destroy+create.
- Write-only visibility and tenant_ids — the single GET endpoint does not return these fields; values are preserved from plan/prior state in getRuleGroupAsState. On import both are null; users re-apply to restore them. tenant_ids guards against unknown-after-apply (Optional+Computed with no Default resolves to unknown on new-resource Create) by substituting an empty set.
- status and external_id exposed as computed — external_id is the NSX-T string (e.g. "group-123") used as parent_id on hpe_morpheus_network_router_firewall_rule.

**Acceptance tests**
Three tests, all passing:
```
--- PASS: TestAccMorpheusNetworkRouterFirewallRuleGroupResourceImportOk (10.74s)
--- PASS: TestAccMorpheusNetworkRouterFirewallRuleGroupResourceExampleOk (13.79s)
--- PASS: TestAccMorpheusNetworkRouterFirewallRuleGroupResourceUpdateOk (16.64s)
PASS
```


jasonquigleyhpecom (changes requested):
---

## 📋 Review: `hpe_morpheus_network_router_firewall_rule_group`

**Scope:** 🎯 PR diff vs `origin/dev-1.6` — 29 files · agents: patterns, spec-coverage, structure · UI val: config-check · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 3 |
| 🟠 Violations | 2 |
| 🟡 Warnings | 2 |
| 🔵 Recommendations | 0 |

**Verdict:** 🔴 **FAIL** — 7 issues found (5 critical)

---

### 🔴 1. Gaps

- **`morpheus/framework/resources/networkrouterfirewallrulegroup`**: Missing corresponding singular data source. Every resource must have a paired data source implementation (`datasource.go`, `schema_gen.go`, `*_test.go`), and it must be registered in `morpheus/datasources.go`.
- **`morpheus/framework/resources/networkrouterfirewallrulegroup/sweep/sweep.go:83`**: The `networkrouterfirewallrulegroup` resource acts as a parent for `networkrouterfirewallrule` (the rule uses the group's `external_id` as its `parent_id`). The sweeper is missing the `testsweep.WithDependencies[ruleGroupSweeperItem]("hpe_morpheus_network_router_firewall_rule")` configuration to ensure child rules are deleted before the group.
- **`morpheus/framework/resources/networkrouterfirewallrulegroup/sweep_import.go:9`**: Missing the blank import for the dependent child sweeper `_ "github.com/HPE/terraform-provider-hpe/morpheus/framework/resources/networkrouterfirewallrule/sweep"`.

### 🟠 2. Violations

- **`morpheus/sdkv2/resources/task/resource_task_powershell_script.go:146`**: Line length exceeds the 120 character limit (131 chars) on the description for `remote_target_password_wo`.
- **`morpheus/sdkv2/resources/task/resource_task_shell_script.go:161`**: Line length exceeds the 120 character limit (131 chars) on the description for `remote_target_password_wo`.

### 🟡 3. Warnings

- **`morpheus/framework/resources/networkrouterfirewallrulegroup/import.go:60`**: The comment on line 53 states "On import these fields will be null", but `prior.TenantIds` is seeded with an empty set here rather than null. (Consider updating the comment to reflect the PR's intentional design decision instead of changing this code).
- **`morpheus/framework/resources/networkrouterfirewallrulegroup/read.go:72`**: The API contract lists `visibility` and `tenants` in the read schema. If the GET API does not actually return them, this indicates a gap in the contract that should be corrected upstream or marked as write-only.

### 🗂️ 4. Pre-existing (out of PR scope)

- **`morpheus/framework/resources/networkrouter/sweep/sweep.go`**: The parent `networkrouter` sweeper was not touched in this PR, but it should be updated to depend on the newly added `hpe_morpheus_network_router_firewall_rule_group` sweeper, ensuring groups are deleted before the parent router during acceptance test cleanups.

---

specs/resources/network_router_firewall_rule_group/config.yaml
```diff
@@ -0,0 +1,69 @@
...
+      state_for_unknown: true
+    router_id:
+      computed_optional_required: required
+      description: The ID of the parent network router
+      requires_replace: true
+    name:
+      computed_optional_required: required
+      description: Network router firewall rule group name
+    description:
+      computed_optional_required: computed_optional
```
matthew-c-hpe, line 39: 🔴 **Gap:** `description`, `priority`, and `tenant_ids` are overridden as `computed_optional`, but the API does not independently compute these. Setting them to computed prevents clearing the fields, as Terraform will substitute an `Unknown` plan value when they are removed from the config. These must be changed to `optional`.

morpheus/framework/resources/networkrouterfirewallrulegroup/update.go
```diff
@@ -0,0 +1,116 @@
...
+	routerID := plan.RouterId.ValueInt64()
+
+	// The SDK's UpdateNetworkRouterFirewallRuleGroupRequest.RuleGroup is typed
+	// as map[string]interface{} — build it manually.
+	ruleGroup := map[string]interface{}{
+		"name":         plan.Name.ValueString(),
+		"externalType": "GatewayPolicy",
+	}
+
+	if !plan.Description.IsNull() && !plan.Description.IsUnknown() {
```
matthew-c-hpe, line 51: 🔴 **Gap:** This check creates a partial update bug. If a user deletes the description (or priority/tenants) from their configuration to clear it, `plan.Description` resolves to `Unknown` (due to the `computed_optional` setting in config.yaml) or `Null`, both of which skip this block. As a result, the clearing value is never sent to the backend. Explicitly handle the clearance (e.g. `""` for strings, `[]` for lists).

morpheus/framework/resources/networkrouterfirewallrulegroup/read.go
```diff
@@ -0,0 +1,126 @@
...
+
+	// Write-only fields: the single GET response does not include these.
+	// Preserve the plan/prior-state values so Terraform does not see spurious
+	// diffs. On import these will be null — users must re-apply to set them.
+	//
+	// Guard against unknown: on Create the plan value for an Optional+Computed
+	// attribute with no Default is unknown (no prior state exists). Propagating
+	// unknown into state is rejected by the framework. Resolve to an empty set
+	// in that case — the semantics are identical (no tenant restriction).
+	state.Visibility = prior.Visibility
```
matthew-c-hpe: 🟠 **Violation:** Unconditional ignore of SDK response field. While the API may currently omit `visibility` and `tenants` on individual GET requests, these are typed fields in the SDK struct. Fallback to `prior` should only occur if the SDK value is actually `nil`, so the provider naturally recovers if the API behavior is patched in the future.

```go
if rg.Visibility != nil {
    state.Visibility = convert.StrToType(rg.Visibility)
} else {
    state.Visibility = prior.Visibility
}
```

matthew-c-hpe (changes requested):
## 📋 Review: `hpe_morpheus_network_router_firewall_rule_group`

**PR:** [#1285 — Adding network_router_firewall_rule_group resource](https://github.com/HPE/terraform-provider-hpe/pull/1285)

**Scope:** 🎯 PR diff vs `origin/dev-1.7` — 20 files · agents: patterns, spec-coverage, structure · UI val: config-check · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 3 |
| 🟠 Violations | 1 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

**Verdict:** 🔴 **FAIL** — 4 issues found (4 critical)

### 📁 1. File Completeness

*   🔴 **Gap:** Missing singular data source for `hpe_morpheus_network_router_firewall_rule_group` (expected `morpheus/framework/datasources/networkrouterfirewallrulegroup/`). While a plural data source exists, every resource must be paired with a corresponding single data source to allow lookup by ID or name.
*   🔴 **Gap:** Missing data source registration in `morpheus/datasources.go` for the new singular data source.

### ⚙️ 4. Pattern Compliance & Attribute Coverage

*(See inline comments for specifics on partial updates and SDK usage).*

---

**Next:** Fix 🔴 Gaps and 🟠 Violations before merge.

specs/resources/network_router_firewall_rule_group/config.yaml
RDavisHPE: Fixed — changed `description`, `priority`, and `tenant_ids` to `optional` in config.yaml, then ran `task recreate-resources -- network_router_firewall_rule_group` + `task generate` to regenerate `schema_gen.go`. All three attributes no longer carry `Computed: true`.

morpheus/framework/resources/networkrouterfirewallrulegroup/update.go
RDavisHPE: Fixed — removed the `IsNull()` gate from `description` and `priority` so `ValueString()`/`ValueInt64()` send `""` / `0` as the clearing value when the field is absent from config. For `tenant_ids`, added an inner `!IsNull()` guard before `ElementsAs` so a null set sends `tenants: []` to clear the restriction.

morpheus/framework/resources/networkrouterfirewallrulegroup/read.go
RDavisHPE: Fixed — replaced the unconditional `state.Visibility = prior.Visibility` with a nil-check as suggested. Applied the same pattern to `rg.Tenants` / `tenant_ids` for consistency. Also removed the now-dead `IsUnknown` guard that was there to handle the old `computed_optional` Unknown case.

specs/resources/network_router_firewall_rule_group/config.yaml
RDavisHPE: Follow-up: acceptance tests revealed that the API always returns `priority: 0` as a Go zero value even when not explicitly set, so `optional` caused an inconsistent-result error (plan=null, state=0). Reverted `priority` to `computed_optional` and added `state_for_unknown: true` — this means `UseStateForUnknown()` is now on `priority` in the schema, preventing always-unknown-after-apply on subsequent plans while allowing the framework to accept the API's 0 on first apply.

`description` and `tenant_ids` remain `optional` as intended.

morpheus/framework/resources/networkrouterfirewallrulegroup/update.go
RDavisHPE: The `!IsUnknown()` guards remain in place. For `priority`, `UseStateForUnknown` ensures the plan value on updates is always the prior state value (never null), so the guard correctly sends it every time. The `description` and `tenant_ids` clearing paths are unchanged.

morpheus/framework/resources/networkrouterfirewallrulegroup/read.go
RDavisHPE: The `visibility` nil-check stays. However, applying the same pattern to `rg.Tenants` uncovered that the API *does* return tenants — specifically, it always returns the root tenant (id=1) regardless of what was configured. Since the returned slice doesn't represent the user-configured restriction, we reverted `tenant_ids` back to unconditional prior preservation. The write-only pattern was correct for this field.

morpheus/resources.go
```diff
@@ -114,6 +115,7 @@ func (p *MorpheusProvider) Resources(
 		networkpoolserver.NewResource,
 		networkrouternat.NewResource,
 		networkrouterfirewallrule.NewResource,
+		networkrouterfirewallrulegroup.NewResource,
```
jasonquigleyhpecom, line 118: 🔴 **Gap**: The new resource `networkrouterfirewallrulegroup` is registered here, but it is missing its paired singular data source. The codebase standard requires every resource to have a corresponding singular data source (e.g. `morpheus/framework/datasources/networkrouterfirewallrulegroup/`) with full test coverage, examples, and documentation. Please add it and register it in `morpheus/datasources.go`.

morpheus/framework/resources/networkrouterfirewallrulegroup/create.go
```diff
@@ -0,0 +1,141 @@
...
+		resp.Diagnostics.AddError(createOperation, "failed to create client: "+err.Error())
+
+		return
+	}
+
+	routerID := plan.RouterId.ValueInt64()
+
+	ruleGroup := sdk.CreateNetworkRouterFirewallRuleGroupRequestRuleGroup{
+		Name:         plan.Name.ValueString(),
+		ExternalType: "GatewayPolicy",
```
jasonquigleyhpecom, line 44: 🔵 **Recommendation**: Hardcoding `ExternalType` to `"GatewayPolicy"` here (and explicitly removing it from the Terraform schema) is a deliberate design decision to reserve `"SecurityPolicy"` for a separate distributed firewall resource. This is acceptable, but consider adding a brief note in the resource description mentioning that this resource strictly configures gateway policies, to clarify the behaviour for users.

jasonquigleyhpecom (dismissed):
## 📋 Review: `hpe_morpheus_network_router_firewall_rule_group`

**PR:** [#1285 — Adding network_router_firewall_rule_group resource](https://github.com/HPE/terraform-provider-hpe/pull/1285)

**Scope:** 🎯 PR diff vs `origin/dev-1.6` — 31 files · agents: patterns, spec-coverage, structure · UI val: config-check · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 1 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 1 |

### 📁 1. File Completeness

🔴 **Gap**: Missing `networkrouterfirewallrulegroup` singular data source. The codebase standard requires every resource to have a corresponding singular data source with full test coverage, examples, and documentation. While the plural/list data source (`networkrouterfirewallrulegroups`) is present, the singular one is absent.

### 🔍 2. Attribute Coverage

🔵 **Recommendation**: The API contract for create specifies `externalType` as a required field (with values like `"SecurityPolicy"` or `"GatewayPolicy"`), but it is intentionally explicitly removed from the Terraform schema and hardcoded to `"GatewayPolicy"`. Since this is a deliberate design decision to reserve `"SecurityPolicy"` for a separate distributed firewall resource (and matches the `hpegl` provider's design), this is acceptable. Consider adding a brief note in the resource description mentioning that this resource strictly configures gateway policies, to clarify the behaviour for users.

---

**Next:** Fix 🔴 Gaps before merge. 🔵 Recommendations are advisory. See inline comments for specifics.

morpheus/resources.go
RDavisHPE: Datasource is out of scope for this PR

specs/resources/network_router_firewall_rule_group/config.yaml
RDavisHPE: Further update: acceptance tests revealed that description has the same issue as priority — the API stores and returns "" (empty string) rather than null when not set, so optional caused an inconsistent-result error on first update (plan=null, state=""). description has also been reverted to computed_optional with state_for_unknown: true, matching the priority pattern.

morpheus/framework/resources/networkrouterfirewallrulegroup/update.go
RDavisHPE: Further update: the description guard has been reverted to !IsNull() && !IsUnknown() for the same reason as above — sending "" when the plan is null caused a state inconsistency because the API treats "" as a real stored value. With UseStateForUnknown now on description, the plan during an update is always the prior state value (never null unless it was truly never set), so the !IsNull() gate correctly prevents sending "" while still allowing explicit description = "" to clear it.

morpheus/framework/resources/networkrouterfirewallrulegroup/create.go
RDavisHPE: Updated description:
```
Manages an NSX-T gateway firewall policy group in Morpheus. Gateway firewall\npolicies are attached to a specific Tier-0 or Tier-1 router and control\nnorth-south traffic traversing that gateway. The external type is fixed as\nGatewayPolicy; for east-west distributed firewall policies, use\nhpe_morpheus_network_firewall_rule_group instead.
```

matthew-c-hpe (approved):

PR #1294, opened 2026-07-16
MORPH-13276 - Fix os_type_image inconsistent os_type_id after apply

matthew-c-hpe:
## Summary

Fixes a "Provider produced inconsistent result after apply" error on `hpe_morpheus_os_type_image`, where `os_type_id` (a required attribute) collapsed from its configured value to `null` after apply.

## Root cause

The read logic derived `os_type_id` from the associated virtual image's own `osType.id`. That field is optional (nullable) and independent of the os_type_image association — the API's `GET` os type image endpoint does not return the OS type at all. When the virtual image had no OS type set (e.g. a plain qcow2 image), the derived value was `null`, diverging from the configured value and failing the apply.

## Fix

- Preserve the already-known `os_type_id` from plan (create) / prior state (read) instead of re-deriving it from the virtual image.
- The virtual-image-derived fallback now runs only during import, where the value is otherwise unknown.
- Document the import limitation: on import `os_type_id` cannot be reliably recovered from the API and should be verified/set explicitly.

## Testing

- `go build`, `go vet`, and `golangci-lint` pass clean.
- Existing acceptance tests: `TestAccMorpheusOsTypeImageResourceExampleOk`, `TestAccMorpheusOsTypeImageResourceRequiredAttrsOk`.
- Docs regenerated via `make docs`.

## Jira

- https://hpe.atlassian.net/browse/MORPH-13276

### DIFF
--- a/morpheus/framework/resources/ostypeimage/resource.go
+++ b/morpheus/framework/resources/ostypeimage/resource.go
@@ -87,10 +87,6 @@ func getOsTypeImageAsState(
 		state.ProvisionTypeId = types.Int64Value(*provisionType)
 	}
 
-	if img.VirtualImageId == nil {
-		return state, diags
-	}
-
 	// os_type_id is not returned by the GET osTypeImage endpoint. It is a
 	// Required, RequiresReplace attribute, so when the value is already known
 	// (create/read) we preserve it rather than deriving it from the virtual
@@ -99,6 +95,11 @@ func getOsTypeImageAsState(
 	if !knownOsTypeId.IsNull() && !knownOsTypeId.IsUnknown() {
 		state.OsTypeId = knownOsTypeId
 
+		return state, diags
+	}
+
+	// The virtual-image fallback (import only) dereferences VirtualImageId.
+	if img.VirtualImageId == nil {
 		return state, diags
 	}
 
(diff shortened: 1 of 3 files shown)


morpheus/framework/resources/ostypeimage/resource.go
```diff
@@ -87,6 +91,17 @@ func getOsTypeImageAsState(
 		return state, diags
 	}
 
+	// os_type_id is not returned by the GET osTypeImage endpoint. It is a
+	// Required, RequiresReplace attribute, so when the value is already known
+	// (create/read) we preserve it rather than deriving it from the virtual
+	// image's osType, which is optional (nullable) and independent of the
+	// association. The virtual-image fallback below only runs during import.
```
RDavisHPE, line 94: Does the fallback actually only run during import? I would've expected a name == null check before this snippet  if it did. 
```
	if vi.OsType != nil && vi.OsType.Id != nil {
		state.OsTypeId = types.Int64Value(*vi.OsType.Id)
	}
```

RDavisHPE (commented):
---

## 📋 Review: `hpe_morpheus_os_type_image`

**Scope:** 🎯 PR diff vs `origin/dev-1.6` — 3 files · agents: patterns, structure · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

morpheus/framework/resources/ostypeimage/resource.go
RDavisHPE: Oh right - so since its not known on import this works too - all good!

RDavisHPE (dismissed):

jasonquigleyhpecom (approved):

PR #1295, opened 2026-07-16
MORPH-14024: Make task remote_target_password write-only

matthew-c-hpe:
## Summary
Fixes perpetual `remote_target_password` drift on the PowerShell and Shell script task resources. Morpheus returns the password as a hash, so the plaintext in config never matched the hash written to state, producing a non-empty plan on every run.

## Changes
- Deprecate `remote_target_password` (now `Sensitive`; Read no longer writes the hash back into state, which removes the drift).
- Add write-only `remote_target_password_wo` plus a `remote_target_password_wo_version` trigger, mutually exclusive with the old field via `ConflictsWith`. Non-breaking: existing configs keep working.
- Update Create/Update to source the secret from `GetRawConfig()` and re-send it when the version is bumped.
- Add a migration note to both resource docs and list both resources under WriteOnly (requires Terraform 1.11+).

## Testing
- `make lint` clean; `go build` and `go vet` clean; `make docs` regenerated and idempotent.
- `InternalValidate` verified for both resources (first SDKv2 WriteOnly field in the repo).

## Jira

- https://hpe.atlassian.net/browse/MORPH-14024

### DIFF
--- a/docs/resources/morpheus_task_powershell_script.md
+++ b/docs/resources/morpheus_task_powershell_script.md
@@ -83,7 +83,7 @@ resource "hpe_morpheus_task_powershell_script" "tfexample_powershell_git" {
 - `execute_target` (String) The execute target for the powershell script (local, remote or resource)
 - `labels` (Set of String) The organization labels associated with the task (Only supported on Morpheus 5.5.3 or higher)
 - `remote_target_host` (String) The hostname or ip address of the remote target
-- `remote_target_password` (String) The password of the user account used to authenticate to the remote target
+- `remote_target_password` (String, Sensitive) The password of the user account used to authenticate to the remote target
 - `remote_target_port` (String) The port used to connect to the remote target
 - `remote_target_username` (String) The username of the user account used to authenticate to the remote target
 - `repository_id` (Number) The ID of the git repository integration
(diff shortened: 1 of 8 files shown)


docs/resources/morpheus_task_powershell_script.md
```diff
@@ -77,13 +77,17 @@ resource "hpe_morpheus_task_powershell_script" "tfexample_powershell_git" {
 
 ### Optional
 
+> **NOTE**: [Write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments) are supported in Terraform 1.11 and later.
```
jasonquigleyhpecom, line 80: Maybe you should mention the OpenTofu version too

morpheus/sdkv2/resources/task/resource_task_powershell_script.go
```diff
@@ -135,16 +133,26 @@ func ResourceTaskPowerShellScript() *schema.Resource {
...
 			"remote_target_password": {
-				Type:        schema.TypeString,
-				Description: "The password of the user account used to authenticate to the remote target",
+				Type:          schema.TypeString,
+				Description:   "The password of the user account used to authenticate to the remote target",
+				Optional:      true,
+				Computed:      true,
+				Sensitive:     true,
+				Deprecated:    "Use remote_target_password_wo instead. This attribute stores the password hash in state.",
+				ConflictsWith: []string{"remote_target_password_wo"},
```
jasonquigleyhpecom, line 142: Should the diff suppression function still be here?

matthew-c-hpe: The `_wo` attribute never stores to state so a DiffSuppressFunc would be irrelevant here.

docs/resources/morpheus_task_powershell_script.md
matthew-c-hpe: Done, it's also 1.11 for OpenTofu.

jasonquigleyhpecom (approved):

PR #1302, opened 2026-07-23
Add TF_ACC_SWEEP_PREFIX env var to override the sweeper prefix

jasonquigleyhpecom:
## Summary

Two related changes to the acceptance-test sweeper:

1. An optional `TF_ACC_SWEEP_PREFIX` environment variable to override the hardcoded `TestAccMorpheus` resource-name prefix for targeted cleanup (defaulting to `TestAccMorpheus` when unset, so normal sweeps are unchanged).
2. A fix for the Search-based sweepers, which crashed while decoding the global Search API response, plus a matcher fix so the load balancer orphaned-instance sweep honors the configured prefix.

## Sweep prefix override

- `TestResourcePrefix` is now resolved from `TF_ACC_SWEEP_PREFIX` at startup, falling back to the default `TestAccMorpheus` when the variable is unset or blank. All existing sweepers consume it unchanged.
- The value is trimmed and guarded against empty input — an empty prefix would make `strings.HasPrefix` match every resource and sweep the whole appliance.
- When an override is active it is logged, so operators can confirm exactly which prefix a (destructive) sweep will target.
- `make sweep` gains a `SWEEP_PREFIX` passthrough: `make sweep SWEEP_SYSTEMS=<system> SWEEP_PREFIX=<prefix>`.
- Added a unit test for the resolver.

This overrides only what the sweeper matches and deletes; it does not change how acceptance tests name resources (those follow the Go test function name). The default remains `TestAccMorpheus`, so existing `make sweep` behaviour is unchanged.

## Search-based sweeper fix

The global Search API (used by the load balancer orphaned-instance and storage volume sweepers) returns `dateCreated` timestamps with a non-RFC3339 offset (for example `2026-05-12T14:36:16+0000`) that the SDK cannot decode, so those sweepers errored out before matching anything.

- Added a shared `testsweep.SearchHits` helper that reuses the SDK Search request (auth, URL, paging) but reads the raw response body and decodes only `id`/`name`/`type` into a minimal local struct — `dateCreated` is never parsed, so the problem is sidestepped without any SDK change. The helper pages through all results.
- The load balancer and storage volume sweepers now use the helper.
- Replaced the load balancer orphaned-instance matcher (a rigid `^TestAccMorpheusL\d{16}$` regex that ignored the prefix) with a prefix match on `Instance`-type hits, so it honors `TF_ACC_SWEEP_PREFIX` and matches real container instance names.
- Added a decode regression test built from a captured Search response containing the offending timestamps.

## Testing

- `go build ./...` and `go build -tags sweep ./...`
- `go test ./morpheus/testhelpers/sweep/...` (prefix resolver + Search decode tests)
- `make lint` — clean (including `--build-tags sweep` on the sweepers)
- Live, non-destructive sweep against a test appliance using a probe prefix that matches nothing: the override log fired and zero resources were swept.


RDavisHPE (commented):
## 📋 Review: Sweeper Prefix & Search Fix

**PR:** [#1302 — Add TF_ACC_SWEEP_PREFIX env var to override the sweeper prefix](https://github.com/HPE/terraform-provider-hpe/pull/1302)

**Scope:** 🎯 PR diff vs `origin/dev-1.7` — 8 files · agents: review-patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 1 |
| 🔵 Recommendations | 1 |

**Verdict:** 🟡 **PASS WITH WARNINGS** — 2 issues found (0 critical)

### 🟡 Warnings

### 🔵 Recommendations

- 🔵 **Recommendation**: In `morpheus/framework/resources/storagevolume/sweep/sweep.go`, the delete closure silently masks `strconv.ParseInt` failures on the item ID by returning `&http.Response{StatusCode: http.StatusOK}` without logging anything. Consider bringing it in line with the load balancer sweeper which explicitly logs an `[ERROR]` when an ID fails to parse.

---

**Next:** 🟡 Warnings are advisory.

morpheus/testhelpers/sweep/search.go
```diff
@@ -0,0 +1,103 @@
...
+		// decode fails on this endpoint. A missing response or non-200 status is
+		// still surfaced below, with hresp preserved for list-status handling.
+		_, hresp, err := client.SearchAPI.Search(ctx).
+			Phrase(phrase).
+			Max(searchPageSize).
+			Offset(offset).
+			Execute()
+		if hresp == nil || hresp.StatusCode != http.StatusOK {
+			return nil, hresp, fmt.Errorf(
+				"failed to search for %q: %s", phrase, errfmt.ErrMsg(err, hresp),
```
RDavisHPE: 🟡 **Warning:** `SearchHits` pre-formats the search error with `errfmt.ErrMsg(err, hresp)`. When used by a typed sweeper like `storagevolume` (which returns the error to the framework), the caller `RegisterTypedAPISweeper` also calls `errfmt.ErrMsg(err, hresp)` on the returned error. Because `errfmt.ErrMsg` reads and exhausts `hresp.Body`, the first call consumes it. The second call then appends the status text again but with an empty body, resulting in redundant status code appending and a trailing empty body string in the final log (e.g., `... (Forbidden): {"error": "..."} (Forbidden): `). Consider just wrapping the error (`fmt.Errorf("...: %w", err)`) and letting the caller format it.

morpheus/testhelpers/sweep/search_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// TestParseSearchPage feeds a real global Search API response (captured from an
+// appliance) through the sweeper's decoder. The payload contains dateCreated
+// values with a non-RFC3339 offset ("...+0000") that break the SDK's typed
+// decode; parseSearchPage must ignore them and still extract id/name/type.
+func TestParseSearchPage(t *testing.T) {
```
RDavisHPE, line 15: 🔵 **Recommendation:** Consider renaming the test functions to start with `TestUnit` (e.g., `TestUnitParseSearchPage`) to explicitly distinguish them from acceptance (`TestAcc`) and sweep (`TestSweep`) targets, aligning with standard provider conventions.

RDavisHPE: I think Test works fine here

morpheus/testhelpers/sweep/search.go
jasonquigleyhpecom: Good catch on the body exhaustion. Fixed in 9d0242df: `SearchHits` now wraps the error with `%w` and returns `hresp`, so it is formatted exactly once — by the framework for the storage-volume sweeper and by `findOrphanedInstances` for the LB sweeper. Also added the `[ERROR]` log on storage-volume ID parse failure to match the LB sweeper, and left the `Test*` names as-is per your follow-up.

RDavisHPE (approved):

PR #1305, opened 2026-07-24
Fix flaky image datasource tests caused by concurrent image creation

jasonquigleyhpecom:
## What

Consolidate the three `hpe_morpheus_image` datasource acceptance tests — lookup **by id**, **by name**, and **by name + image_type** — into a single `TestAccMorpheusImageDatasource` that creates one image and reads it back all three ways in one apply.

## Why

The three tests previously each created their own image resource and ran in parallel (`t.Parallel()`). Under that concurrency the image-create `POST` intermittently returned a transient `403 Forbidden` ("You do not have permissions to access this api endpoint"), even though the test account has full permissions. In the last nightly run this failed **2 of the 3** tests — the third, identical, test passed — with all 403s clustered in a single minute, pointing to concurrent-create contention rather than a real authorization problem.

Consolidating cuts image-create calls from **three to one**, removing the contention and the associated flakiness while keeping identical datasource coverage.

## Changes

- Merge `…ById`, `…ByName`, and `…ByNameAndImageType` into one `TestAccMorpheusImageDatasource` (one image; three data-source lookups; six attribute checks).
- Leave the two negative unit tests and the `//go:generate` example directives untouched.

## Testing

- `go build ./...`, `gofmt`, `golangci-lint` (0 issues), `make docs` (clean tree).
- Ran the consolidated test against a live backend — **passes**. Also re-ran the other previously-failing acceptance tests (form, job, network, option-list): **22/22 pass**.

Test-only change; no provider runtime code is affected.

### DIFF
--- a/morpheus/framework/datasources/image/image_test.go
+++ b/morpheus/framework/datasources/image/image_test.go
@@ -37,139 +37,13 @@ func TestMain(m *testing.M) {
 	os.Exit(code)
 }
 
-func TestAccMorpheusImageDatasourceById(t *testing.T) {
-	defer testhelpers.RecordResult(t)
-
-	capabilities.MustHaveOrSkip(t, capabilities.All)
-
-	t.Parallel()
-
-	if testing.Short() {
-		t.Skip("Skipping slow test in short mode")
-	}
-
-	name := acctest.RandomWithPrefix(t.Name())
-
-	providerConfig := testhelpers.ProviderBlock()
-
-	imageConfig, err := image.RenderImageConfig(t, map[string]string{
-		"Name":              name,
-		"OsTypeId":          "data.hpe_morpheus_os_type.test.id",
-		"StorageProviderId": "data.hpe_morpheus_storage_bucket.test.id",
-	})
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	dependencyConfig := `
-data "hpe_morpheus_os_type" "test" {
-	name = "linux"
-}
-
-data "hpe_morpheus_storage_bucket" "test" {
-	name = "Local Storage"
-}
-` + imageConfig
-
-	dataSourceConfig, err := testhelpers.RenderExample(t,
-		"example-id.tf.tmpl", "Id", "hpe_morpheus_image.example_image.id")
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	checks := []resource.TestCheckFunc{
-		resource.TestCheckResourceAttr(
-			"data.hpe_morpheus_image.test",
-			"name",
-			name,
-		),
-		resource.TestCheckResourceAttr(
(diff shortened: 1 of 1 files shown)


morpheus/framework/datasources/image/image_test.go
```diff
@@ -137,92 +78,31 @@ data "hpe_morpheus_storage_bucket" "test" {
...
-			{
-				Config: providerConfig + dependencyConfig + dataSourceConfig,
-				Check:  checkFn,
-			},
-		},
-	})
+	// The example-*.tf.tmpl templates hardcode the "test" datasource label, so
+	// they cannot be combined in a single config. Inline the three lookups here
+	// so they all reference the one image created above.
+	dataSourceConfig := `
```
jasonquigleyhpecom, line 84: 🟠 **Violation**: By inlining the `dataSourceConfig` lookups instead of using `testhelpers.RenderExample(...)`, the acceptance test no longer validates the actual user-facing examples (`example-*.tf.tmpl`). If those templates fall out of sync with the schema, the tests will pass but the generated docs will be broken. This breaks the testing pattern.

*Recommendation:* You can avoid concurrent-creation 403s while still validating the example templates by splitting the lookups across sequential `resource.TestStep`s in the same `TestCase`. Each step should include the `dependencyConfig` (which provisions the image once in the first step) and one rendered example template.

jasonquigleyhpecom (commented):
---

## 📋 Review: `hpe_morpheus_image`

**PR:** [#1305 — Consolidate image datasource acceptance tests](https://github.com/HPE/terraform-provider-hpe/pull/1305)

**Scope:** 🎯 PR diff vs `origin/dev-1.7` — 1 files · agents: review-patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 1 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

**Verdict:** 🔴 **FAIL** — 1 issues found (1 critical)

---

### ⚙️ 4. Pattern Compliance

- 🟠 **Violation** (`morpheus/framework/datasources/image/image_test.go:84`): By inlining the `dataSourceConfig` lookups instead of using `testhelpers.RenderExample(...)`, the acceptance test no longer validates the actual user-facing examples (`example-*.tf.tmpl`). If those templates fall out of sync with the schema, the tests will pass but the generated docs will be broken. This breaks the testing pattern.

  *Recommendation:* You can avoid concurrent-creation 403s while still validating the example templates by splitting the lookups across sequential `resource.TestStep`s in the same `TestCase`. Each step should include the `dependencyConfig` (which provisions the image once in the first step) and one rendered example template.

---

matthew-c-hpe (approved):

PR #1306, opened 2026-07-27
Redact credentials from API traces

jasonquigleyhpecom:
API tracing (`MORPHEUS_API_HTTPTRACE`) dumps full HTTP requests and
responses **including bodies** via `httputil.DumpRequestOut(req, true)`, with
no redaction. Those traces are captured in `go test -json` output and kept as
build artifacts, so credentials were being written to logs in clear text:

- the `Authorization: Bearer …` header on every API call, and
- the appliance password, when the provider authenticates with
  username/password and POSTs the credentials to `/oauth/token`, plus the
  `access_token` / `refresh_token` in the response.

## Changes

- Add `Redact()`, applied to both the request and response dumps. It scrubs
  `Authorization` / `Proxy-Authorization` / `Cookie` / `Set-Cookie` headers,
  `password` / `client_secret` / `access_token` / `refresh_token` in
  form-encoded bodies, and the same keys in JSON bodies, replacing each value
  with `[REDACTED]`.
- Add `MORPHEUS_API_HTTPTRACE_SHOW_SECRETS` to disable redaction when
  debugging an authentication problem locally.
- Emit a one-time warning when redaction is disabled.
- Promote the env var names to exported constants (`EnvEnabled`,
  `EnvShowSecrets`).

Non-secret parts of the trace are deliberately preserved (request line,
`username=`, other headers) so traces remain useful for debugging.

### Opt-out semantics

Unlike `IsEnabled()`, which only tests for presence, `ShowSecrets()` requires
an affirmative value. This is a security control, so an accidentally blank
export must not silently expose credentials.

| Value | Redaction |
|---|---|
| unset, `""`, `0`, `false`, `no`, `off` | **on** (default) |
| `1`, `true`, `yes`, any other value | off |

## Testing

- New unit tests: redaction vectors (OAuth password grant, `Authorization`
  header, token response, JSON password, `Set-Cookie`), a 12-case
  `ShowSecrets` matrix, the opt-out path, and `IsEnabled`.
- Verified end-to-end against a live appliance by running an acceptance test
  with tracing enabled: 10 request/response traces captured, `Authorization:
  [REDACTED]`, and the real token appeared **nowhere** in the log. Re-running
  with `MORPHEUS_API_HTTPTRACE_SHOW_SECRETS=true` restored the raw value and
  logged the warning exactly once.
- `make lint` clean (0 issues); `make unit-tests` passes.

### DIFF
--- a/utils/httptrace/httptrace.go
+++ b/utils/httptrace/httptrace.go
@@ -7,16 +7,104 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"os"
+	"regexp"
+	"strings"
+	"sync"
 
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 )
 
 var _ http.RoundTripper = TraceRoundTripper{}
 
+const (
+	// EnvEnabled turns API request/response tracing on when set.
+	EnvEnabled = "MORPHEUS_API_HTTPTRACE"
+
+	// EnvShowSecrets disables redaction of credentials in those traces.
+	// #nosec G101 -- environment variable name, not an embedded credential.
+	EnvShowSecrets = "MORPHEUS_API_HTTPTRACE_SHOW_SECRETS"
+)
+
+// redactedValue replaces any secret found in a traced request or response.
+const redactedValue = "[REDACTED]"
+
+var (
+	// Credential-bearing headers. Matched per line against a dumped
+	// request/response, which uses CRLF separators.
+	secretHeaderRE = regexp.MustCompile(
+		`(?im)^(Authorization|Proxy-Authorization|Cookie|Set-Cookie):[^\r\n]*`,
+	)
+
+	// Secrets in form-encoded bodies, e.g. the OAuth password grant
+	// "grant_type=password&username=u&password=p" sent to /oauth/token.
+	secretFormRE = regexp.MustCompile(
+		`(?i)\b(password|client_secret|access_token|refresh_token)=[^&\r\n]*`,
+	)
+
+	// Secrets in JSON bodies, e.g. the token response
+	// {"access_token":"...","refresh_token":"..."}.
+	secretJSONRE = regexp.MustCompile(
+		`(?i)"(password|clientSecret|client_secret|accessToken|access_token` +
+			`|refreshToken|refresh_token)"(\s*:\s*)"[^"]*"`,
+	)
+
+	// showSecretsWarning ensures the clear-text warning is emitted once per
+	// process rather than on every request.
+	showSecretsWarning sync.Once
+)
+
+// IsEnabled reports whether API tracing is turned on.
 func IsEnabled() bool {
-	_, enabled := os.LookupEnv("MORPHEUS_API_HTTPTRACE")
+	_, enabled := os.LookupEnv(EnvEnabled)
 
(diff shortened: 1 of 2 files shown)


utils/httptrace/httptrace_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+// secret values that must never survive Redact.
+const (
+	secretPassword = "sup3r-s3cret-pw"
+	secretToken    = "abcdef0123456789"
+)
+
+// These tests mutate the environment, so they must not run in parallel.
+
+func TestRedact(t *testing.T) {
```
jasonquigleyhpecom, line 20: 🟡 **Warning** — Unit tests should be prefixed with `TestUnit` (e.g., `TestUnitRedact`) to clearly distinguish them from `TestAcc` acceptance tests.

utils/httptrace/httptrace.go
```diff
@@ -7,18 +7,106 @@ import (
...
+
+	// Secrets in form-encoded bodies, e.g. the OAuth password grant
+	// "grant_type=password&username=u&password=p" sent to /oauth/token.
+	secretFormRE = regexp.MustCompile(
+		`(?i)\b(password|client_secret|access_token|refresh_token)=[^&\r\n]*`,
+	)
+
+	// Secrets in JSON bodies, e.g. the token response
+	// {"access_token":"...","refresh_token":"..."}.
+	secretJSONRE = regexp.MustCompile(
```
jasonquigleyhpecom, line 46: 🔵 **Recommendation** — The `[^"]*` expression will stop matching early if a password contains an escaped quote (e.g., `{"password": "foo\"bar"}`). This leaves trailing string fragments in the trace, making it malformed JSON (`"password":"[REDACTED]bar"`). Since this is only for logging, it's not critical, but `(?:[^"\\]|\\.)*` would handle escaped quotes correctly.

jasonquigleyhpecom (commented):
---

## 📋 Review: `httptrace`

**PR:** [#1306 — Redact credentials from API traces](https://github.com/HPE/terraform-provider-hpe/pull/1306)

**Scope:** 🎯 PR diff vs `origin/dev-1.7` — 2 files · agents: patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 1 |
| 🔵 Recommendations | 1 |

**Verdict:** 🟢 **PASS** — 2 issues found (0 critical)

---

### ⚙️ 4. Pattern Compliance

- 🟡 **Test Naming Convention**: Unit tests (`TestRedact`, `TestShowSecrets`, `TestRedactDisabled`, `TestIsEnabled`) do not follow the provider's unit test naming convention. They should be prefixed with `TestUnit` (e.g., `TestUnitRedact`) to clearly distinguish them from `TestAcc` acceptance tests. (See inline comment)

### 💡 6. Code Quality (Recommendations)

- 🔵 **Regex Escaping**: The `secretJSONRE` expression uses `[^"]*` which could malform the JSON trace if a secret contains an escaped quote. (See inline comment)

---

matthew-c-hpe (approved):

PR #1307, opened 2026-07-27
Fix load balancer sweeper leaks and post-apply contract violations

jasonquigleyhpecom:
Fixes two independent classes of load balancer test failure: sweepers that could
not keep up with leaked NSX-T resources, and resources that returned state
inconsistent with the plan.

---

# 1. Sweepers leaked NSX-T load balancer services

Acceptance tests leaked NSX-T load balancer services until the edge node ran out
of capacity. In one run, 10 of 27 `POST /api/load-balancers` attempts failed
before the suite had created anything — the edge node was already full when the
run started:

```
Errors validating path=[/infra/lb-services/1b5a6427-…].
There is no available capacity on edge node …/edge-nodes/1 to deploy a SMALL
load balancer service … the existed load balancer services on the edge node are
/infra/lb-services/d633f22b-…(SMALL), /infra/lb-services/63754663-…(SMALL), … (10+)
```

The `hpe_morpheus_load_balancer` sweeper already existed and was correctly
registered — it just could not keep up, for four independent reasons.

### 1.1 Lists were capped at 25 records

`ListLoadBalancers(ctx).Execute()` was called with no `Max`. When the `max` query
parameter is omitted the API applies a default of **25**, so the sweeper only ever
saw the first 25 load balancers (and, in the child sweepers, the first 25 children
of each). Anything older was invisible and could never be reclaimed, so the leak
only ever grew. Confirmed against a live appliance:

```json
{ "meta": { "offset": 0, "max": 25, "size": 1, "total": 1 } }
```

Fixed across all 9 list calls in the 5 `loadbalancer*` sweepers via a new shared
`testsweep.ListPageSize` constant, rather than scattering a magic number.

### 1.2 Load balancers were deleted before their children

`WithDependencies` listed only `monitor` and `virtual_server`, omitting `pool` and
`profile`. NSX-T refuses to delete an lb-service that still has children attached.
Both are now declared.

### 1.3 Tier-1 gateways were stranded

Each NSX-T load balancer test creates a per-test tier-1 gateway. The
`network_router` sweeper had no ordering relative to load balancers, so NSX-T
rejected the delete and both objects were left behind:

```
network router 2502 DELETE failed: 400 (Bad Request): {"msg":"The object
path=[/infra/tier-1s/50cd2ad5-…] cannot be deleted as either it has children or
it is being referenced by other objects path=[/infra/lb-services/35f2e021-…]"}
```

`hpe_morpheus_network_router` now depends on `hpe_morpheus_load_balancer`.

### 1.4 One failing sweeper aborted the whole run

`make sweep` did not pass `-sweep-allow-failures`. Sweepers execute in map
iteration order, so a single failure aborted the run and left every
not-yet-executed sweeper's resources on the appliance.

---

# 2. Post-apply contract violations

Three resources returned state that did not match the plan, so applies failed with
`Provider produced inconsistent result after apply` or `Provider returned invalid
result object after apply`. This accounted for 9 failing acceptance tests.

### 2.1 `load_balancer_monitor` — configured values nulled out

```
.receive_data: was cty.StringVal(""), but now null.
.data_length:  was cty.NumberIntVal(0), but now null.
```

Both attributes are sent on create, but the API does not echo them back: an empty
`receive_data` reads back as `null`, and `data_length` is only round-tripped for
ICMP monitors (these tests use an HTTP monitor). The response mapper therefore
turned a configured value into `null`.

Added `preserveUnreturnedFields`, called from create, update and read alongside
the existing `monitor_type` fixup. Unknown prior values are deliberately left
alone so they resolve from the API response rather than leaking into state.

### 2.2 `load_balancer_profile` — unknown values leaked into state

```
After the apply operation, the provider still indicated an unknown value for
  …config_client_ssl.supported_ssl_ciphers / .supported_ssl_protocols
  …config_cookie_persistence.cookie_path / .max_cookie_age / .max_idle_time
  …config_http.redirect_address / .request_body_size / .x_forwarded_for
```

The `config_*` block was copied verbatim from the plan. Every nested
`Optional+Computed` attribute the configuration omits is **unknown** in the plan,
and those unknowns were written straight into state.

Rebuilding the block from the API response was not an option either — that
overwrites configured values with the API's own defaults (for example
`x_forwarded_for` defaults to `INSERT` server-side), which is the very reason the
previous code copied the plan wholesale.

The fix merges per attribute: a configured (known) value always wins, and only
unknown attributes are resolved from the API response, falling back to null.
This satisfies both constraints, and computed attributes now correctly reflect the
server-side value. Applied to all eight config variants, not just the three that
were failing.

### 2.3 `network_router_bgp_neighbor` — configured description nulled out

```
.description: was cty.StringVal("TestAcc…"), but now null.
```

The API accepts `description` on create and update but never returns it, so the
mapper unconditionally nulled it. Now preserved from plan/prior state, guarded on
`IsUnknown` so an unset `Optional+Computed` description cannot leak an unknown
into state.

---

# Testing

The target appliance is not reachable from the development environment, so these
changes are covered by unit tests and will be exercised against a live appliance
separately.

- **New unit tests** (5 groups, 16 cases) covering the preserve/merge semantics:
  configured values win, unknowns resolve from the API response, absent values
  fall back to null, null blocks stay null, and no unknown ever survives a merge.
- **The tests were verified to actually catch the bug.** Temporarily reverting the
  merge to the previous blanket copy makes `TestMergeConfigBlocksLeavesNoUnknowns`
  fail with:
  ```
  config_http.request_body_size is still unknown after merge
  config_http.redirect_address is still unknown after merge
  config_http.x_forwarded_for is still unknown after merge
  ```
  — precisely three of the eight attributes reported by the failing acceptance
  tests.
- Sweeper ordering was verified against a live appliance: `profile`, `pool`,
  `monitor` and `virtual_server` all now execute **before**
  `hpe_morpheus_load_balancer`.
- `go build`, `go vet` (with and without the `sweep` build tag), `make lint`
  (0 issues) and `make unit-tests` all pass.

# Notes for reviewers

- **`GetNetworkRouters` cannot be paginated.** The generated SDK exposes no
  `Max`/`Offset` builders for that endpoint, so the router sweeper is still limited
  to the API default page size. An explicit `NOTE` comment marks this rather than
  leaving it silently broken; lifting it needs the query parameters declared in the
  API spec first.
- **All 57 sweepers share the same 25-record cap.** Only the load balancer and
  router sweepers are fixed here to keep this reviewable. The rest is a worthwhile
  follow-up.
- The sweeper change is **preventive**: the appliance had already been cleaned up
  by the time the fix was exercised, so it stops the leak recurring rather than
  performing the reclaim itself.
- No schema or codegen changes — the contract fixes are all in state mapping.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,6 +65,9 @@ docs: build-render-tool
 	go generate ./...
 	cd tools && go generate
 
+# -sweep-allow-failures keeps the run going when an individual sweeper fails.
+# Without it the first failure aborts the whole sweep in map-iteration order,
+# so unrelated resources are left on the appliance and the leak compounds.
 sweep:
 	env TF_ACC_SWEEP_PREFIX=$(SWEEP_PREFIX) \
-	go test -v -tags sweep ./morpheus/testhelpers/sweep/... -sweep=$(SWEEP_SYSTEMS) $(SWEEP_RUN_ARGS)
+	go test -v -tags sweep ./morpheus/testhelpers/sweep/... -sweep=$(SWEEP_SYSTEMS) -sweep-allow-failures $(SWEEP_RUN_ARGS)
--- a/morpheus/framework/resources/loadbalancer/sweep/sweep.go
+++ b/morpheus/framework/resources/loadbalancer/sweep/sweep.go
@@ -36,7 +36,8 @@ func init() {
 			*http.Response,
 			error,
 		) {
-			resp, hresp, err := client.LoadBalancersAPI.ListLoadBalancers(ctx).Execute()
+			resp, hresp, err := client.LoadBalancersAPI.ListLoadBalancers(ctx).
+				Max(testsweep.ListPageSize).Execute()
 			if resp == nil {
 				return nil, hresp, err
 			}
@@ -67,9 +68,14 @@ func init() {
 
 			return hresp, err
 		},
+		// Children must be removed before the load balancer itself: NSX-T
+		// refuses to delete an lb-service that still has pools or profiles
+		// attached.
 		testsweep.WithDependencies[sdk.ListLoadBalancers200ResponseAllOfLoadBalancersInner](
 			"hpe_morpheus_load_balancer_monitor",
 			"hpe_morpheus_load_balancer_virtual_server",
+			"hpe_morpheus_load_balancer_pool",
+			"hpe_morpheus_load_balancer_profile",
 		),
 	)
 
--- a/morpheus/framework/resources/loadbalancer/sweep_import.go
+++ b/morpheus/framework/resources/loadbalancer/sweep_import.go
@@ -7,5 +7,7 @@ package loadbalancer
 import (
 	_ "github.com/HPE/terraform-provider-hpe/morpheus/framework/resources/loadbalancer/sweep"
 	_ "github.com/HPE/terraform-provider-hpe/morpheus/framework/resources/loadbalancermonitor/sweep"
+	_ "github.com/HPE/terraform-provider-hpe/morpheus/framework/resources/loadbalancerpool/sweep"
+	_ "github.com/HPE/terraform-provider-hpe/morpheus/framework/resources/loadbalancerprofile/sweep"
 	_ "github.com/HPE/terraform-provider-hpe/morpheus/framework/resources/loadbalancervirtualserver/sweep"
 )
--- a/morpheus/framework/resources/loadbalancermonitor/resource_create.go
+++ b/morpheus/framework/resources/loadbalancermonitor/resource_create.go
@@ -271,5 +271,8 @@ func (r *Resource) Create(
 		state.Config = plan.Config
 	}
 
+	// Restore fields the API accepts but does not return.
+	preserveUnreturnedFields(&state, plan)
+
 	resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
 }
--- a/morpheus/framework/resources/loadbalancermonitor/resource_read.go
+++ b/morpheus/framework/resources/loadbalancermonitor/resource_read.go
@@ -86,6 +86,9 @@ func (r *Resource) Read(
 		state.Config = data.Config
 	}
 
+	// Restore fields the API accepts but does not return.
+	preserveUnreturnedFields(&state, data)
+
 	resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
 }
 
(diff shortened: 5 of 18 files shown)


jasonquigleyhpecom (commented):
---

## 📋 Review: `hpe_morpheus_load_balancer_*`, `hpe_morpheus_network_router_*`

**PR:** [#1307 — Fix load balancer sweeper leaks and post-apply contract violations](https://github.com/HPE/terraform-provider-hpe/pull/1307)

**Scope:** 🎯 PR diff vs `origin/dev-1.7` — 16 files · agents: patterns, structure · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

RDavisHPE (dismissed):
---

## 📋 Review: `hpe_morpheus_load_balancer` (and related sweepers)

**PR:** [#1307 — Fix load balancer sweeper leaks and post-apply contract violations](https://github.com/HPE/terraform-provider-hpe/pull/1307)

**Scope:** 🎯 PR diff vs `origin/dev-1.7` — 16 files · agents: patterns, structure · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 2 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

### 📁 1. File Completeness

🔴 **Gap: Missing sweeper blank imports in `loadbalancer/sweep_import.go`**
When `loadbalancer` declares dependencies on `loadbalancerpool` and `loadbalancerprofile` in its sweeper via `testsweep.WithDependencies(...)`, it must blank-import their `sweep` packages in `sweep_import.go`. Without this, the Go test runner will not execute their `init()` functions during the `loadbalancer` package tests, meaning those sweepers won't be registered when they are needed.
- **File:** `morpheus/framework/resources/loadbalancer/sweep_import.go` (Not touched in PR, but required by changes in `sweep.go`)
- **Missing:**
  ```go
  _ "github.com/HPE/terraform-provider-hpe/morpheus/framework/resources/loadbalancerpool/sweep"
  _ "github.com/HPE/terraform-provider-hpe/morpheus/framework/resources/loadbalancerprofile/sweep"
  ```

🔴 **Gap: Missing sweeper blank import in `networkrouter/sweep_import.go`**
Similarly, `networkrouter` added `hpe_morpheus_load_balancer` to its sweeper dependencies, so it must blank-import the load balancer sweeper to ensure it is registered when testing `networkrouter`.
- **File:** `morpheus/framework/resources/networkrouter/sweep_import.go` (Not touched in PR, but required by changes in `sweep.go`)
- **Missing:**
  ```go
  _ "github.com/HPE/terraform-provider-hpe/morpheus/framework/resources/loadbalancer/sweep"
  ```

---

**Next:** Fix 🔴 Gaps before merge.

eoghanino (approved):

PR #1311, opened 2026-07-28
MORPH-14683: Keep instance disks backed by a storage server datastore

eamonnotoole:
Fixes a regression that breaks `hpe_morpheus_instance` whenever its disks are provisioned onto a datastore that is backed by a storage server, such as an Alletra MP datastore.

## Symptom

```

When applying changes to hpe_morpheus_instance.test, provider produced an
unexpected new value: .volumes: was cty.ListVal([...]), but now null.
```

## Cause

Instance volumes were filtered out whenever they carried a `storageServer` reference. That reference does not by itself make a volume external: when a disk is provisioned onto a datastore that is backed by a storage server, the API copies the datastore's storage server onto the instance's own disk, setting the datastore and the storage server together, and clearing them together when the disk is detached.

The instance's own disks were therefore discarded, the volume list was emptied, and an empty slice converts to a **null** list rather than an empty one — so `volumes` was written to state as null while the plan held a populated list.

## Scope

The read path is shared by create, update, refresh and import, so an affected instance fails to create, and on refresh has its volumes nulled in state. A subsequent plan then sees configured volumes against a null state value.

Instances on ordinary datastores are unaffected, as those volumes carry no storage server.

## Fix

A volume is now only treated as externally attached when it comes from a storage server **and** was not provisioned through a datastore. The root volume is never excluded.

Externally attached array LUNs — created by `hpe_morpheus_storage_volume` and exported to the instance's host — are still filtered, exactly as before. That behaviour was deliberate and is preserved; the existing unit test covering it is unchanged and still passes.

Also fails loudly rather than writing null volumes into state if filtering ever removes every volume while the configuration declares some, so a similar problem surfaces as a clear error instead of an opaque inconsistent-result failure.

## Tests

- Existing `TestUnitRemoveExternalStorageVolumes` unchanged and passing, confirming externally attached LUNs are still excluded.
- New table-driven `TestUnitIsExternallyAttachedVolume` covering ordinary disks, an externally attached LUN, root and data volumes on a storage-server-backed datastore, and a root volume from a storage server with no datastore.
- New `TestUnitRemoveExternalStorageVolumesKeepsDatastoreBackedDisks` reproducing the reported case.

## Targeting

Raised against `dev-1.6` because 1.6.0 is due to ship imminently and the regression is not in any released version — the newest tag is v1.5.0. The same fix is prepared for the 1.7 line and will follow.

## Notes

An acceptance test that provisions an instance whose root volume sits on a storage-server-backed datastore would have caught this and is worth adding; there is currently no coverage for that configuration.

### DIFF
--- a/morpheus/framework/resources/instance/instance_read.go
+++ b/morpheus/framework/resources/instance/instance_read.go
@@ -1210,6 +1210,28 @@ func getVolumes(
 	// instance's host — which are not part of the instance's provisioned disks.
 	apiVolumes = removeExternalStorageVolumes(apiVolumes)
 
+	// Filtering must never remove every volume while the configuration declares
+	// some. convert.ToListType maps an empty slice to a null list, so silently
+	// continuing would write `volumes = null` into state and surface as an
+	// opaque "Provider produced inconsistent result after apply" rather than a
+	// usable error. Fail loudly instead.
+	if len(apiVolumes) == 0 && len(plan.Volumes.Elements()) > 0 {
+		diags.AddError(
+			"no instance volumes left after filtering",
+			fmt.Sprintf(
+				"instance %d returned %d volume(s), but none were recognised as "+
+					"belonging to the instance, while the configuration declares %d. "+
+					"This is a provider bug: the instance's volumes would be written "+
+					"to state as null.",
+				instanceIDValue(instance),
+				len(serverVolumes),
+				len(plan.Volumes.Elements()),
+			),
+		)
+
+		return basetypes.NewListNull(VolumesValue{}.Type(ctx)), diags
+	}
+
 	// Import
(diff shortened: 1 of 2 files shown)


jasonquigleyhpecom (commented):
---

**PR:** [#1311 — MORPH-14683: Keep instance disks backed by a storage server datastore](https://github.com/HPE/terraform-provider-hpe/pull/1311)

**Scope:** 🎯 PR diff vs `origin/dev-1.6` — 2 files · agents: patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

jasonquigleyhpecom (approved):

PR #1312, opened 2026-07-28
MORPH-14919: Stop unrelated computed attributes churning the instance plan

eamonnotoole:
Changing one attribute of an `hpe_morpheus_instance` — for example `service_plan_options.max_memory` — also showed `connection_info`, `labels`, and every computed field of `network_interfaces` and `volumes` as `(known after apply)`, even though none of them changed. A small edit produced a diff too large to review, which is a blocker for using the resource in production: you cannot safely approve an apply whose plan contains changes you did not ask for.

## Cause

The plugin framework marks every computed attribute that is null in the configuration as unknown whenever the planned state differs from the prior state (`MarkComputedNilsAsUnknown`). That is deliberate — the framework's own comment notes that later plan modifier passes are expected to put known values back — but this resource had no such pass, so everything surfaced as unknown.

The values themselves are stable: comparing two plans taken either side of a real resize apply, the interface id, interface name, `primary_interface`, volume id and `connection_info` were all identical before and after.

## Why not UseStateForUnknown

A blanket `UseStateForUnknown` would not be safe. Update re-reads the instance from the API and writes whatever it returns into state, so pinning an attribute the API legitimately changes during the update would raise `Provider produced inconsistent result after apply` — a hard failure, and a worse outcome than a noisy plan. Two cases matter:

- `connection_info` holds the instance addresses. A resize that cannot be performed hot stops and restarts the instance, so the address can change as a direct result of the update.
- Interface and volume identity can change when those collections are themselves reconfigured. Note that from 8.1.2 a network change is applied in place rather than forcing replacement, and the API may recreate interfaces.

Applying it per nested attribute would also be positional, and would pin values from the wrong element when interfaces or volumes are added, removed or reordered.

## Approach

A conditional restore in the resource-level `ModifyPlan`, which runs after the framework has marked attributes unknown and can therefore correct the plan. Two rules keep it safe:

1. **Only stable, API-assigned identity attributes are restored** — interface `id`, `name` and `primary_interface`, and volume `id`. Attributes the post-apply read sources from the API, such as volume `storage_profile` and `controller_mount_point`, are deliberately excluded.
2. **A value is only filled where the plan says unknown, and only from a prior-state value that is itself known and non-null.** A null is never pinned.

Each group is additionally gated on its own triggers:

| Group | Restored unless this changed |
|---|---|
| `network_interfaces` identity | `network_interfaces` |
| `volumes` identity | `volumes` |
| `labels` | `labels` |
| `connection_info` | `service_plan_options`, `network_interfaces`, `volumes`, `config`, `plan_id`, `layout_id` |

Deciding whether something changed ignores values the framework marked unknown, since those carry no practitioner intent: unknowns are filled from prior state and the result compared with prior state. Anything that cannot be resolved — a newly added element, an attribute missing from state — is treated as a change, so the plan is left alone rather than restoring a value that cannot be verified.

Restoring per attribute rather than per collection is safe against positional aliasing because each group stays gated on its triggers, and a path with no counterpart in prior state is left unknown.

Placed in hand-written code rather than the generated schema so it survives regeneration.

## Why not restore whole collections

An earlier iteration of this change restored each collection wholesale. That is not safe, and acceptance testing caught it.

Volume `storage_profile` is null in prior state after a create or update, because the post-apply read prefers the planned value only when that value is known and non-null, and otherwise takes the API value. Restoring the whole collection therefore placed a **known null** into the plan, which apply then contradicted:

```
  .volumes[0].storage_profile: was null, but now cty.StringVal("kvm-cache-none")
```

That is the same failure class this change exists to avoid. It was intermittent — roughly one run in three — because whether prior state holds the null or the real value depends on whether a refresh ran first.

Rule 2 above makes that failure impossible by construction rather than by exclusion list: since a null is never pinned, no restored value can contradict what the read path supplies.

## Result

- Metadata-only edits (tags, labels, name, description, instance_context) now plan cleanly.
- A memory resize drops from eleven churning lines to one: `connection_info` deliberately stays unknown, because it genuinely can change.

## Tests

**Unit tests** over the comparison and restore logic, which are pure and need no appliance:

- `TestUnitAttributeChanged` — identical values, computed-only unknowns (must not count as a change), a practitioner edit, and element added / removed / reordered.
- `TestUnitAttributeChangedMissingAttribute` — an unresolvable attribute is reported as changed, so the plan is left alone.
- `TestUnitFillUnknownsFromState` and `TestUnitFillUnknownsFromStateLeavesUnmatchedUnknown` — unknowns are taken from prior state, and an unknown with no counterpart stays unknown.
- `TestUnitMatchesRestoreRule` — asserts that `storage_profile` and `controller_mount_point` are **not** covered by the volume rule.
- `TestUnitRestoreFromStateNeverPinsNull` — the regression test for the failure above: a null in prior state is left unknown rather than pinned.
- `TestUnitRestoreFromStateLeavesUncoveredAttributesAlone` — an attribute outside the rule's fields is untouched even when prior state holds a good value.

**Acceptance tests** covering both halves of the contract — that values are restored when they should be, and deliberately left unknown when they should not be:

- `TestAccMorpheusInstanceResourceTagUpdateDoesNotChurnPlan` — a metadata-only edit restores interface and volume identity, so the plan stays reviewable.
- `TestAccMorpheusInstanceResourceResizeKeepsInterfaceAndVolumeIdentity` — the same on a service plan change. Addresses are deliberately not asserted, since a resize that cannot be performed hot restarts the instance and the address can legitimately change.
- `TestAccMorpheusInstanceResourceVolumeAddLeavesVolumeIdentityUnknown` — reconfiguring the volumes leaves volume identity unknown, so a positional restore cannot pin the previous element's values onto a different volume.

## Verification

Run against a live Morpheus appliance. Because the earlier failure was intermittent, the suite was run **five consecutive times** rather than once:

| Runs | Tests | Result |
|---|---|---|
| 5 | 15 | all pass, no `inconsistent result` errors |

Provider trace logging confirms a real lifecycle rather than a short-circuit: an instance is provisioned with an assigned address, then created, read, updated and deleted.

To confirm the acceptance tests are genuine regression guards rather than assertions that would pass either way, the plan modifier was temporarily disabled and the tag test re-run. It fails, on exactly the attributes this change addresses:

```
Step 2/2 error: Pre-apply plan check(s) failed:
    path not found: specified key id not found in map at volumes.0.id
    path not found: specified key id not found in map at network_interfaces.0.id
    path not found: specified key name not found in map at network_interfaces.0.name
```

Unknown values are omitted from the plan JSON entirely, which is why the keys are absent. So the tests pass with this change and fail without it.

Five clean runs is strong evidence rather than proof; the unit tests are the firmer guarantee, since they pin the invariant instead of sampling it.

## Notes

One gap remains: an in-place network change is not covered, because it needs a second valid network id, which is environment-specific. That is the case where the gating matters most for `network_interfaces`, since from 8.1.2 a network change is applied in place rather than forcing replacement, and the API may recreate interfaces. Worth adding once a second network is available in the test environment.

### DIFF
--- a/morpheus/framework/resources/instance/instance_plan_modifiers.go
+++ b/morpheus/framework/resources/instance/instance_plan_modifiers.go
@@ -5,17 +5,11 @@ package instance
 import (
 	"context"
 	"errors"
+	"slices"
 
-	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
-	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-go/tftypes"
 )
-
-// errUnexpectedPathValue is returned when walking an attribute path yields
-// something other than a value, which should not happen for a well-formed plan
-// or state.
-var errUnexpectedPathValue = errors.New("unexpected type at attribute path")
 
 // The plugin framework marks every computed attribute that is null in the
 // configuration as unknown whenever the planned state differs from the prior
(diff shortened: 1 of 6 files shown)


morpheus/framework/resources/instance/instance_plan_modifiers.go
```diff
@@ -0,0 +1,226 @@
...
+	attribute string,
+) {
+	p := path.Root(attribute)
+
+	// These attributes are all collections; read them as their concrete types so
+	// the framework performs the conversion.
+	switch attribute {
+	case "labels":
+		var value types.Set
+		if diags := req.State.GetAttribute(ctx, p, &value); diags.HasError() {
```
jasonquigleyhpecom: 🟠 **Violation: Silently swallowed diagnostics**

When `req.State.GetAttribute()` returns an error diagnostic, this code returns early but fails to append `diags` to `resp.Diagnostics`.

This swallows the framework error silently, masking potential state schema mismatches or provider bugs from the practitioner. You must append the error diagnostics before returning:

```go
if diags := req.State.GetAttribute(ctx, p, &value); diags.HasError() {
	resp.Diagnostics.Append(diags...)
	return
}
```
*(This also applies to the `default:` block below at line 220).*

jasonquigleyhpecom (dismissed):
---

**PR:** [#1312 — MORPH-14919: Stop unrelated computed attributes churning the instance plan](https://github.com/HPE/terraform-provider-hpe/pull/1312)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 6 files · agents: patterns, structure · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 1 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

**Verdict:** 🔴 **FAIL** — 1 issues found (1 critical)

---

### ⚙️ 4. Pattern Compliance

🟠 **Violation: Silently swallowed diagnostics in plan modifier**
- **File:** `morpheus/framework/resources/instance/instance_plan_modifiers.go`
- **Lines:** 211-213, 218-220
- **Description:** In `restoreAttributeFromState`, when `req.State.GetAttribute()` returns an error diagnostic, the code returns early but fails to append `diags` to `resp.Diagnostics`. This swallows the framework error silently, masking potential state schema mismatches or provider bugs from the practitioner. You must append the error diagnostics before returning (see inline comment).

- 🟠 **Violation:** `morpheus/framework/resources/instance/sweep/sweep.go` does not use the mandated `testsweep.WithIgnoreListStatuses[T](http.StatusNotFound, http.StatusForbidden)` parameter during sweeper registration to handle missing resources gracefully. Both `sweeperName` and `unmanagedServerSweeperName` lack this requirement.

---

morpheus/framework/resources/instance/instance_plan_modifiers.go
eamonnotoole: Fixed in `8efa36aff`. The diagnostics from `req.State.GetAttribute` are now appended to `resp.Diagnostics` before returning, in both branches:

```go
diags := req.State.GetAttribute(ctx, p, &value)
resp.Diagnostics.Append(diags...)

if diags.HasError() {
    return
}
```

I also added an early exit in the calling loop, so one failure is reported once rather than repeated for each remaining attribute.

Worth flagging that reviewing this file closely was well founded: acceptance testing against a live appliance then caught a second, more serious problem in the same function. Restoring whole collections could pin a value that the post-apply read subsequently overrides, producing `Provider produced inconsistent result after apply` on `volumes[].storage_profile` — the exact failure class this change is meant to prevent. It was intermittent, roughly one run in three.

The restore is now per attribute rather than per collection: only stable identity attributes are eligible, and a value is only filled where the plan is unknown and prior state is known and non-null, so a null can never be pinned. `TestUnitRestoreFromStateNeverPinsNull` covers it, and the suite was run five consecutive times against a live appliance with no failures. Details are in the PR description.

jasonquigleyhpecom (commented):
---

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 6 files · agents: patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

jasonquigleyhpecom (approved):

PR #1313, opened 2026-07-29
Fix acceptance test failures across load balancer, network and integration resources

jasonquigleyhpecom:
Addresses 35 of the 71 acceptance test failures in the 2026-07-29 run.

## Provider defects

- **Load balancer profile** — the config block was selected by response shape rather than `service_type`. The config schema has no discriminator, so the SDK decodes the same payload into every `anyOf` variant and keeps any that is not empty; `ntlmAuthentication` appears in the generic NSX-T blob, so the HTTP variant was always present. Every non-HTTP profile therefore reconstructed `config_http` on import and read another variant's tags. Both paths now switch on `service_type`.
- **Storage bucket** — create always failed with `unexpected end of JSON input`, because the `config` oneOf is always serialised but marshals to nothing when no variant is set. Fixing it exposed two silent bugs in the same path: credentials were sent at the top level of the body, where the API discards them, and the write-only `access_key`/`secret_key` were read from the plan, where write-only attributes are always null.
- **BGP neighbour** — create failed with `General error has occurred.` whenever `bfd_interval`/`bfd_multiple` were omitted. They were not sent, and the API forwards a BFD object to NSX-T without dropping unset members, so NSX-T rejected the body during validation and masked the real error underneath. The provider now always sends them, using the same defaults the UI applies.
- **Integrations and tasks** — Chef, vRO and the Chef bootstrap task wrote server-side secret hashes into state, causing permanent diffs. The vRO `tenant` is not a secret but was stored as a hash; that read-back and the `DiffSuppressFunc` masking it are both removed. The vRO update path also sent the tenant under a different key than create.

## Test fixtures

Several suites failed before reaching the provider at all. The most common cause was fixtures building overrides from `os.Getenv` and assigning unconditionally, so an unset variable replaced a working default with `""` and the template rendered invalid HCL. `testhelpers.MergeOverrides` now ignores empty overrides, and tests that genuinely need a pre-provisioned fixture skip explicitly.

Firewall rule tests never supplied the required `parent_id`. They now build an NSX-T Tier-1 gateway and their own rule group and reference it directly, matching the fixture already used by the passing rule group suites, so they are self-contained and need no new environment variable. The published example demonstrates the documented lookup through the `hpe_morpheus_network_router_firewall_rule_group` data source instead of a placeholder id.

Also fixed: `config_http` written as a block where the schema is an attribute; a data source test referencing a router it never rendered and asserting against an always-empty list; a required variable declared with no value, failing the plan before the provider ran; `hop_limit` and `restart_mode` values the API rejects; and two by-name data sources read before the resource they looked up existed.

## Verification

`go build`, `make lint` (0 issues), `make unit-tests`, and all `morpheus/...` packages pass without `TF_ACC`. `make docs` is idempotent. New unit tests cover profile variant selection, the storage bucket request body, and the BFD defaults; the profile test was confirmed to fail against the previous behaviour before the fix was applied.

Acceptance tests run against a firewalled appliance rather than in CI, so they are not part of the automated checks on this PR; they are exercised manually.

## Not included

Two failures need an API specification change plus an SDK regeneration and are deliberately out of scope here: load balancer profile `tags` key names, and `network.pool` being sent as a bare integer where an object is expected. The remaining failures in the run are missing appliance fixture data rather than code defects.

### DIFF
--- a/docs/resources/morpheus_load_balancer_monitor.md
+++ b/docs/resources/morpheus_load_balancer_monitor.md
@@ -140,3 +140,5 @@ Load balancer monitors can be imported using the composite ID `loadBalancerId.mo
 ```shell
 terraform import hpe_morpheus_load_balancer_monitor.example 42.101
 ```
+
+~> The API accepts `receive_data` and `data_length` on create and update but never returns them (`receive_data` is stored encrypted, and `data_length` is only round-tripped for ICMP monitors). They therefore cannot be populated on import and will be null in the imported state. Set them in your configuration and apply to reconcile.
--- a/docs/resources/morpheus_network_router_bgp_neighbor.md
+++ b/docs/resources/morpheus_network_router_bgp_neighbor.md
@@ -19,6 +19,7 @@ resource "hpe_morpheus_network_router_bgp_neighbor" "example" {
   weight      = 100
   keep_alive  = 60
   hold_down   = 180
+  hop_limit   = 2
 }
 ```
 
--- a/docs/resources/morpheus_network_router_firewall_rule.md
+++ b/docs/resources/morpheus_network_router_firewall_rule.md
@@ -11,8 +11,16 @@ description: |-
 ## Example Usage
 
 ```terraform
+# Firewall rules attach to a rule group on the router. Look the group up and
+# use its external_id as the rule's parent_id.
+data "hpe_morpheus_network_router_firewall_rule_group" "example" {
+  router_id = 1
+  name      = "Example Firewall Rule Group"
+}
+
 resource "hpe_morpheus_network_router_firewall_rule" "example" {
   router_id = 1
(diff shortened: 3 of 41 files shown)


eamonnotoole (commented):
---

## 📋 Review: PR #1313

**PR:** [#1313 — Fix acceptance test failures across load balancer, network and integration resources](https://github.com/HPE/terraform-provider-hpe/pull/1313)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 41 files · agents: review-patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

eamonnotoole (approved):

PR #1314, opened 2026-07-30
Fix BGP neighbor config on import and unblock acceptance tests

jasonquigleyhpecom:
## Summary

Addresses the genuine code/fixture defects in the 2026-07-30 acceptance run — as opposed to environment drift or backend outages, which make up the rest.

**Three fixes are confirmed working against the 2026-07-31 nightly.** A fourth (storage bucket) turned out to need infrastructure the appliance does not have, so it is now gated honestly instead of failing permanently.

> **Draft:** the remaining verification is a targeted run on the firewalled appliance. Everything verifiable offline is green.

## Verified against the 2026-07-31 nightly

Confirmed this build contained the changes before drawing conclusions: the storage bucket request now carries a populated `bucketName`, and the load balancer profile request no longer carries `tags`.

| Test | 07-30 | 07-31 |
|---|---|---|
| `…NetworkRouterBgpNeighborResourceImport` | fail | **pass** |
| `…PolicyResourceTypesOk` | fail | **pass** |
| `…LoadBalancerProfileResourceHttpExampleOk` | fail | **pass** |
| `…StorageBucketResource{Example,Update}Ok` | fail | fail → now gated |

**No regressions.** All 10 BGP neighbor tests pass, including `…ResourceWithNsxtConfig` — the real risk, since change 1 touches the shared read path.

Total failures went 38 → 35 on that run. With the gating change the next run should reach 33.

### 1. BGP neighbor: config blocks lost on import — *provider bug*

`getNetworkRouterBgpNeighborAsState` decided which typed config block to populate from the **plan**. `ImportState` seeds only `router_id` and `id`, so on import every plan value is null and the branches never ran. Read has to be self-sufficient.

Now driven by the **content of the API response**, with the plan only as a secondary trigger:

- `config_nsxt` populates when the response carries `sourceAddresses`, or when the plan already had the block set.
- `config_nsxv` populates when the response carries `routerId`/`interface`, or when the plan had it set.
- The plan test is now `!IsNull() && !IsUnknown()`.

That last point also fixes a **latent bug affecting every NSX-T apply**. An unset `Optional+Computed` block arrives *unknown*, not null, so the old `!IsNull()` guard read "user did not set it" as "user set it". Combined with the config schema being an `anyOf` whose variants each carry a `,remain` catch-all — an NSX-T payload decodes into a non-nil NSX-V variant too, because the unrecognised key lands in `AdditionalProperties` and the struct survives the SDK's `IsEmpty` check — this wrote a junk `config_nsxv` object of two nulls into state on every apply. That is now gone.

Content alone was not a safe gate: both blocks are `Optional+Computed`, so a user writing `config_nsxt = {}` would have gone `object → null` and tripped "Provider produced inconsistent result after apply". Hence the two-trigger form.

### 2. BGP neighbor: `description` not returned by the API

Confirmed from the run's HTTP traces — the GET response carries no `description` key at all. The resource already falls back to the planned value on apply, but import has nothing to fall back to, so the round trip cannot work. Added to `ImportStateVerifyIgnore` alongside `password_wo`, with the reason recorded inline.

### 3. Load balancer profile: `tags` rejected by NSX-T

Failed with `400 {"msg":"Json de-serialization error: property name is unrecognized."}`.

Morpheus models profile tags as `{name, value}` — which is what this provider and the API spec describe — but passes the list straight through to NSX-T on the profile create path without translating it. NSX-T tags are `{scope, tag}`. The reverse direction *is* translated on sync, and the load balancer **pool** path translates on the way out too, so this is an upstream gap specific to profile create.

Removed the `tags` block and its assertion, with a comment describing what to restore once the platform side is fixed. Every other profile test passed — this was the only one exercising `tags`.

Confirmed fixed on 07-31: the profile was created and the API echoed back the description.

### 4. Policy test: stale hardcoded `pool_id`

Failed on its *inline network dependency*, not on any policy:

```
400 {"errors":{"pool":"Could not find matching constructor for: com.morpheus.NetworkPool(Integer)"}}
```

The network create path binds `pool` into the pool association before its own scalar-id handler runs, so a bare integer — exactly what the API spec declares — fails for any network type that allows pool assignment. The provider payload is spec-conformant; this is an upstream defect.

Dropped `pool_id` from the test's scaffolding network. The test asserts policy scoping across associated resource types; the IP pool is irrelevant to that, and the ID was hardcoded and already drifting. Pool coverage stays in the network package's own tests.

### 5. Storage bucket: two-stage fix

Originally failed with `{"errors":{"bucketName":"bucketName is required"}}` — the example template set only `name` and `provider_type`. Added `bucket_name` to the template, render defaults and `go:generate` line, and regenerated the example and docs.

That cleared the reported error and revealed the real blocker on 07-31:

```
400 {"msg":"Error Validating Storage Provider: Unable to load AWS credentials
     from any provider in the chain: ..."}
```

Before saving a bucket the API performs a **real round-trip** — write a test object, read it back, delete it. With `provider_type = "s3"` and no credentials that can never pass, and there is no offline path: `storage_server` (which would bypass validation) and `create_bucket` are not exposed in the resource schema.

These tests were gated on `alletra` — **block** storage, unrelated to the **object** storage they exercise. That mis-gate is why they ran, and failed, on an appliance with no AWS credentials. Re-gated on the existing `aws` capability, which already signals AWS credentials are configured.

`bucket_name` stays — it is genuinely required for s3 and will matter wherever the test does run. It also keeps the published documentation example realistic.

## Verification

| Check | Result |
|---|---|
| `go build ./...` | pass |
| `go vet` (touched packages) | pass |
| `make lint` | **0 issues** |
| `make unit-tests` | pass |
| Package unit tests (`-short -skip TestAcc`) | pass |
| `make docs` | idempotent |
| Capability gate, without `aws` | skips: `missing required capabilities: aws` |
| Capability gate, with `aws` | passes through to the test body |

### Acceptance tests

```bash
TF_ACC=1 TF_ACC_CAPABILITIES=network_router \
  go test ./morpheus/framework/resources/networkrouterbgpneighbor/... -count 1 -v

TF_ACC=1 TF_ACC_CAPABILITIES=network_dhcp \
  go test ./morpheus/framework/resources/policy/... \
    -run TestAccMorpheusPolicyResourceTypesOk -count 1 -v

TF_ACC=1 TF_ACC_CAPABILITIES=nsxt \
  go test ./morpheus/framework/resources/loadbalancerprofile/... -count 1 -v

# only meaningful on an appliance with AWS credentials
TF_ACC=1 TF_ACC_CAPABILITIES=aws \
  go test ./morpheus/framework/resources/storagebucket/... -count 1 -v
```

## Known risks

- Change 1 alters state shape for NSX-T users: `config_nsxv` no longer appears as an empty object. Behaviourally correct — it only ever held nulls — but worth a changelog line.
- `AWS` sits under *Cloud Types* and strictly means "an AWS cloud is configured". If `aws` is added to the capability set to pick up the existing AWS-gated tests, the bucket tests come with it and will fail unless that appliance also has usable S3 credentials. Narrow in practice, but the reason is recorded at the gate.

## Follow-ups (not in this PR)

- Platform: network create binds a scalar `pool` before its own scalar handler, breaking pool assignment for affected network types.
- Platform: load balancer profile create does not translate tags to the NSX-T `{scope, tag}` shape, unlike the pool path.
- API: BGP neighbor GET does not return `description`, making import lossy.
- Bundled SDK: `bucketName` and `config` are non-pointer and always serialised on storage bucket create, so an unset optional still ships `""`. Drifted from the current spec; needs an SDK regeneration.

The remaining failures are environment issues — stale hardcoded IDs, missing named fixtures, an unreachable placeholder integration host, and NSX-T backend failures — and are out of scope here.

⚠️ For whoever picks up the fixture work: the network resource tests also hardcode the same `pool_id`. They currently fail earlier on a stale cloud id, so fixing that may immediately surface the same pool-binding defect side-stepped in change 4.

⚠️ The NSX-T failures are flaky, not deterministic: between the two runs `LoadBalancerPoolResourceNsxtExampleOk` went fail→pass while `…NsxtUpdateOk` went pass→fail, with an identical signature. Four tests still leave dangling resources via failed post-test destroy; a sweeper run is worth doing regardless.


eamonnotoole (commented):
---

## 📋 Review: `multiple resources`

**PR:** [#1314 — Fix BGP neighbor config on import and unblock acceptance tests](https://github.com/HPE/terraform-provider-hpe/pull/1314)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 9 files · agents: patterns, structure · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

- 🔴 **Gap**: `morpheus/framework/resources/storagebucket` lacks a paired data source. Every resource must have a corresponding data source with full test coverage, but `morpheus/framework/datasources/storagebucket` does not exist.
- 🟠 **Violation**: File completeness: the `storagebucket` resource package does not split CRUD operations into separate files (`create.go`, `read.go`, `update.go`, `delete.go`, `import.go`). They are currently all combined inside `resource.go`.
- 🟠 **Violation**: `morpheus/framework/resources/storagebucket/resource_test.go` lacks the required copyright header (`// (C) Copyright 2026 Hewlett Packard Enterprise Development LP`).

---

eamonnotoole (approved):

PR #1324, opened 2026-08-05
Add hpe_morpheus_instance_node resource (MORPH-13978)

eamonnotoole:
Implements MORPH-13978. Design agreed in an internal plan.

Adds and removes a single node on an existing instance. On bare metal each node
can be placed in its own resource pool, which is what lets an HA instance span
several failure domains.

## Not restricted to bare metal

The add-node action is generic and works on virtual instances — verified on a
live appliance against kvm. Only `selectedResourcePoolId` is bare-metal
specific, because only the HPE provision provider reads it, and **no other
provider validates it**: sending a pool to a VMware or KVM instance returns
`200 success` while placing the node in the instance's own pool.

So `resource_pool_id` is optional, and the provision-type check fires only when
a pool is actually set. That case is refused rather than allowed to succeed
misleadingly.

The check keys on `layout.provisionTypeCode == "hpe-baremetal-plugin.provision"`,
not `format == "bareMetal"` — core seeds `bareMetalHost: true` for eleven server
types, and only the HPE plugin honours the pool.

## Wire format

Parameters go inside an `{"action": {...}}` envelope. The action endpoints do not
read a top-level JSON body, which is a controller defect rather than the intended
contract — 19 other API controllers merge the body via
`convertParamsToPayload()`; these two do not. Raised as MORPH-15280.

The generated request body is therefore deliberately unused: it would serialise
`selectedResourcePoolId` at the top level, where it is discarded silently. The
envelope was verified to deliver non-code keys (an envelope carrying `count=2`
produced exactly two containers). It works either side of the MORPH-15280 fix,
and the code comment says so, so the workaround has an expiry rather than looking
like design.

## Lifecycle details worth review

- **Container id** comes from the add-node response, which returns it
  synchronously for the provisioning path. The pre-provisioned path returns no
  ids, so there the containers are diffed against a pre-call snapshot.
- **`success: true` is not a health signal** — the API returns it even for a
  failed instance — so the container id's presence is the success criterion.
- **Delete waits for the container, not the server.** On bare metal `cleanServer`
  returns `removeServer: false` and the server persists as `available`. Waiting
  for it would hang every destroy until timeout.
- **The action code is per-layout** (`generic-add-node`, `ubuntu-add-node`), so
  it is discovered from `/actions` rather than assumed.
- **Volumes, networks and plan are not modelled.** They belong to the instance
  and a resize applies them to every node, so modelling them here would show as
  permanent drift.

## Plan-time pre-flight

The provision type is also checked in `ModifyPlan`, so a pool set against a
virtual instance fails before anything is built. It skips silently when the
instance id is unknown (the common first-apply case) or cannot be read, and
errors only on a positive determination — so it can only ever move an error
earlier, never invent one.

## Testing

Verified against a live appliance:

- `TestAccMorpheusInstanceNodeVirtual` — **passes**, adds and removes a node
- `TestAccMorpheusInstanceNodePoolOnVirtualError` — **passes**, guard fires

Multi-pool placement and the IP wait are gated on a new `metal` capability and
need an appliance running the bare-metal plugin.

24 unit tests cover the guard (including rejecting `format == "bareMetal"` with a
non-HPE provision type), the `ModifyPlan` skip cases, envelope construction
(`pre_provisioned` omitted rather than sent as `"off"`, which Groovy treats as
truthy), and the IP predicate (IPv6 ready, sentinels not).


morpheus/framework/resources/instancenode/create.go
```diff
@@ -0,0 +1,381 @@
...
+			ctx, client, instanceID, containerID, createTimeout,
+		)
+		if waitErr != nil {
+			resp.Diagnostics.AddError(
+				"IP address wait failed",
+				fmt.Sprintf("instance %d container %d: %s",
+					instanceID, containerID, waitErr.Error()),
+			)
+
+			return
```
rory-mcevoy-hpe, line 224: 🔴 **Gap — State Leak on IP Wait Error**
If `containerip.Wait` fails, this returns immediately without ever calling `resp.State.Set(ctx, &plan)` or `cleanup.TaintResourceState()`. Because it returns before saving state, Terraform loses track of the provisioned container entirely, leaking it on the appliance.

morpheus/framework/resources/instancenode/create.go
```diff
@@ -0,0 +1,381 @@
...
+		resp.Diagnostics.AddWarning(
+			"node partially created",
+			fmt.Sprintf("instance %d container %d was added but state "+
+				"read failed: %s. On next apply, Terraform will attempt "+
+				"to reconcile.", instanceID, containerID, readErr.Error()),
+		)
+		// Still set state so the container_id is tracked.
+		resp.Diagnostics.Append(resp.State.Set(ctx, &plan)...)
+
+		return
```
rory-mcevoy-hpe, line 200: 🟠 **Violation — Post-Create Failure Tainting**
If `refreshNodeState` fails, partial state is set and a warning is added, but it omits `cleanup.TaintResourceState()`. This leaves the resource tracked but in a potentially broken state without prompting Terraform to recreate it, deviating from the proven post-create failure pattern.

morpheus/framework/resources/instancenode/create_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+package instancenode
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+func TestBuildAddNodeEnvelope_PreProvisionedFalse(t *testing.T) {
```
rory-mcevoy-hpe: 🟠 **Violation — Test Naming Convention**
Unit tests do not follow the established `TestUnit*` naming pattern found throughout the provider. This should be prefixed with `TestUnit` (e.g. `TestUnitBuildAddNodeEnvelope_PreProvisionedFalse`).

morpheus/framework/resources/instancenode/metal_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+package instancenode
+
+import (
+	"testing"
+
+	sdk "github.com/HPE/terraform-provider-hpe/internal/sdk/oapigen"
+)
+
+func TestIsMetal_CorrectCode(t *testing.T) {
```
rory-mcevoy-hpe: 🟠 **Violation — Test Naming Convention**
This should be prefixed with `TestUnit` (e.g. `TestUnitIsMetal_CorrectCode`).

morpheus/framework/resources/instancenode/modifyplan_test.go
```diff
@@ -0,0 +1,110 @@
...
+	if diags := p.Set(ctx, &m); diags.HasError() {
+		t.Fatalf("build plan: %v", diags)
+	}
+
+	return p
+}
+
+// TestModifyPlan_NoPoolOnNonMetal_NoDiagnostics verifies that ModifyPlan
+// produces no diagnostics when resource_pool_id is null (virtual path).
+func TestModifyPlan_NoPoolOnNonMetal_NoDiagnostics(t *testing.T) {
```
rory-mcevoy-hpe: 🟠 **Violation — Test Naming Convention**
This should be prefixed with `TestUnit` (e.g. `TestUnitModifyPlan_NoPoolOnNonMetal_NoDiagnostics`).

morpheus/framework/resources/instancenode/read.go
```diff
@@ -0,0 +1,116 @@
...
+		return fmt.Errorf("instance %d: response is nil", instanceID)
+	}
+
+	for i := range getResp.Instance.ContainerDetails {
+		cd := &getResp.Instance.ContainerDetails[i]
+		if cd.Id != nil && *cd.Id == containerID {
+			state.ContainerID = types.Int64Value(containerID)
+
+			if cd.Server != nil && cd.Server.Id != nil {
+				state.ServerID = types.Int64Value(*cd.Server.Id)
```
rory-mcevoy-hpe: 🔵 **Recommendation — Manual Pointer Conversion**
While functionally safe, using the provider's `utils/convert` package (i.e. `convert.Int64ToType(cd.Server.Id)`) is slightly more idiomatic and abstracts the inner nil check.

rory-mcevoy-hpe (dismissed):
---

## 📋 Review: `hpe_morpheus_instance_node`

**PR:** [#1324 — Implements MORPH-13978](https://github.com/HPE/terraform-provider-hpe/pull/1324)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 24 files · agents: patterns, structure · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 3 |
| 🟠 Violations | 2 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 1 |

**Verdict:** 🔴 **FAIL** — 6 issues found (5 critical)

---

### 📁 1. File Completeness

🔴 **Missing Paired Data Source (`morpheus/framework/datasources/instancenode/`)**
The directory and its required files (`datasource.go`, `schema_gen.go`, `*_test.go`) are absent. A corresponding data source must be implemented and registered in `morpheus/datasources.go`.

🔴 **Missing Import File (`morpheus/framework/resources/instancenode/import.go`)**
The `import.go` file is missing. Even though the documentation notes that instance nodes cannot be imported, the file must exist and explicitly return a not-supported error to comply with the provider structure checklist.

---

morpheus/framework/resources/instancenode/create.go
matthew-c-hpe, line 224: 🔴 **Gap:** `Create` drops state and leaks resource on IP wait failure.

If `containerip.Wait` returns an error, the function adds an Error diagnostic and returns immediately without calling `resp.State.Set(ctx, &plan)`. Because the container was already successfully added to the instance upstream and its `ContainerID` was placed in the plan, skipping `resp.State.Set` discards the tracked identity from Terraform state entirely, causing a permanent resource leak. You must call `resp.State.Set(ctx, &plan)` before returning to ensure the created node is tracked.

morpheus/framework/resources/instancenode/resource.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// ConfigValidators implements resource.ResourceWithConfigValidators.
+func (r *Resource) ConfigValidators(
+	_ context.Context,
+) []resource.ConfigValidator {
+	return []resource.ConfigValidator{
+		resourcevalidator.RequiredTogether(
+			path.MatchRoot("pre_provisioned"),
+			path.MatchRoot("selected_server_id"),
```
matthew-c-hpe: 🟠 **Violation:** Flawed `RequiredTogether` validator on a boolean field.

The Terraform Plugin Framework's `RequiredTogether` validator triggers if *any* of the specified paths are configured with a known, non-null value. Because `pre_provisioned` is a boolean, if a user explicitly writes `pre_provisioned = false` in their HCL, the framework considers it configured/known. The validator will then incorrectly block the plan, demanding that `selected_server_id` also be set. You must replace this with a custom validator that only enforces the presence of `selected_server_id` when `pre_provisioned` is exactly `true`.

matthew-c-hpe (changes requested):
## 📋 Review: `hpe_morpheus_instance_node`

**PR:** [#1324 — Add hpe_morpheus_instance_node resource (MORPH-13978)](https://github.com/HPE/terraform-provider-hpe/pull/1324)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 24 files · agents: patterns, structure, spec-coverage · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 2 |
| 🟠 Violations | 1 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 1 |

**Verdict:** 🔴 **FAIL** — 4 issues found (3 critical)

### 🔴 Gaps

- **Sweeper dependency missing on the parent resource.** Because `instancenode` is a child of `instance`, the `instance` sweeper must be updated to depend on it. Otherwise, both sweepers may run concurrently and conflict when deleting test resources. Update `morpheus/framework/resources/instance/sweep_import.go` to blank-import `instancenode/sweep` and add it to `testsweep.WithDependencies` in the instance sweeper registration. *(Cannot comment inline as the instance sweeper file is not in this PR's diff)*.
- **`Create` drops state and leaks resource on IP wait failure** (see inline comment in `create.go`).

### 🟠 Violations

- **Flawed `RequiredTogether` validator on a boolean field** (see inline comment in `resource.go`).

### 🔵 Recommendations

- **Expose additional node metadata:** The `instancenode` schema currently exports `ip_address`, `container_id`, and `server_id`. The underlying `InstanceContainer3` SDK model read back in `refreshNodeState` contains many other potentially useful read-only properties (like `hostname`, `internal_ip`, `external_fqdn`, and `mac_address`). Consider adding these as computed attributes.

---

morpheus/framework/resources/instancenode/create.go
matthew-c-hpe, line 224: 🔴 **Gap:** `Create` drops state and leaks resource on IP wait failure.

If `containerip.Wait` returns an error, the function adds an Error diagnostic and returns immediately without calling `resp.State.Set(ctx, &plan)`. Because the container was already successfully added to the instance upstream and its `ContainerID` was placed in the plan, skipping `resp.State.Set` discards the tracked identity from Terraform state entirely, causing a permanent resource leak. You must call `resp.State.Set(ctx, &plan)` before returning to ensure the created node is tracked.

morpheus/framework/resources/instancenode/resource.go
matthew-c-hpe: 🟠 **Violation:** Flawed `RequiredTogether` validator on a boolean field.

The Terraform Plugin Framework's `RequiredTogether` validator triggers if *any* of the specified paths are configured with a known, non-null value. Because `pre_provisioned` is a boolean, if a user explicitly writes `pre_provisioned = false` in their HCL, the framework considers it configured/known. The validator will then incorrectly block the plan, demanding that `selected_server_id` also be set. You must replace this with a custom validator that only enforces the presence of `selected_server_id` when `pre_provisioned` is exactly `true`.

matthew-c-hpe (changes requested):
## 📋 Review: `hpe_morpheus_instance_node`

**PR:** [#1324 — Add hpe_morpheus_instance_node resource (MORPH-13978)](https://github.com/HPE/terraform-provider-hpe/pull/1324)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 24 files · agents: patterns, structure, spec-coverage · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 2 |
| 🟠 Violations | 1 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 1 |

**Verdict:** 🔴 **FAIL** — 4 issues found (3 critical)

### 🔴 Gaps

- **Sweeper dependency missing on the parent resource.** Because `instancenode` is a child of `instance`, the `instance` sweeper must be updated to depend on it. Otherwise, both sweepers may run concurrently and conflict when deleting test resources. Update `morpheus/framework/resources/instance/sweep_import.go` to blank-import `instancenode/sweep` and add it to `testsweep.WithDependencies` in the instance sweeper registration. *(Cannot comment inline as the instance sweeper file is not in this PR's diff)*.
- **`Create` drops state and leaks resource on IP wait failure** (see inline comment in `create.go`).

### 🟠 Violations

- **Flawed `RequiredTogether` validator on a boolean field** (see inline comment in `resource.go`).

### 🔵 Recommendations

- **Expose additional node metadata:** The `instancenode` schema currently exports `ip_address`, `container_id`, and `server_id`. The underlying `InstanceContainer3` SDK model read back in `refreshNodeState` contains many other potentially useful read-only properties (like `hostname`, `internal_ip`, `external_fqdn`, and `mac_address`). Consider adding these as computed attributes.

---

morpheus/framework/resources/instancenode/create.go
eamonnotoole: Confirmed and fixed in a499d8b.

You were right that it returned before `resp.State.Set`, so a node that existed on the appliance was dropped from state entirely.

Fixed via `cleanup.TaintResourceState(...)` rather than a bare `State.Set` — my first attempt saved state and let the error diagnostic taint implicitly, which is not the pattern here. It now matches `loadbalancerpool/create.go`, so the node is recorded and tainted for replacement.

morpheus/framework/resources/instancenode/create.go
eamonnotoole: Fixed in a499d8b — same treatment as the wait path.

`refreshNodeState` failure now goes through `cleanup.TaintResourceState(...)` with the config struct, matching `loadbalancerpool/create.go`. Previously it set partial state and warned, so the resource stayed tracked in a state we could not confirm.

morpheus/framework/resources/instancenode/create_test.go
eamonnotoole: Renamed in a499d8b: `TestBuildAddNodeEnvelope_*` and `TestExtractContainerIDFromResults_*` now carry the `TestUnit` prefix. Applied across `metal_test.go`, `modifyplan_test.go` and `utils/containerip` too.

For the record, it is 32 `TestUnit*` against 33 plain `Test*` in the provider, so it reads as a preference rather than a settled rule — but it costs nothing and matches the direction of travel, so happy to follow it.

morpheus/framework/resources/instancenode/metal_test.go
eamonnotoole: Renamed in a499d8b — `TestUnitIsMetal_*`, `TestUnitProvisionTypeCode_*`, `TestUnitNotMetalDetail_*`.

morpheus/framework/resources/instancenode/modifyplan_test.go
eamonnotoole: Renamed in a499d8b — `TestUnitModifyPlan_*`.

morpheus/framework/resources/instancenode/read.go
eamonnotoole: Adopted in a499d8b — `convert.Int64ToType(cd.Server.Id)` for the server id.

Worth noting I nearly dismissed this one: my first search was scoped to `morpheus/` and found nothing, so I went looking properly and it is at the top-level `utils/convert/convert.go:147`. Good rec.

I left the IP conversion as-is, since it carries an extra `containerip.Ready()` semantic check beyond the nil test rather than being a plain unwrap. Say the word if you would rather it were uniform.

morpheus/framework/resources/instancenode/create.go
eamonnotoole: Confirmed and fixed in a499d8b — you and Rory flagged this independently.

The container was already created and its id was in `plan`, so returning without recording it lost a real node. Both this path and the `refreshNodeState` path now go through `cleanup.TaintResourceState(...)`, matching `loadbalancerpool/create.go`, so the node is tracked and tainted rather than abandoned.

morpheus/framework/resources/instancenode/resource.go
eamonnotoole: Good catch — confirmed against the validator source and fixed in a499d8b.

`internal/configvalidator/required_together.go` counts a path as configured when `!value.IsNull()`, so an explicit `pre_provisioned = false` is configured and would have demanded `selected_server_id`. Exactly as you described.

Replaced with `preProvisionedRequiresServerID`, which applies only when the value is exactly `true` and ignores null and unknown. Four unit tests in `validator_test.go`, including the regression case: `pre_provisioned = false` alone must not error.

morpheus/framework/resources/instancenode/create.go
eamonnotoole: Fixed in a499d8b — see the reply on the duplicate thread above. Both post-create failure paths now record and taint via `cleanup.TaintResourceState(...)`.

morpheus/framework/resources/instancenode/resource.go
eamonnotoole: Fixed in a499d8b — see the reply on the duplicate thread above. Replaced with a validator that only applies when `pre_provisioned` is exactly `true`, with the `= false` regression covered by a test.

morpheus/framework/resources/instancenode/resource_test.go
matthew-c-hpe, line 1: Could we move the contents of acceptance_test.go into this file

matthew-c-hpe (dismissed):

matthew-c-hpe (approved):

morpheus/framework/resources/instancenode/resource.go
```diff
@@ -0,0 +1,285 @@
...
+func (r *Resource) Metadata(
+	_ context.Context,
+	req resource.MetadataRequest,
+	resp *resource.MetadataResponse,
+) {
+	resp.TypeName = req.ProviderTypeName + "_instance_node"
+}
+
+// Schema implements resource.Resource.
+func (r *Resource) Schema(
```
matthew-c-hpe, line 73: 🔴 **Gap:** Bypassed code generation. The schema must be generated via `config.yaml` in the `specs/` directory to produce `schema_gen.go`. Hand-coding `Schema()` inside `resource.go` is not permitted, even if the API uses an unusual envelope.

morpheus/framework/resources/instancenode/resource.go
```diff
@@ -0,0 +1,285 @@
...
+
+// Interface compliance assertions.
+var (
+	_ resource.Resource                     = &Resource{}
+	_ resource.ResourceWithModifyPlan       = &Resource{}
+	_ resource.ResourceWithConfigValidators = &Resource{}
+)
+
+// instanceNodeModel is the Terraform state model.
+type instanceNodeModel struct {
```
matthew-c-hpe, line 30: 🔴 **Gap:** Missing `id` attribute in schema and model. The `cleanup.TaintResourceState` helper explicitly sets the ID using `path.Root("id")`. Without an `id` attribute defined in the schema, this hardcoded state write will fail with a schema error, breaking the post-create failure taint handling.

🔵 **Recommendation:** You should also consider exposing `name` and `uuid` attributes on the resource as `Computed: true` to provide practitioners with the node's system-generated name and UUID.

morpheus/framework/resources/instancenode/sweep/sweep.go
```diff
@@ -0,0 +1,72 @@
...
+			return getsafe.Get(&resp.Instances), hresp, err
+		},
+		// Is this a test instance?
+		func(item sdk.ListInstances200ResponseAllOfInstancesInner) bool {
+			name, ok := getsafe.GetOk(item.Name)
+			if !ok || name == nil {
+				return false
+			}
+
+			return strings.HasPrefix(*name, testsweep.TestResourcePrefix)
```
matthew-c-hpe, line 46: 🟠 **Violation:** Sweeper uses incorrect global prefix. The sweeper defines a local constant `TestResourcePrefix = "TestAccMorpheusInstanceNode"` but actually evaluates `strings.HasPrefix(*name, testsweep.TestResourcePrefix)` (which resolves to the global `"TestAccMorpheus"`). This broad match will incorrectly sweep all test instances across the entire provider, not just `instancenode` fixtures.

morpheus/framework/resources/instancenode/import.go
```diff
@@ -0,0 +1,38 @@
+package instancenode
```
matthew-c-hpe, line 3: 🟠 **Violation:** Missing copyright header.

morpheus/framework/resources/instancenode/import_test.go
```diff
@@ -0,0 +1,36 @@
+package instancenode
```
matthew-c-hpe, line 3: 🟠 **Violation:** Missing copyright header.

morpheus/framework/resources/instancenode/instance_node_example.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"fmt"
+	"path/filepath"
+	"runtime"
+	"testing"
+
+	"github.com/HPE/terraform-provider-hpe/morpheus/testhelpers"
+)
+
+//go:generate ../../../../bin/render -out examples/resources/hpe_morpheus_instance_node/example.tf example.tf.tmpl InstanceId 100 PoolB 2 PoolC 3
```
matthew-c-hpe, line 14: 🟠 **Violation:** Line length exceeds the 120 character limit (enforced by `lll`).

morpheus/framework/resources/instancenode/read.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/hashicorp/terraform-plugin-log/tflog"
+
+	sdk "github.com/HPE/terraform-provider-hpe/internal/sdk/oapigen"
+	"github.com/HPE/terraform-provider-hpe/morpheus/utils/containerip"
+	"github.com/HPE/terraform-provider-hpe/morpheus/utils/errfmt"
+	"github.com/HPE/terraform-provider-hpe/utils/convert"
+)
+
+// Read implements resource.Resource.
+func (r *Resource) Read(
```
matthew-c-hpe, line 21: 🟡 **Warning:** Unread `resource_pool_id`. The `resource_pool_id` is an `Optional` field that forces replacement but is not read back from the API. If out-of-band drifts are intended to be detected and reverted/replaced, you must read the field back.

morpheus/framework/resources/instancenode/read.go
```diff
@@ -0,0 +1,148 @@
...
+		return
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &state)...)
+}
+
+// refreshNodeState reads the instance and populates state fields for the
+// given container. If the container is not found, state.ContainerID is set
+// to null to signal removal.
+func refreshNodeState(
```
matthew-c-hpe: 🔵 **Recommendation:** Rename read helpers. Consider renaming `refreshNodeState` (and `populateNodeMetadata`) to follow the standard `getInstanceNodeAsState` naming convention.

matthew-c-hpe (changes requested):
---

## 📋 Review: `hpe_morpheus_instance_node`

**PR:** [#1324 — Add hpe_morpheus_instance_node resource (MORPH-13978)](https://github.com/HPE/terraform-provider-hpe/pull/1324)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 29 files · agents: structure, patterns, spec-coverage · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 3 |
| 🟠 Violations | 3 |
| 🟡 Warnings | 1 |
| 🔵 Recommendations | 2 |

**Verdict:** 🔴 **FAIL** — 9 issues found (6 critical)

---

### 📁 1. File Completeness

- 🔴 **Gap:** Missing paired data source. The PR adds the `instancenode` resource but does not include the corresponding `hpe_morpheus_instance_node` data source package under `morpheus/framework/datasources/instancenode/` (nor its registration in `morpheus/datasources.go`). Every resource requires a paired data source with tests and examples.

---

morpheus/framework/resources/instancenode/resource.go
matthew-c-hpe: I think we can ignore this - it doesn't strictly map to an API resource, right?

morpheus/framework/resources/instancenode/resource.go
eamonnotoole: Confirmed and fixed in 61244d7b — this was the most valuable comment of the round.

You were exactly right: `cleanup.TaintResourceState` writes `path.Root("id")` directly, and the schema had no `id`. So both post-create failure paths — the ones added last round specifically to stop a created node being abandoned — would have failed with a schema error. They were broken in precisely the case they exist for.

Added `id` (the container id) and, more importantly, a test that drives the helper against the **real schema**. I checked it actually catches the bug: removing `id` makes it fail with the runtime error.

```
Error: AttributeName("id") still remains in the path: could not find attribute or block "id" in schema
```

My previous tests asserted the error path returned an error, which passes whether or not the state write works. That is the gap that let this through.

`container_id` is kept alongside `id`; they hold the same value, which the descriptions say. Happy to drop one if you would rather not have both.

Also took the `name` and `uuid` recommendation — both verified populated on live data.

morpheus/framework/resources/instancenode/sweep/sweep.go
eamonnotoole: Confirmed and fixed in 61244d7b. Good catch, and worse than a naming slip: a sweep would have destroyed other resources fixtures, not just this one.

It declared `TestResourcePrefix = "TestAccMorpheusInstanceNode"` and then evaluated `testsweep.TestResourcePrefix`, which resolves to the global `"TestAccMorpheus"`. Now uses the local constant.

I checked all 57 sweepers in the provider — **only this one** had the fault.

morpheus/framework/resources/instancenode/import.go
eamonnotoole: Fixed in 61244d7b. I checked every file added on this branch; these two were the only omissions.

morpheus/framework/resources/instancenode/import_test.go
eamonnotoole: Fixed in 61244d7b, along with `import.go`.

morpheus/framework/resources/instancenode/resource_test.go
eamonnotoole: Done in 61244d7b — `acceptance_test.go` merged into `resource_test.go` and deleted, with `TestMain` and all four tests preserved. This does match how other packages here lay them out.

morpheus/framework/resources/instancenode/read.go
eamonnotoole: Done in 61244d7b — `refreshNodeState` is now `getInstanceNodeAsState`, and `populateNodeMetadata` follows as `populateInstanceNodeMetadata`.

morpheus/framework/resources/instancenode/instance_node_example.go
eamonnotoole: I think this one is a false positive — happy to be corrected.

The line is a `//go:generate` directive, and `.golangci.yml` exempts exactly those from `lll`:

```yaml
      - linters:
          - lll
        source: '^//go:generate '
```

`golangci-lint run ./morpheus/framework/resources/instancenode/...` reports 0 issues, and repo-wide `make lint` is also clean, so CI is not flagging it. The directive cannot be wrapped without breaking it, which is presumably why the exemption exists.

Left as-is. If the exemption is considered wrong, that is a repo-config discussion rather than something to work around in this file.

morpheus/framework/resources/instancenode/read.go
eamonnotoole: Valid concern, and I tried your fix — it broke the virtual path, so I have solved it a different way in 61244d7b. Worth explaining, because the failure is instructive.

Reading `resource_pool_id` back means making it Computed. Terraform then carries that value into the next plan, and `ModifyPlan` can no longer distinguish "the practitioner asked for this pool" from "we read the server current pool". The metal guard fired against a virtual instance that had never requested a pool:

```
Error: resource_pool_id is only valid for bare-metal instances
```

That is the acceptance test for the virtual path failing on the post-apply plan.

The root problem is that `resource_pool_id` is a create-time **placement request**, not observable state. The server current pool is readable, but it is a different thing, and conflating them breaks the guard.

So: `resource_pool_id` stays `Optional` and is not read back, and the pool the server actually belongs to is exposed as a new computed `server_resource_pool_id`, populated for every instance type. Drift is visible by comparing the two, without the provider having to guess intent. A unit test asserts the request field is never written by a read, so this cannot regress silently.

morpheus/framework/resources/instancenode/resource.go
eamonnotoole: Leaving the schema hand-written, per your follow-up and the `instancepowerstate` precedent.

For the record on what I checked: `instancepowerstate` is the one other hand-written schema in the provider and has **no `specs/` entry**, and I could not find the codegen requirement documented anywhere in the repo. Both are action-driven resources whose attributes do not correspond to a single API schema — this one draws inputs from the add-node action and reads from `containerDetails`, plus provider-only attributes like `wait_for_ip_address` that no API schema carries.

Happy to revisit if a config.yaml can express it cleanly.

morpheus/framework/resources/instancenode/resource.go
```diff
@@ -0,0 +1,330 @@
...
```
matthew-c-hpe, line 30: 🔴 **Gap**: Hand-written state model and missing `schema_gen.go`. The schema and state model must be defined in the provider's `specs/` configuration and generated, rather than hand-edited, to ensure consistency and correct docs generation.

morpheus/framework/resources/instancenode/sweep/sweep.go
```diff
@@ -0,0 +1,72 @@
...
+
+	testsweep "github.com/HPE/terraform-provider-hpe/morpheus/testhelpers/sweep"
+	"github.com/HPE/terraform-provider-hpe/morpheus/utils/getsafe"
+)
+
+const sweeperName = "hpe_morpheus_instance_node"
+
+// TestResourcePrefix is used by acceptance tests for resource naming so
+// sweepers can identify and clean up test resources.
+const TestResourcePrefix = "TestAccMorpheusInstanceNode"
```
matthew-c-hpe: 🔵 **Recommendation**: The custom `TestResourcePrefix` (`TestAccMorpheusInstanceNode`) deviates from the standard pattern. Consider adding a comment explaining this deviation (e.g., preventing the sweeper from destroying parent instances).

matthew-c-hpe (changes requested):
---

## 📋 Review: `hpe_morpheus_instance_node`

**PR:** [#1324 — Add hpe_morpheus_instance_node resource (MORPH-13978)](https://github.com/HPE/terraform-provider-hpe/pull/1324)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 28 files · agents: patterns, spec-coverage, structure · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 3 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 1 |

**Verdict:** 🔴 **FAIL** — 4 issues found (3 critical)

---

### 📁 1. File Completeness

- 🔴 **Gap**: `schema_gen.go` is entirely missing from the `morpheus/framework/resources/instancenode` package. The schema and model types must be generated via the provider's `specs/` configuration and never hand-edited.
- 🔴 **Gap**: Missing paired data source. Every resource must have a corresponding data source with its own `datasource.go`, generated schema, and test coverage.
- 🔴 **Gap**: Missing registration for the data source in `morpheus/datasources.go`.

### ✨ 5. Code Quality

- 🔵 **Recommendation**: In `morpheus/framework/resources/instancenode/sweep/sweep.go`, the sweeper defines a custom `TestResourcePrefix` (`TestAccMorpheusInstanceNode`) rather than using the standard framework prefix directly. While likely intentional to prevent the sweeper from deleting parent instance test resources, consider adding a brief comment explaining this deviation.

---

morpheus/framework/resources/instancenode/sweep/sweep.go
matthew-c-hpe: Clarification, just remove this block and should be good to go, the sweeper code you implemented already uses the correct computed prefix from package `testsweep`

eamonnotoole: Done in 6709b6cd — comment added, and pushed.

You are right that it looked like an unexplained deviation, and it had already caused one defect this round, so the reasoning is now written next to the constant rather than left to be inferred:

> This sweeper does not delete resources it matches — it strips the extra containers from them — and its fixtures are whole instances created by other packages tests as well as this one. Matching the shared prefix would make it act on every test instance in the provider, removing nodes from fixtures it does not own while those tests are still running.

It closes with "Keep this narrower than the shared prefix", so the next person to see it does not helpfully align it with the standard.

eamonnotoole: You are right, and I had this backwards — thanks for the clarification, it caught a bug I introduced.

Reverted in b27bc5f8: the shared prefix is restored and the local constant is deleted.

The local prefix **never matched anything**. Fixture instances are named:

```go
name := fmt.Sprintf("TestAccMorpheus-%s-%s", t.Name(), rand.Text())
// e.g. TestAccMorpheus-TestAccMorpheusInstanceNodeVirtual-63IUUSIRCSCSXAOA4YTAEOTYSM
```

so `HasPrefix(name, "TestAccMorpheusInstanceNode")` fails on the hyphen straight after `TestAccMorpheus`:

```
HasPrefix(name, "TestAccMorpheus")             -> true
HasPrefix(name, "TestAccMorpheusInstanceNode") -> false
```

The sweeper had been selecting nothing since I changed it. Your original comment read the unused constant as the intent and the shared prefix as the bug; it was the other way round — the constant was dead code, which is exactly what you said here.

I made it worse in between: having "fixed" it, I then wrote a comment explaining why the narrower prefix was deliberate, which documented behaviour that was not happening. Both are gone.

I should have checked what the fixtures are actually named before changing the predicate — I had those names in front of me in acceptance output the whole time.

Repo-wide `make lint` clean, `go vet -tags sweep` clean, unit tests pass.

eamonnotoole: Adding a further reason the revert in b27bc5f8 is right, from PR #1302 which is already merged into `dev-2.0`.

That PR made the shared prefix overridable for targeted cleanup:

```go
// morpheus/testhelpers/sweep/sweep.go
const EnvSweepPrefix = "TF_ACC_SWEEP_PREFIX"
var TestResourcePrefix = resolveSweepPrefix(os.Getenv(EnvSweepPrefix))
```

It is a `var` resolved at startup, not a constant.

So a package-local `const TestResourcePrefix` was worse than I described. Beyond never matching the fixture names, it would have been **fixed at compile time and silently ignored `TF_ACC_SWEEP_PREFIX`** — defeating the override for this resource while every other sweeper honoured it. Anyone running a targeted sweep would have found instance-node fixtures skipped with no indication why.

I checked the rest of the provider: no other sweeper shadows the shared prefix with a local constant, so this package was the only one that would have diverged.

The code now consumes `testsweep.TestResourcePrefix` directly, so it matches the fixtures and picks up the override.

Thanks for pointing at #1302 — I had verified my revert against the fixture names but not against the override mechanism, and the two arguments are independent.

matthew-c-hpe (approved):

PR #1329, opened 2026-08-07
Add opt-in IP-address wait to hpe_morpheus_instance (MORPH-12804)

eamonnotoole:
Implements MORPH-12804. Design agreed in an internal plan.

Companion to #1324, which shipped `hpe_morpheus_instance_node` and, landing first, carried the shared readiness predicate and wait helper. This PR is the instance-side wiring.

## What it does

`wait_for_ip_address`, opt-in and defaulting to `false`, holds the apply until at least one container on the instance reports a usable address. Without it the address recorded in state can be the `0.0.0.0` placeholder Morpheus returns for a container that has not reported yet.

**At least one container, not all.** The instance owns the container it provisions; containers added later by `hpe_morpheus_instance_node` belong to that resource and are waited on there. Requiring all of them would make an instance apply block on nodes it does not own, and fail whenever any node never reports.

**On expiry it warns and continues**, never fails — the instance provisioned correctly and only the address is missing.

## Points for review

- **The attribute is provider-only and code-generated.** Declared as a `templates:` entry plus a `moves:` line in `specs/resources/instance/config.yaml`, then regenerated — `schema_gen.go` is not hand-edited. The regeneration touched 12 other `schema_gen.go` files, all reverted; the diff here is +8 lines in the instance schema only.
- **It never reaches a request.** It is referenced only by the wait, the read preservation, and the update-needed check, where changing it alone returns "no API call needed", mirroring how `timeouts` is handled.
- **`containerip` gains only `WaitAny`.** `Ready` and `Wait` are untouched, so #1324's behaviour is unchanged.

## Four drive-by fixes in the interface read path

Specified in the plan and fixed here because this work already touches the file. Three replace a panic with a skip or a null — a duplicate-name group where no row carried network information, a sub-interface with no matching top-level interface, and an unguarded dereference of a nil `Network`.

The fourth is the one to look at: diagnostics were being **overwritten rather than accumulated**, so each loop iteration discarded the previous one's. Fixing it can surface errors that were previously swallowed, which is why the existing instance acceptance tests were run rather than just the new one.

Note the plan claims the nil-`Network` dereference occurs at two sites. It does not — the first was already guarded, and only the second needed fixing.

## Verification

Against a live appliance:

- `TestAccMorpheusInstanceResourceWaitForIPAddress` — **passes**, and asserts the recorded address is **not** a placeholder. An earlier version of this test used `TestCheckResourceAttrSet`, which would have passed whether or not the wait did anything, since an unreported container returns a sentinel rather than nothing. That is the specific trap the plan warns about.
- Existing instance acceptance tests — `ExampleOk`, `UpdateName`, `UpdateInstanceContext`, `UpdateTags`, `UserGroupStorageProfile` — **all pass**, confirming the diagnostics fix surfaces nothing.

`make lint` clean repo-wide, `make docs` idempotent, appliance left as found.

## One correction carried from #82

The plan states that a node without "guest tools" never reports an address, and that positive test cases therefore need a layout with guest tools. Both are wrong — that is VMware terminology, the Morpheus concept is the agent, and neither governs the address, which comes from the platform. An instance with `noAgent: true` was observed carrying a real address. No such layout constraint is applied here.


morpheus/utils/containerip/containerip.go
```diff
@@ -29,6 +29,70 @@ func Ready(ip string) bool {
...
+		return struct{}{}, fmt.Errorf("instance %d: no container has a ready IP", instanceID)
+	}
+
+	_, err = backoff.Retry(
+		ctx,
+		poll,
+		backoff.WithBackOff(backoff.NewConstantBackOff(10*time.Second)),
+		backoff.WithMaxElapsedTime(timeout),
+	)
+	if err != nil {
```
matthew-c-hpe, line 86: 🔴 **Gap: WaitAny masks permanent API errors and renders caller checks dead code**

`WaitAny` unconditionally catches any `err` from `backoff.Retry`, logs it as a warning, and returns `true, nil`. This means `waitErr` in `instance_create.go` and `instance_update.go` will never be non-nil, making those `if waitErr != nil` checks dead code. If a permanent error occurs during the wait (e.g., a 403 or 500 API failure), it is incorrectly masked as an "IP address wait timed out" warning rather than returned. `WaitAny` must distinguish between a timeout (which can safely return `warned=true, err=nil`) and a permanent API error.

morpheus/framework/resources/instance/instance_update.go
```diff
@@ -827,6 +850,11 @@ func isAPIUpdateNeeded(plan, state InstanceModel) bool {
 		return true
 	}
 
+	// wait_for_ip_address — provider-only, no API call needed
+	if !plan.WaitForIpAddress.Equal(state.WaitForIpAddress) {
```
matthew-c-hpe: 🟡 **Warning: Dangerous short-circuit in `isAPIUpdateNeeded`**

Returning `false` immediately when `wait_for_ip_address` differs short-circuits the rest of the function. While currently functionally harmless because `timeouts` is the only remaining check and also returns `false`, it sets a trap where any future API field added below it would silently fail to trigger an update. Provider-only fields that don't trigger API calls should ideally be omitted from this function entirely.

morpheus/framework/resources/instance/instance_wait_ip_internal_test.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// TestUnitWaitForIpAddressDefaultFalse verifies that the schema default
+// for wait_for_ip_address is false.
+func TestUnitWaitForIpAddressDefaultFalse(t *testing.T) {
+	t.Parallel()
+
+	// A zero-value types.Bool is null, but when the schema default applies,
+	// it should be false. We verify by checking the default value directly.
+	defaultVal := types.BoolValue(false)
```
matthew-c-hpe: 🟡 **Warning: Useless test for schema default**

Constructing a `types.BoolValue(false)` and then asserting `.ValueBool() == false` does not test the resource schema's default value; it only tests the underlying boolean type behavior. To verify a schema default, inspect `InstanceResourceSchema(ctx)` directly.

matthew-c-hpe (changes requested):
---

**PR:** [#1329 — Add opt-in IP-address wait to hpe_morpheus_instance (MORPH-12804)](https://github.com/HPE/terraform-provider-hpe/pull/1329)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 9 files · agents: patterns, spec-coverage, structure · UI val: config-check · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 1 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 2 |
| 🔵 Recommendations | 0 |

**Verdict:** 🔴 **FAIL** — 3 issues found (1 critical)

---

### ⚙️ 4. Pattern Compliance

🔴 **WaitAny masks permanent API errors and renders caller checks dead code**
`WaitAny` unconditionally catches any `err` from `backoff.Retry`, logs it as a warning, and returns `true, nil`. This means `waitErr` in `instance_create.go` and `instance_update.go` will never be non-nil, making those error checks dead code. If a permanent error occurs during the wait (e.g., a 403 or 500 API failure), it is incorrectly masked as an "IP address wait timed out" warning rather than returned. `WaitAny` must distinguish between a timeout (which can safely return `warned=true, err=nil`) and a permanent API error.

🟡 **Dangerous short-circuit in `isAPIUpdateNeeded`**
Returning `false` immediately when `wait_for_ip_address` differs short-circuits the rest of the function. While currently functionally harmless because `timeouts` is the only remaining check and also returns `false`, it sets a trap where any future API field added below it would silently fail to trigger an update. Provider-only fields that don't trigger API calls should ideally be omitted from this function entirely.

🟡 **Useless test for schema default**
`TestUnitWaitForIpAddressDefaultFalse` constructs a `types.BoolValue(false)` and then asserts `.ValueBool() == false`. This does not test the resource schema's default value; it only tests the underlying boolean type behavior. To verify a schema default, inspect `InstanceResourceSchema(ctx)` directly.

---

🟡 **Sweeper Graceful Degradation**: The sweeper (`morpheus/framework/resources/instance/sweep/sweep.go`) does not utilize `testsweep.WithIgnoreListStatuses[T](http.StatusNotFound, http.StatusForbidden)` to handle missing resources gracefully.

*(+2 more recommendations available on request)*

---

morpheus/utils/containerip/containerip.go
```diff
@@ -29,6 +29,70 @@ func Ready(ip string) bool {
...
+		tflog.Warn(ctx, "IP address wait timed out; instance provisioned but address not yet available",
+			map[string]any{
+				"instance_id": instanceID,
+				"error":       err.Error(),
+			},
+		)
+
+		return true, nil
```
jasonquigleyhpecom, line 105: 🟠 **Violation**: `WaitAny` unconditionally catches all errors, creating dead code in the caller (`if waitErr != nil` blocks in `instance_create.go` and `instance_update.go` will never be reached).

If the wait encounters a permanent API error (e.g. 500 Internal Server Error, `failed to read instance`, or `response is nil`) or if the context is cancelled (`ctx.Err()`), it gets mislabeled as an "IP address wait timed out" warning, `waitErr` evaluates to `nil`, and the caller never knows to fail the apply or invoke `taintResourceState()`. 

**Fix**: `WaitAny` should differentiate between an actual timeout (which safely becomes a warning) and genuine API/context failures (which should be returned via `err` to fail the apply and correctly trigger the resource taint). Note that the pre-existing `Wait` helper below appears to suffer from the same pattern.

jasonquigleyhpecom (changes requested):

**PR:** [#1329 — Add opt-in IP-address wait to hpe_morpheus_instance (MORPH-12804)](https://github.com/HPE/terraform-provider-hpe/pull/1329)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 9 files · agents: patterns, spec-coverage, structure · UI val: config-check · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 1 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

**Verdict:** 🟠 **FAIL** — 1 issue found (1 critical)

---

### ⚙️ 4. Pattern Compliance

🟠 **Violation**: `WaitAny` unconditionally catches all errors, creating dead code in the caller.

In `containerip.go`, the helper `WaitAny` is documented and signed to return `(warned bool, err error)`. However, it unconditionally catches any error returned by `backoff.Retry()`, logs a warning, and returns `true, nil`.

Because `err` is never actually returned to the caller, the error-handling blocks in both `instance_create.go` and `instance_update.go` are completely unreachable (`if waitErr != nil { ... }`). If the wait encounters a permanent API error (e.g. 500 Internal Server Error, `failed to read instance`, or `response is nil`) or if the context is cancelled (`ctx.Err()`), it gets mislabeled as an "IP address wait timed out" warning, `waitErr` evaluates to `nil`, and the caller never knows to invoke `taintResourceState()`.

**Fix**: `WaitAny` should differentiate between an actual timeout (which safely becomes a warning) and genuine API/context failures (which should be returned via `err` to fail the apply and correctly trigger the resource taint). Note that the pre-existing `Wait` helper in the same file appears to suffer from the same pattern.

See inline comment for specifics.

---

**Next:** Fix 🟠 Violations before merge.

morpheus/utils/containerip/containerip.go
eamonnotoole: Confirmed and fixed in b0c09b7c. This was the most valuable finding of the round, and it was worse than the PR shows.

You are right on the mechanism: both helpers already wrap hard failures in `backoff.Permanent`, so those errors genuinely reach the caller — and then the handler logged every one of them as a timeout and returned nil. The `if waitErr != nil` checks were dead.

**The same defect is in `Wait`, which shipped in #1324.** There, `Create` calls `cleanup.TaintResourceState` on the wait-error path specifically to stop a created node being abandoned — a fix added at review request. Because `Wait` never returned an error, **that path was unreachable and the protection was inert.** Fixing the shared helper makes it live, which is why this PR now touches merged code.

Both functions now separate three outcomes: permanent error returned as an error, cancelled context returned as an error, genuine timeout warns and continues. Doc comments updated, since they previously claimed a timeout was the only failure mode.

Tests drive a real 500 from an `httptest` server and assert `err != nil` and `warned == false` — they fail against the old behaviour rather than merely passing against the new.

I also re-ran the instancenode acceptance tests, since this changes merged code out from under that resource. They pass.

morpheus/utils/containerip/containerip.go
eamonnotoole: Confirmed and fixed in b0c09b7c — you and Matthew found this independently.

Your note that `Wait` suffers the same pattern was the important part. It shipped in #1324, where `Create` taints on a wait failure to stop abandoning a created node. That path was **unreachable**, so the protection was inert. Fixed now.

Both helpers distinguish permanent error, cancelled context and genuine timeout; only the last warns and continues. Unit tests cover all three for both functions against a real HTTP error. I re-ran the instancenode acceptance tests because the change lands under merged code.

morpheus/framework/resources/instance/instance_update.go
eamonnotoole: Fixed in b0c09b7c. Fair point, and I had looked at this and talked myself out of it — I checked that it mirrored the `timeouts` handling below and concluded it was fine, which missed that the ordering is the fragile part rather than the pattern.

Restructured so provider-only fields cannot swallow later checks: they are excluded from the comparison entirely, as you suggested, rather than short-circuiting.

Worth noting deleting the check outright would have been wrong — the function defaults to `return true`, so a lone toggle change would then have caused a needless API call. The exclusion is written so it is safe in both directions, with a comment saying so.

morpheus/framework/resources/instance/instance_wait_ip_internal_test.go
eamonnotoole: Fixed in b0c09b7c — you are right that it tested the framework, not us.

Replaced with a test that obtains the resource schema, asserts `wait_for_ip_address` is optional and computed, and evaluates its `Default` to confirm it produces `false`.

matthew-c-hpe (commented):
---

**PR:** [#1329 — Implements MORPH-12804](https://github.com/HPE/terraform-provider-hpe/pull/1329)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 13 files · agents: patterns, spec-coverage, structure · UI val: config-check · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

🟡 **Warning: Non-standard file names**
* The CRUD operations in `morpheus/framework/resources/instance/` are named `instance_create.go`, `instance_read.go`, etc. The provider standard dictates these should simply be named `create.go`, `read.go`, `update.go`, `delete.go`, and `import.go`.

🔵 **Recommendation: Sweeper naming convention**
* In `sweep/sweep.go`, `unmanagedServerSweeperName` is set to `"hpe_morpheus_instance_unmanaged_server"`. While logical, the convention recommends that sweeper names exactly match a real Terraform resource name.

🟡 **Warning: API spec discrepancies (UI Validation)**
* A `config-check` validation of the API implementation revealed pre-existing spec gaps across all operations. For example, `CREATE` is missing ~30 fields in the spec (e.g., `uuid`, `server`, `createBackup`), while `UPDATE` misses `serverUUIDs` and `tags`. These are out of scope for this PR but indicate the API contract requires a refresh.

---

matthew-c-hpe (dismissed):

specs/resources/instance/config.yaml
```diff
@@ -50,11 +50,14 @@ instance:
           string: {}
         description: |-
           Optional UUIDs to assign to the servers provisioned for this instance.
-          Each UUID must be unique - Morpheus rejects a value already in use by
-          another server. Set at provision time only; changing it forces
+          Values are applied by position: the first UUID is assigned to the
```
jasonquigleyhpecom: 🟠 **Violation** — Schema type mismatch destroys user ordering. This description states values are applied by position, but `server_uuids` is defined as a `set`. Terraform sets are unordered; when `plan.ServerUuids.ElementsAs()` converts the set to a slice in `Create()`, the array is ordered arbitrarily by element hash, destroying the positional sequence. To support positional assignment, `server_uuids` must be changed from `set:` to `list:`.

morpheus/framework/resources/instance/instance_read.go
```diff
@@ -322,6 +322,8 @@ func getInstanceAsState(
 	} else {
 		state.ServerUuids = serverUUIDsFromContainerDetails(instance.ContainerDetails)
 	}
+	// wait_for_ip_address is provider-only; preserve from plan.
+	state.WaitForIpAddress = plan.WaitForIpAddress
```
jasonquigleyhpecom: 🟡 **Warning** — Spurious diff on import. When imported, `plan.WaitForIpAddress` is `Null`, so this direct assignment leaves the state `Null`. Because the attribute has a default (`false`), the first plan after import will apply the default against the `Null` state, resulting in a spurious diff (`null -> false`). Initialize to the default if null:

```go
if plan.WaitForIpAddress.IsNull() {
	state.WaitForIpAddress = types.BoolValue(false)
} else {
	state.WaitForIpAddress = plan.WaitForIpAddress
}
```

jasonquigleyhpecom (changes requested):
---

**PR:** [#1329 — Add opt-in IP-address wait to hpe_morpheus_instance (MORPH-12804)](https://github.com/HPE/terraform-provider-hpe/pull/1329)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 13 files · agents: patterns, spec-coverage, structure · UI val: config-check · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 1 |
| 🟡 Warnings | 1 |
| 🔵 Recommendations | 0 |

---

### 🟠 Violations

- **`server_uuids` schema type mismatch destroys user ordering**
  **File:** `specs/resources/instance/config.yaml:53`
  The PR modifies the description to state: *"Values are applied by position: the first UUID is assigned to the first server, and so on."* However, `server_uuids` remains configured as a `set`. Terraform sets are unordered; when `plan.ServerUuids.ElementsAs()` converts the set to a slice in `InstanceCreate`, the resulting array is ordered arbitrarily by element hash, destroying the user's intended positional sequence. To support positional assignment, `server_uuids` must be changed from `set:` to `list:` in the schema configuration.

### 🟡 Warnings

- **Spurious diff on import for `wait_for_ip_address`**
  **File:** `morpheus/framework/resources/instance/instance_read.go:326`
  When a resource is imported, `ImportState` only populates the `id`. During the subsequent `Read()` call, `plan.WaitForIpAddress` will be `.IsNull()`. The new read logic does a direct assignment (`state.WaitForIpAddress = plan.WaitForIpAddress`), leaving the state value as `Null`. Because the attribute has a schema default of `false`, the first `terraform plan` after import will apply the default against the `Null` state, resulting in a spurious diff (`null -> false`). To avoid this, initialize the value to its default if it's null.

- **Line length violation:** Untouched code at `morpheus/framework/resources/instance/instance_update.go:239` exceeds 120 characters.
- **OpenAPI Coverage Gaps:** The `instance` resource `config.yaml` uses an `includes:` whitelist, omitting several valid OpenAPI properties from the generated schema (e.g. `copies`, `metadata`, `accountId` mapping to `tenant_id`, etc.) that are not explicitly documented in the `removes:` block.

---

morpheus/framework/resources/instance/instance_read.go
eamonnotoole: Confirmed and fixed in eb8d6d9b. Good catch.

One correction to the route: `hpe_morpheus_instance` has no `ImportState`, so import is not actually possible for it. But the same thing happens by a **more likely** path — a provider upgrade. State written before this attribute existed carries no value for it, so the read copied the null through, the default supplied `false` at plan time, and the first plan after upgrading showed `null -> false` on an instance nobody had touched.

That makes it worse than an import-only edge case: every existing user of the resource would have seen it once.

Fixed exactly as you suggested, with a comment recording why, so it is not simplified back to a direct assignment later. Acceptance tests re-run against a live appliance since the read runs on every apply.

specs/resources/instance/config.yaml
eamonnotoole: You are right, and this is already tracked — **MORPH-15552**, raised before these reviews landed, covers changing the type.

Your mechanism is sharper than what I had written down, and I have used it. I had established that Morpheus assigns by index (`ServerService.groovy:623`, `serverUUIDs[index]`) and corrected the description to say so — but as you point out, saying "applied by position" while the attribute is a **set** promises something the provider cannot deliver, because `ElementsAs` yields hash order.

So the description was still misleading, just differently. It now says to supply at most one value and explains why: the API assigns by position, this is a set, sets are unordered, so with more than one value it is undefined which UUID reaches which server. That is honest about the current state rather than describing the API and leaving the reader to discover the gap.

Not changing `set` to `list` here. It shipped in v1.6.0, so it is a breaking change and belongs on a major version — MORPH-15552 records that `dev-2.0` is the cheapest opportunity, and the non-breaking fallback if that is not wanted.

Two other things on that ticket, from the same investigation:

- `hpe_morpheus_instance_node` cannot set a UUID, though the API allows it. **Verified on a live appliance** — a node added with `serverUUIDs` in the action envelope carried the requested value. Since node scaling is the only scaling path, that gap matters.
- Reading assigned UUIDs back is blocked on an ownership question: the list covers every container including ones added by the node resource, so reconciling naively would make `server_uuids` grow after scaling and, because it forces replacement, ask to destroy a working instance.

jasonquigleyhpecom (commented):
---

**PR:** [#1329 — Add opt-in IP-address wait to hpe_morpheus_instance (MORPH-12804)](https://github.com/HPE/terraform-provider-hpe/pull/1329)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 13 files · agents: patterns, spec-coverage, structure · UI val: config-check · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

The following API contract discrepancies exist in the provider configuration but were not introduced by this PR:

* **CREATE**:
  * **ONLY in contract**: `config`, `copies`, `evars`, `instance`, `metadata`, `networkInterfaces`, `ports`, `servicePlanOptions`, `taskSetId`, `taskSetName`, `volumes`, `zoneId`
  * **ONLY in implementation**: `ansibleGroup`, `ansibleRepoId`, `backup`, `backupRepository`, `backupTemplate`, `chefServerId`, `configEnabled`, `configId`, `createBackup`, `customOptions`, `environmentPrefix`, `expireDays`, `labelArray`, `lbInstances`, `loadBalancerDescription`, `loadBalancerId`, `loadBalancerName`, `loadBalancerProxyProtocol`, `loadBalancerScheme`, `loadBalancerSslCert`, `maxStorage`, `memoryDisplay`, `powerScheduleType`, `process`, `processId`, `providerBackupType`, `replicationGroup`, `server`, `shutdownDays`, `uuid`
*(+3 more operations with pre-existing discrepancies available on request)*

---

jasonquigleyhpecom (approved):

PR #1330, opened 2026-08-10
Replace server_uuids with a single server_uuid, and allow it on nodes (MORPH-15552)

eamonnotoole:
Implements MORPH-15552. Follows on from #1329, which shipped the create-time detection this builds on.

## Why

`server_uuids` shipped in v1.6.0 as a `Set of String`. Morpheus assigns UUIDs strictly by position — `ServerService.groovy:623`, `serverUUIDs[index]` — and an instance provisions exactly one server, since `layout_size` is always 1 and scaling is done with `hpe_morpheus_instance_node`.

So only the first value was ever used, the rest were discarded without a word, and a set discards the ordering the API depends on.

## What changed

**`server_uuid` (String) is added; `server_uuids` is deprecated, not removed.** Both work, and a config validator makes them mutually exclusive. `server_uuids` keeps its existing create and read behaviour unchanged — it is deprecated, so its semantics should not shift under anyone still using it. The deprecation reaches practitioners through the schema, which emits a warning naming the replacement.

**No changelog entry.** Release notes for this line should be written for the release as a whole rather than accumulated a PR at a time — doing it piecemeal here produced a wrong framing, listing an attribute added to `hpe_morpheus_instance_node` as an enhancement to an existing resource when the resource itself is new in this line. Whoever assembles the 2.0.0 notes will need to pick up, from this PR: the `server_uuids` deprecation, the new `container_id` on the instance, and the new `server_uuid` and renamed `container_uuid` on the node.

**`hpe_morpheus_instance` gains a computed `container_id`.** This is what makes reading the UUID back safe. Without it the resource could not distinguish its own server from one added by `hpe_morpheus_instance_node`, and reading all containers would have changed the value after scaling — and because the attribute forces replacement, asked Terraform to **destroy a working instance**. The read is scoped to the recorded container, with a regression test for that case.

**`hpe_morpheus_instance_node` gains `server_uuid`.** Verified against a live appliance before the attribute was written — a node added with `serverUUIDs` in the action envelope carried the requested value:

```
cid=10903  server=2447  uuid=e9fc6207-cb2d-408c-8709-84b7975d2415   (original)
cid=10904  server=2448  uuid=deadbeef-0000-4000-8000-1786115008000  (requested, applied)
```

It carries the same post-create verification as the instance, because the API silently ignores a UUID already in use on that path too.

**`hpe_morpheus_instance_node.uuid` is renamed `container_uuid`.** It holds the container's UUID and was unambiguous while it was the only UUID on the resource. With `server_uuid` beside it, a bare `uuid` gave no way to tell which object it identified, when every neighbouring attribute is explicit. The pairs are now symmetric:

```
container_id / container_uuid
server_id    / server_uuid
```

This attribute has not appeared in a release — it merged to `dev-2.0` in #1324 — so no published configuration is affected. Once this line ships, correcting it would be breaking.

## One thing worth reviewing closely

`server_uuid` is **Optional + Computed**, not Optional alone. My first attempt made it Optional-only and the acceptance suite caught it:

```
.server_uuid: was null, but now cty.StringVal("e3bc5121-...")
```

When unset, the planned value is null but the read supplies what Morpheus generated. `Computed` plus `UseStateForUnknown` is what `server_uuids` already had, and is needed here for the same reason.

## Verification

Against a live appliance, after every change including the rename: `ExampleOk` and `WaitForIPAddress` on the instance, and both `instancenode` acceptance tests. `make lint` clean repo-wide, docs regenerated, appliance left as found.

### DIFF
--- a/morpheus/framework/resources/instance/instance_read_internal_test.go
+++ b/morpheus/framework/resources/instance/instance_read_internal_test.go
@@ -521,3 +521,86 @@ func TestUnitContainerIdPreservedWhenAbsent(t *testing.T) {
 		t.Errorf("expected null server_uuid when container absent, got %q", uuid.ValueString())
 	}
 }
+
+// TestUnitServerUUIDsFromContainerDetails verifies that serverUUIDsFromContainerDetails
+// builds the correct set from containerDetails[].server.uuid.
+func TestUnitServerUUIDsFromContainerDetails(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name       string
+		containers []sdk.InstanceContainer2
+		wantNull   bool
+		wantUUIDs  []string
+	}{
+		{
+			name: "single container with UUID",
+			containers: []sdk.InstanceContainer2{
+				{Server: &sdk.InstanceContainerServer2{Uuid: sdk.PtrString("uuid-a")}},
+			},
+			wantUUIDs: []string{"uuid-a"},
+		},
+		{
+			name: "multiple containers with UUIDs",
+			containers: []sdk.InstanceContainer2{
+				{Server: &sdk.InstanceContainerServer2{Uuid: sdk.PtrString("uuid-a")}},
+				{Server: &sdk.InstanceContainerServer2{Uuid: sdk.PtrString("uuid-b")}},
+			},
+			wantUUIDs: []string{"uuid-a", "uuid-b"},
+		},
+		{
+			name: "container with no server",
+			containers: []sdk.InstanceContainer2{
+				{Server: nil},
+			},
+			wantNull: true,
+		},
+		{
+			name: "container with server but no UUID",
+			containers: []sdk.InstanceContainer2{
+				{Server: &sdk.InstanceContainerServer2{Uuid: nil}},
+			},
+			wantNull: true,
+		},
+		{
+			name:       "empty containers",
+			containers: nil,
+			wantNull:   true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			got := serverUUIDsFromContainerDetails(tt.containers)
+			if tt.wantNull {
+				if !got.IsNull() {
+					t.Errorf("expected null set, got %v", got)
+				}
+
+				return
+			}
+
+			ctx := context.Background()
+			var uuids []string
+			if d := got.ElementsAs(ctx, &uuids, false); d.HasError() {
+				t.Fatalf("ElementsAs failed: %v", d)
+			}
+
+			if len(uuids) != len(tt.wantUUIDs) {
+				t.Fatalf("got %d UUIDs, want %d", len(uuids), len(tt.wantUUIDs))
+			}
+
+			// Sets are unordered, so check that all expected UUIDs are present.
+			uuidSet := make(map[string]struct{}, len(uuids))
+			for _, u := range uuids {
+				uuidSet[u] = struct{}{}
+			}
+			for _, want := range tt.wantUUIDs {
+				if _, ok := uuidSet[want]; !ok {
+					t.Errorf("expected UUID %q not found in result %v", want, uuids)
+				}
+			}
+		})
+	}
+}
--- a/morpheus/framework/resources/instance/instance_read_internal_test.go
+++ b/morpheus/framework/resources/instance/instance_read_internal_test.go
@@ -4,12 +4,12 @@ package instance
 
 import (
 	"context"
-	"sort"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-framework/types"
 
 	sdk "github.com/HPE/terraform-provider-hpe/internal/sdk/oapigen"
+	"github.com/HPE/terraform-provider-hpe/utils/convert"
 )
 
 // TestUnitGetChildNetworksReadsSubnetID verifies that getChildNetworks reads subnet_id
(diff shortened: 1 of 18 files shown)


morpheus/framework/resources/instancenode/create.go
```diff
@@ -521,3 +554,30 @@ func extractContainerIDFromResults(results any, instanceID int64) (int64, error)
...
+// resolveNodeServerUUID reads the instance and returns the server UUID for
+// the container matching containerID. Returns empty string if not found.
+func resolveNodeServerUUID(
+	ctx context.Context,
+	client *sdk.APIClient,
+	instanceID int64,
+	containerID int64,
+) string {
+	getResp, _, err := client.InstancesAPI.GetInstance(ctx, instanceID).Execute()
+	if err != nil || getResp == nil || getResp.Instance == nil {
```
jasonquigleyhpecom: 🟠 **Violation: Swallowed API error breaks `errfmt` pattern**

The SDK call `client.InstancesAPI.GetInstance(ctx, instanceID).Execute()` correctly attempts to fetch the instance to validate the server UUID, but it completely swallows `err` by returning `""`. Masking the API error causes the caller (`Create`) to incorrectly flag a validation failure with a misleading error message (`... The server was assigned UUID "" instead.`). This diverges from the established error-handling pattern for UUID post-create validation seen in `hpe_morpheus_instance` (specifically `assignedServerUUIDs` in `instance_create.go`), which properly surfaces the API read failure via `errfmt.ErrMsg(err, hresp)`.

morpheus/framework/resources/instancenode/resource.go
```diff
@@ -211,7 +212,7 @@ func (r *Resource) Schema(
 				MarkdownDescription: "The name of the node container, " +
 					"assigned by the appliance.",
 			},
-			"uuid": schema.StringAttribute{
+			"container_uuid": schema.StringAttribute{
```
jasonquigleyhpecom, line 215: 🔴 **Gap: Removed committed schema attribute (backward-incompatible)**

The `uuid` schema attribute was removed and replaced with `container_uuid`. Removing an existing committed attribute key is a backward-compatibility break. (Note: The PR body mentions this attribute hasn't appeared in a release yet as it merged in #1324, but this strictly breaches the prohibition on removing committed schema attributes unless explicitly requested).

jasonquigleyhpecom (changes requested):
---

## 📋 Review: `hpe_morpheus_instance` / `hpe_morpheus_instance_node`

**PR:** [#1330 — Replace server_uuids with a single server_uuid, and allow it on nodes (MORPH-15552)](https://github.com/HPE/terraform-provider-hpe/pull/1330)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 16 files · agents: patterns, spec-coverage, structure · UI val: config-check · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 1 |
| 🟠 Violations | 1 |
| 🟡 Warnings | 1 |
| 🔵 Recommendations | 0 |

**Verdict:** 🔴 **FAIL** — 3 issues found (2 critical)

---

### ⚙️ 4. Pattern Compliance

### ✨ 5. Code Quality

🟡 **Warning: Missing parity in paired data source**
- **File:** `specs/datasources/instance/config.yaml`
- **Details:** The newly added computed attributes `server_uuid` and `container_id` in the `hpe_morpheus_instance` resource are missing from the paired data source. A user provisioning an instance with a `server_uuid` may expect to read it back via the data source.

- 🔴 **Gap:** `hpe_morpheus_instance_node` lacks a paired data source (`morpheus/framework/datasources/instancenode`). Data sources are equally important as resources and every resource must have a paired implementation.
- 🔴 **Gap:** `hpe_morpheus_instance_node` does not use code generation. Its schema is completely hand-written in `resource.go` and missing a `schema_gen.go` / `config.yaml` definition.
- 🔴 **Gap:** The `copies` and `taskSetName` properties from the API contract `instanceCreate` schema are missing from the Terraform schema (`schema_gen.go`) and CRUD logic for `hpe_morpheus_instance`.

---

morpheus/framework/resources/instancenode/resource.go
```diff
@@ -260,6 +261,26 @@ func (r *Resource) Schema(
 					"Compare with `resource_pool_id` (the placement requested at " +
 					"create time, bare-metal only) to detect drift.",
 			},
+			"server_uuid": schema.StringAttribute{
+				Optional: true,
```
matthew-c-hpe, line 265: 🔴 **Gap**: `server_uuid` is configured as `Optional: true` but is missing `Computed: true` (and `stringplanmodifier.UseStateForUnknown()`). Morpheus generates a server UUID if one isn't explicitly provided. By not marking it computed (and failing to read it back in `read.go`), Terraform loses track of the generated value and imports will leave the attribute permanently null. This breaks the pattern correctly established in `hpe_morpheus_instance`.

morpheus/framework/resources/instancenode/create.go
```diff
@@ -521,3 +554,30 @@ func extractContainerIDFromResults(results any, instanceID int64) (int64, error)
...
+		return ""
```
matthew-c-hpe: 🔴 **Gap**: Network errors are silently swallowed here. If the API read fails, this returns `""`, which causes the validation check in `Create` to evaluate `actualUUID != requestedUUID` as true. The provider then misreports this as "Morpheus silently ignored the requested server UUID" and taints the resource, hiding the true API/network failure. Return an error diagnostic instead.

matthew-c-hpe (dismissed):
## 📋 Review: `hpe_morpheus_instance` / `hpe_morpheus_instance_node`

**PR:** [#1330 — Replace server_uuids with a single server_uuid, and allow it on nodes (MORPH-15552)](https://github.com/HPE/terraform-provider-hpe/pull/1330)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 16 files · agents: patterns, spec-coverage, structure · UI val: config-check · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 2 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

*   🔴 **Gap**: Missing paired data source for `hpe_morpheus_instance_node`. Every resource must have a corresponding data source with full test coverage and registration.
*   🔴 **Gap**: Missing `schema_gen.go` for `hpe_morpheus_instance_node`. The schema appears to be entirely hand-written in `resource.go`.
*   🔵 **Recommendation**: The sweeper for `hpe_morpheus_instance_node` (`instancenode/sweep/sweep.go`) deletes test instances by invoking the API directly. This duplicates the `hpe_morpheus_instance` sweeper but lacks its robust logic (which stops and removes the underlying host directly to prevent unmanaged VMs from being left behind). Consider having the node sweeper defer to the parent instance sweeper.

---

morpheus/framework/resources/instancenode/create.go
eamonnotoole: Confirmed and fixed in f68909be. You and Matthew found this independently.

It was worse than an unhelpful message: a transient read failure caused the provider to conclude Morpheus had ignored the UUID and **taint the node**, destroying something that had provisioned correctly, while reporting the wrong cause.

Four outcomes were collapsing into `""` — API error, nil response, container absent, and a genuine mismatch. They are now distinct, and the read failure surfaces the underlying error via `errfmt.ErrMsg(err, hresp)` as the instance path already did.

While fixing it I found a second problem worth flagging. The helper had been split into a "pure-logic core" for testability that **nothing called** — the real path duplicated the same checks inline. So three of the new unit tests exercised a parallel copy while the executed code went uncovered, and the two were free to drift. There is now one implementation and the real path delegates to it, with a comment saying why it must stay that way.

morpheus/framework/resources/instancenode/create.go
eamonnotoole: Confirmed and fixed in f68909be — see the reply on Jason's thread for detail.

Your framing is the accurate one: the true failure was hidden and misreported. The three cases are now separate, and a read failure names the underlying error rather than being presented as a UUID mismatch.

morpheus/framework/resources/instancenode/resource.go
eamonnotoole: Confirmed and fixed in f68909be. You are right, and this is a mistake I had already made and fixed once on the instance side of this same PR without carrying the lesson across.

`server_uuid` on the node is now `Optional: true, Computed: true` with `UseStateForUnknown()` alongside `RequiresReplace()`, and the read populates it from the owned container. Morpheus generates a UUID when the practitioner does not supply one, and without this that value was never visible.

For the record on why `Computed` is load-bearing rather than cosmetic: with `Optional` alone and a read that populates the value, Terraform raises

```
.server_uuid: was null, but now cty.StringVal(...)
```

which is exactly what the acceptance suite threw at me on `hpe_morpheus_instance` earlier in this PR. Same defect, same fix, one resource later.

morpheus/framework/resources/instancenode/resource.go
eamonnotoole: I want to leave this one as it is, but the decision is yours rather than mine — say the word and I will revert it.

Your own note names the exception: *"unless explicitly requested"*. It was explicitly requested. The maintainer asked for the rename after I raised the ambiguity, so it is a deliberate maintainer decision rather than something I did in passing.

On the substance: the attribute has never appeared in a release. It merged to `dev-2.0` in #1324, and the latest tag is v1.6.0, so no published configuration can reference it. This is the last point at which the rename is free.

What it fixes is a genuine ambiguity that this PR creates. The resource now carries two UUIDs:

```
container_id / container_uuid   <- the container
server_id    / server_uuid      <- the underlying compute server
```

Before the rename that read `container_id`, `server_id`, `uuid`, `server_uuid` — where a bare `uuid` gives no way to tell which object it identifies, while every neighbouring attribute names its subject explicitly. Leaving it would mean shipping that ambiguity permanently, since correcting it after 2.0 would be the breaking change your rule is protecting against.

Both descriptions now state which object they identify, and the trade-off is recorded on the PR.

If the prohibition is meant to hold regardless of release status, I am happy to revert to `uuid` and raise the rename separately for 3.0 — it is a small change either way. I would rather you made that call than have me argue past a rule I did not write.

morpheus/framework/resources/instance/instance_read_internal_test.go
```diff
@@ -94,73 +94,75 @@ func TestUnitGetChildNetworksNoSubnet(t *testing.T) {
...
-				{Server: &sdk.InstanceContainerServer2{Uuid: nil}},
+				{Id: sdk.PtrInt64(200), Server: &sdk.InstanceContainerServer2{Uuid: sdk.PtrString("uuid-2")}},
 			},
-			wantNull: true,
+			containerID: types.Int64Value(999),
+			wantNull:    true,
+		},
+		{
+			name:        "null container_id returns null",
+			containers:  []sdk.InstanceContainer2{{Id: sdk.PtrInt64(1), Server: &sdk.InstanceContainerServer2{Uuid: sdk.PtrString("x")}}},
```
jasonquigleyhpecom, line 139: 🟡 **Warning**: Line is 129 characters long, which violates the strict 120-character maximum.

jasonquigleyhpecom (commented):
## 📋 Review: `hpe_morpheus_instance` & `hpe_morpheus_instance_node`

**PR:** [#1330 — Implements MORPH-15552](https://github.com/HPE/terraform-provider-hpe/pull/1330)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 17 files · agents: patterns, spec-coverage, structure · UI val: config-check · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 1 |
| 🔵 Recommendations | 0 |

---

*   **Missing carry-forward for computed attributes in `instancenode` Update:** In `morpheus/framework/resources/instancenode/update.go`, the computed fields `server_resource_pool_id`, `hostname`, `internal_ip`, `external_fqdn`, and `mac_address` lack `UseStateForUnknown()` plan modifiers in `resource.go` and are not explicitly carried forward from state to plan during `Update()`. Because they start as null in the plan, they will be wiped from the Terraform state if the resource is updated.
*   **API spec gaps (Instances):** UI validation for `instances` endpoints (`POST` and `PUT`) surfaces several undocumented attributes in the API contract (e.g., `server`, `ansibleGroup`, `backup`, `uuid` for CREATE, and `serverUUIDs` for UPDATE). These do not affect the PR's implementation but indicate the API contract is missing fields.

---

jasonquigleyhpecom (approved):

PR #1331, opened 2026-08-10
Add PCE Identity and PCE Disconnected Identity authentication (MORPH-15611)

matthew-c-hpe:
## Summary

Adds PCE (Private Cloud Enterprise) Identity support to the `morpheus` provider block, so
Morpheus connection details can be obtained from GreenLake rather than configured directly.

Two new mutually-exclusive blocks:

- **`pce_identity`** — Connected PCE, using GLCS IAM and scoped by GreenLake Space
- **`pce_disconnected_identity`** — Disconnected PCE, using GLP IAM and scoped by GreenLake Workspace

Each accepts either GreenLake API client credentials (`client_id`, `client_secret`,
`issuer_url`) or a pre-generated `iam_token`. The provider exchanges those for a Morpheus
URL and access token, so `url` and `access_token` are not set by hand.

## Authentication

The exchange has two legs: GreenLake API client credentials are traded for an IAM token,
which is then traded with the PCE broker for Morpheus connection details.

The two deployment types differ only in the IAM dialect and how the broker request is
scoped. Connected uses the HPE-hosted broker; Disconnected requires an operator-supplied
`broker_url`, as there is no hosted broker to fall back to.

## Schema

Both blocks are limited to a single element and cannot be combined with each other, nor
with `url`, `username`, `password`, `access_token` or `tenant_subdomain` — the exchange
supplies those, so configuring both is ambiguous. `iam_token` likewise conflicts with the
credentials it would otherwise be generated from.

`location` is required on both blocks, and `workspace_id` and `broker_url` on the
disconnected block.

## Notes for reviewers

- **The schemas are muxed.** The provider serves a framework provider and an SDKv2
  provider whose schema is derived from the framework one at start-up. The hand-written
  SDKv2 schema is what tests exercise, so the two are compared attribute-by-attribute in
  `TestHandWrittenSdkv2SchemaMatchesDerived`, including descriptions.
- **`MaxItems` is derived by reflection.** `utils/convert` reads the block's `SizeAtMost`
  validator by concrete type name, so wrapping or replacing it would silently drop the
  constraint. `TestIdentityBlocksDeriveMaxItems` pins this.
- **The exchange is not cached.** Both muxed providers perform their own exchange. Tokens
  already issued are not invalidated by issuing another, so they do not disturb each
  other's credentials.
- **The bundled SDKs are adapted from public HPE libraries** and keep their upstream
  identifier names, which is why `CMP` appears in code but not in documentation.

## Testing

- Provider configuration is validated end-to-end through `ValidateProviderConfig`,
  covering both wire representations of an unconfigured block
- The published examples are themselves validated, so a schema change cannot silently
  leave the documentation describing an invalid configuration
- Full suite passes; `make lint` is clean


morpheus/pce/sdk/token/serviceclient/handler.go
```diff
@@ -0,0 +1,253 @@
...
+// TokenChannels return channels for token retrieve function
+func (h *Handler) TokenChannels() (chan common.Result, chan int) { // nolint golint
+	return h.resultCh, h.exitCh
+}
+
+// startRetrieveThread start the token retrieve thread
+// function in an infinite loop, it puts the return value of retrieveToken into h.resultCh by default
+// if a signal on exitCh is received the thread exits
+func (h *Handler) startRetrieveThread() {
+	go func() {
```
matthew-c-hpe: 🔴 **Gap:** Goroutine Leak.

`NewHandler` spawns an infinite `startRetrieveThread` background goroutine that pushes to an unbuffered `h.resultCh`. The consumer in `retrieve.NewTokenRetrieveFunc` (called once per configuration in `exchange.go`) reads a single token and returns without sending a signal to `exitCh` on the success path. This leaves the background goroutine blocked forever on `h.resultCh <- h.retrieveToken()`, leaking a goroutine and making a redundant HTTP call on every provider configuration.

morpheus/pce/sdk/broker/client/api.go
```diff
@@ -0,0 +1,101 @@
...
+	localVarHTTPResponse, err := a.client.callAPI(req)
+	if err != nil || localVarHTTPResponse == nil {
+		return err
+	}
+	if localVarHTTPResponse.StatusCode >= 300 {
+		return ParseError(localVarHTTPResponse)
+	}
+
+	localVarBody, err := io.ReadAll(localVarHTTPResponse.Body)
+	defer localVarHTTPResponse.Body.Close()
```
matthew-c-hpe: 🟠 **Violation:** HTTP Response Body Leak.

If `localVarHTTPResponse.StatusCode >= 300`, the code early-returns `ParseError(localVarHTTPResponse)` on line 91. `ParseError` reads the body but does not close it. Because `defer localVarHTTPResponse.Body.Close()` is declared *after* this early return, the defer is never registered and the body is leaked. Move the defer to line 89, immediately after the nil/error check.

morpheus/pce/sdk/broker/client/api.go
```diff
@@ -0,0 +1,101 @@
...
+	// removeVmaasCMPBasePath is used to remove the base path of the vmaas-cmp API, for use by the broker API
+	removeVmaasCMPBasePath bool
+}
+
+// do will call the API provided. this function will not return any response, but
+// response should be catched from jsonParser function itself
+func (a *api) do(ctx context.Context, request interface{}, queryParams map[string]string) error {
+	currentVersion, err := parseVersion(a.compatibleVersion)
+	if err != nil {
+		panic(fmt.Sprintf("failed to parse the current version, error: %v", err))
```
matthew-c-hpe: 🟠 **Violation:** `panic()` in SDK methods.

Provider SDKs should never panic on configuration or validation errors, as this crashes the entire Terraform process. Return a standard Go `error` instead.

morpheus/pce/sdk/broker/client/api.go
```diff
@@ -0,0 +1,101 @@
...
+
+		return errVersion
+	}
+	var (
+		localVarHTTPMethod = strings.ToUpper(a.method)
+		localVarFileName   string
+		localVarFileBytes  []byte
+	)
+	if a.path == "" || a.method == "" || a.client == nil || a.jsonParser == nil {
+		panic("api not properly configured")
```
matthew-c-hpe: 🟠 **Violation:** `panic()` in SDK methods.

Provider SDKs should never panic on configuration or validation errors, as this crashes the entire Terraform process. Return a standard Go `error` instead.

morpheus/sdkv2/provider.go
```diff
@@ -280,25 +283,182 @@ func providerSchemaMorpheus() *schema.Schema {
...
+		return missingMorpheusBlock, nil
+	}
+
+	blocks, ok := morph.([]interface{})
+	if !ok || len(blocks) == 0 {
+		return missingMorpheusBlock, nil
 	}
 
-	morpheusConfig := morph.([]interface{})[0].(map[string]interface{})
+	// A morpheus block with no attributes set, such as one that only carries a
```
matthew-c-hpe: 🟡 **Warning:** Outdated/Contradictory Comment.

The comment claims a block carrying only `pce_identity` evaluates to a `nil` element. However, because `provider/adapter/provider.go` was updated in this PR to explicitly copy `Blocks`, the block will correctly evaluate to a map containing the nested block, making `ok` true. The comment should be removed or updated.

morpheus/provider.go
```diff
@@ -62,17 +66,72 @@ func (p *MorpheusProvider) Configure(ctx context.Context, req provider.Configure
...
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Failed to obtain Morpheus connection details from GreenLake",
+			err.Error(),
+		)
+
+		return
+	}
+
+	if url != "" {
```
matthew-c-hpe, line 96: 🔵 **Recommendation:** Late validation of missing connection details.

The Framework `Configure` logic resolves the URL via identity blocks, but if the user provides no `url` and no identity block, it silently creates a client with an empty URL. Adding a diagnostic check for an empty URL (similar to the SDKv2 provider) before creating the client will prevent opaque downstream errors.

matthew-c-hpe (commented):
---

## 📋 Review: `terraform-provider-hpe` Core Auth/PCE Updates

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 33 files · agents: review-patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 1 |
| 🟠 Violations | 3 |
| 🟡 Warnings | 1 |
| 🔵 Recommendations | 1 |

**Verdict:** 🔴 **FAIL** — 6 issues found (4 critical)

---

### 🔴 1. Gaps
* **`morpheus/pce/sdk/token/serviceclient/handler.go:150`** — Goroutine Leak. `NewHandler` spawns an infinite `startRetrieveThread` background goroutine that pushes to an unbuffered `h.resultCh`. The consumer in `retrieve.NewTokenRetrieveFunc` (called once per configuration in `exchange.go`) reads a single token and returns without sending a signal to `exitCh` on the success path. This leaves the background goroutine blocked forever on `h.resultCh <- h.retrieveToken()`, leaking a goroutine and making a redundant HTTP call on every provider configuration.

### 🟠 2. Violations
* **`morpheus/pce/sdk/broker/client/api.go:95`** — HTTP Response Body Leak. If `localVarHTTPResponse.StatusCode >= 300`, the code early-returns `ParseError(localVarHTTPResponse)` on line 91. `ParseError` reads the body but does not close it. Because `defer localVarHTTPResponse.Body.Close()` is declared *after* this early return, the defer is never registered and the body is leaked. Move the defer to line 89, immediately after the nil/error check.
* **`morpheus/pce/sdk/broker/client/api.go:37, 52`** — `panic()` in SDK methods. Provider SDKs should never panic on configuration or validation errors, as this crashes the entire Terraform process. Return a standard Go `error` instead.

### 🟡 3. Warnings
* **`morpheus/sdkv2/provider.go:456`** — Outdated/Contradictory Comment. The comment claims a block carrying only `pce_identity` evaluates to a `nil` element. However, because `provider/adapter/provider.go` was updated in this PR to explicitly copy `Blocks`, the block will correctly evaluate to a map containing the nested block, making `ok` true. The comment should be removed or updated.

### 🔵 4. Recommendations
* **`morpheus/provider.go:95`** — Late validation of missing connection details. The Framework `Configure` logic resolves the URL via identity blocks, but if the user provides no `url` and no identity block, it silently creates a client with an empty URL. Adding a diagnostic check for an empty URL (similar to the SDKv2 provider) before creating the client will prevent opaque downstream errors.

---

matthew-c-hpe (commented):
---

## 📋 Review: `hpe` provider

**PR:** [#1331 — Add PCE Identity and PCE Disconnected Identity authentication (MORPH-15611)](https://github.com/HPE/terraform-provider-hpe/pull/1331)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 19 files · agents: none (manual review) · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 1 |

---

---

morpheus/provider.go
```diff
@@ -116,5 +204,201 @@ func (p *MorpheusProvider) Schema(ctx context.Context, req provider.SchemaReques
...
+								`This should be set to the "Issuer" URL of the API client.`,
+							Optional: true,
+							Validators: []validator.String{
+								stringvalidator.AlsoRequires(
+									path.MatchRelative().AtParent().AtName("client_id"),
+									path.MatchRelative().AtParent().AtName("client_secret"),
+								),
+							},
+						},
+						"iam_token": schema.StringAttribute{
```
matthew-c-hpe, line 255: 🔵 **Recommendation**

Consider adding an early validation rule (such as `listvalidator.ExactlyOneOf` or the equivalent framework string validator) on `client_id` and `iam_token` across the block. Currently, it's possible to configure an empty `pce_identity` or `pce_disconnected_identity` block (since all internal fields are `Optional`), which will pass Terraform's early schema validation and only fail later during the provider `Configure` phase when the token exchange is attempted.

matthew-c-hpe: New custom Validator added for this purpose. The error that Terraform gives by using the AtLeastOneOf validator has very bad UX:


```
│ At least one attribute out of
│ [morpheus[0].pce_disconnected_identity[0].iam_token.<.client_id,morpheus[0].pce_disconnected_identity[0].iam_token.<.client_secret,morpheus[0].pce_disconnected_identity[0].iam_token.<.issuer_url,morpheus[0].pce_disconnected_identity[0].iam_token]
│ must be specified
```

morpheus/utils/validators/credentials_or_token.go
```diff
@@ -0,0 +1,83 @@
...
+		if !ok {
+			continue
+		}
+
+		// A configured value satisfies the rule. An unknown value is not null
+		// either, which defers the decision until Terraform has resolved it
+		// rather than reporting a configuration the user may well have
+		// supplied.
+		if !value.IsNull() {
+			return
```
matthew-c-hpe, line 73: 🟡 **Warning**: `IdentityCredentialsOrTokenValidator` uses `!value.IsNull()` to check if the user provided credentials. In the Plugin Framework, an explicit empty string (e.g., `client_id = ""`) evaluates to a non-null known value, meaning an empty string will falsely satisfy this validator. This would bypass the credential requirement during the plan phase and fail later with an opaque GreenLake API error during `Configure`.

Consider explicitly checking if the string is empty (`value.(types.String).ValueString() != ""` would work if type-asserted, or use `value.IsUnknown() || value.String() != ""` depending on attr), or adding `stringvalidator.LengthAtLeast(1)` to the credential attributes in the provider schema.

morpheus/sdkv2/provider.go
```diff
@@ -280,25 +283,184 @@ func providerSchemaMorpheus() *schema.Schema {
 						"default value is `false`",
 					Default: false,
 				},
+
+				// Mirrors the pce_identity block on the framework
+				// Morpheus provider. Both providers are muxed together and
+				// Terraform requires their schemas to be identical, so this
+				// must match what utils/convert produces from the framework
+				// schema, including descriptions.
+				"pce_identity": {
```
matthew-c-hpe, line 292: 🔵 **Recommendation**: The legacy SDKv2 schema for the `pce_identity` and `pce_disconnected_identity` blocks omit several constraint declarations that are present in their Framework counterparts, specifically:
- `ConflictsWith` for mutual exclusion between the identity blocks and direct connection details (`url`, `username`, `password`, `access_token`).
- `ConflictsWith` on `iam_token` (it should conflict with `client_id`, `client_secret`, and `issuer_url`).
- `RequiredWith` on `client_id`, `client_secret`, and `issuer_url` (mirroring the Framework's `AlsoRequires` validator).

While the Framework provider validation ultimately catches these issues because the providers are muxed, defining them explicitly on the SDKv2 hand-written schema keeps the test definitions fully constrained. The `compareAttrs` check in `TestHandWrittenSdkv2SchemaMatchesDerived` does not reflect on these fields, so adding them will not break parity tests.

matthew-c-hpe (commented):
---

## 📋 Review: `hpe_morpheus_provider`

**PR:** [#1331 — Add PCE Identity and PCE Disconnected Identity authentication (MORPH-15611)](https://github.com/HPE/terraform-provider-hpe/pull/1331)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 46 files · agents: patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 1 |
| 🔵 Recommendations | 1 |

**Verdict:** 🟡 **PASS WITH WARNINGS** — 2 issues found (0 critical)

---

### ⚙️ 1. Pattern Compliance

---

**Next:** 🟡 Warnings are advisory. 🔵 Recommendations are optional.

morpheus/provider.go
matthew-c-hpe: Fixed in 9ceb77977

morpheus/utils/validators/credentials_or_token.go
matthew-c-hpe: client_id = "" is known and non-null, which satisfies typical framework expected behaviour.

We don't have the same protections in place for the `morpheus` block's direct attributes so we'll allow an empty string unless we discover this causes issues.

morpheus/sdkv2/provider.go
matthew-c-hpe: Added the `ConflictsWith` declarations to the hand-written SDKv2 schema:

- **Block level** — both identity blocks now conflict with `url`, `username`, `password`, `access_token` and `tenant_subdomain`. I included `tenant_subdomain`, which wasn't in the list above, because the Framework block carries it and fidelity is the point of the change.
- **`iam_token`** — conflicts with `client_id`, `client_secret` and `issuer_url` within its own block, fully qualified as `morpheus.0.<block>.0.<attr>`.

The mutual conflict between the two blocks is declared on `pce_identity` only, mirroring the Framework, so configuring both is reported once rather than once per block.

`RequiredWith` is not included, keeping this change to the conflict rules.

You're right that parity is unaffected — `compareAttrs` doesn't reflect on `ConflictsWith`. Path correctness is covered by the existing `TestHandWrittenSdkv2SchemaIsInternallyValid`: it runs `InternalValidate`, which is SDKv2's own path checker. I confirmed it bites by mutating a path (`references unknown attribute`) and an index (`can only use the .0. index for TypeList and MaxItems: 1`).

Two things worth recording for whoever reads this later:

1. **These constraints don't execute in production.** `main.go` replaces `sdkv2Provider.Schema` wholesale with the Framework-derived map, and `FwToSdkv2SchemaMap` translates only `MinItems`/`MaxItems` — there is no validator translation. So this is fidelity for the hand-written schema used by tests, not runtime behaviour.
2. **Their runtime behaviour is untested.** The acceptance tests build the Framework provider alone, never the muxed SDKv2 server, so nothing exercises SDKv2 schema validation at runtime. `InternalValidate` proves the paths resolve, not that they fire correctly. That matters because SDKv2 evaluates `ConflictsWith` via `c.Get()` on the raw `ResourceConfig`, and `access_token`, `username` and `password` carry `Default: ""` — the pre-existing `ConflictsWith` entries share that exposure today.

jasonquigleyhpecom (commented):
---

## 📋 Review: `hpe_morpheus_provider`

**PR:** [#1331 — Add PCE Identity and PCE Disconnected Identity authentication (MORPH-15611)](https://github.com/HPE/terraform-provider-hpe/pull/1331)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 46 files · agents: patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

morpheus/provider.go
```diff
@@ -62,17 +67,101 @@ func (p *MorpheusProvider) Configure(ctx context.Context, req provider.Configure
...
+	if url != "" {
+		m.URL = types.StringValue(url)
+		m.AccessToken = types.StringValue(token)
+	}
+
+	// "url" is optional in the schema because an identity block can supply it,
+	// so nothing has rejected a configuration that sets neither. Report it here
+	// rather than building a client that cannot reach anything and failing
+	// later with a less obvious error.
+	if m.URL.ValueString() == "" {
```
eamonnotoole, line 105: 🔴 **Gap: Missing `IsUnknown()` guards in Framework Provider `Configure`**

The provider reads configuration values and performs actions before checking if they are unknown. When dynamic values (e.g., `url = var.morpheus_url` or `client_secret = var.secret`) are used, they are evaluated as Unknown during the plan phase.

- Calling `m.URL.ValueString()` on an unknown value returns `""`. This will falsely trigger the "Missing Morpheus connection details" diagnostic and block `terraform plan`.
- Similarly, `pceIdentityTokenExchange` (line 82) is invoked directly with `.ValueString()` properties. If any fields are unknown, empty strings are sent in the blocking HTTP request, causing the exchange to fail and generating an error diagnostic that also halts the plan.

**Recommendation:** Add `IsUnknown()`/`HasUnknown()` checks for the identity block and `m.URL` properties. The standard Framework pattern is to return early from `Configure` (without error) when provider configuration fields required for client construction are unknown, allowing resources to handle `nil` provider data during the plan.

morpheus/sdkv2/provider.go
```diff
@@ -280,25 +283,212 @@ func providerSchemaMorpheus() *schema.Schema {
...
+						"morpheus.0.pce_disconnected_identity",
+						"morpheus.0.url",
+						"morpheus.0.username",
+						"morpheus.0.password",
+						"morpheus.0.access_token",
+						"morpheus.0.tenant_subdomain",
+					},
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"client_id": {
```
eamonnotoole, line 310: 🟠 **Violation: Framework vs SDKv2 Schema Inconsistency (`RequiredWith`)**

The Framework schema uses `stringvalidator.AlsoRequires` to enforce that `client_id`, `client_secret`, and `issuer_url` must be provided together. However, the SDKv2 schemas for `pce_identity` and `pce_disconnected_identity` are missing the equivalent `RequiredWith` constraints for these attributes. Terraform expects muxed schemas to be conceptually identical; omitting this creates a validation gap for configurations evaluating through the SDKv2 provider paths.

**Recommendation:** Mirror the framework logic by adding `RequiredWith` arrays to `client_id`, `client_secret`, and `issuer_url` in the SDKv2 schema. For example:
```go
RequiredWith: []string{
    "morpheus.0.pce_identity.0.client_secret",
    "morpheus.0.pce_identity.0.issuer_url",
},
```

eamonnotoole (commented):
---

## 📋 Review: `provider`

**PR:** [#1331 — Add PCE Identity and PCE Disconnected Identity authentication (MORPH-15611)](https://github.com/HPE/terraform-provider-hpe/pull/1331)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 15 files · agents: patterns, structure · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 1 |
| 🟠 Violations | 1 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

### ⚙️ 4. Pattern Compliance

🔵 **Recommendation: Provider Configure Diagnostics in SDKv2**
- **File:** `morpheus/sdkv2/provider.go`
- **Details:** When the legacy provider lacks connection details, it returns string messages (`missingMorpheusBlock`, `incompleteMorpheusBlock`) as the `interface{}` client, relying on a downstream type-assertion panic (`meta.(*client.Client)`) in individual resources to surface the message to the user. While this preserves legacy behavior, standard SDKv2 practice is to emit these directly as `diag.Diagnostics` errors from `providerConfigure` so Terraform can fail gracefully.

---

morpheus/provider.go
matthew-c-hpe: This is a hallucination, when using variables terraform will correctly parse a var that has been set. Of course, a var that hasn't ben set will compute to "" with `ValueString()`.

morpheus/sdkv2/provider.go
matthew-c-hpe: Added in fdc0f02

morpheus/provider.go
eamonnotoole, line 105: 🔴 **Gap: Missing `IsUnknown()` guards in Framework Provider `Configure`**

(Still present from previous review) The provider reads configuration values and performs actions before checking if they are unknown. When dynamic values (e.g., `url = var.morpheus_url` or `client_secret = var.secret`) are used, they are evaluated as Unknown during the plan phase.

- Calling `m.URL.ValueString()` on an unknown value returns `""`. This will falsely trigger the "Missing Morpheus connection details" diagnostic and block `terraform plan`.
- Similarly, `pceIdentityTokenExchange` (line 82) is invoked directly with `.ValueString()` properties. If any fields are unknown, empty strings are sent in the blocking HTTP request, causing the exchange to fail and generating an error diagnostic that also halts the plan.

**Recommendation:** Add `IsUnknown()`/`HasUnknown()` checks for the identity block and `m.URL` properties. The standard Framework pattern is to return early from `Configure` (without error) when provider configuration fields required for client construction are unknown, allowing resources to handle `nil` provider data during the plan.

eamonnotoole (commented):
---

## 📋 Review: `provider`

**PR:** [#1331 — Add PCE Identity and PCE Disconnected Identity authentication (MORPH-15611)](https://github.com/HPE/terraform-provider-hpe/pull/1331)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 15 files · agents: patterns, structure · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 1 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

**Verdict:** 🔴 **FAIL** — 1 issue found (1 critical)

---

### ⚙️ 4. Pattern Compliance

🔵 **Recommendation: Provider Configure Diagnostics in SDKv2**
- **File:** `morpheus/sdkv2/provider.go`
- **Details:** When the legacy provider lacks connection details, it returns string messages (`missingMorpheusBlock`, `incompleteMorpheusBlock`) as the `interface{}` client, relying on a downstream type-assertion panic (`meta.(*client.Client)`) in individual resources to surface the message to the user. While this preserves legacy behavior, standard SDKv2 practice is to emit these directly as `diag.Diagnostics` errors from `providerConfigure` so Terraform can fail gracefully.

---

eamonnotoole (approved):

PR #1332, opened 2026-08-10
Add affinity group resources and data sources

eamonnotoole:
Adds cloud- and cluster-scoped affinity group support, so VM placement rules can be managed as code on VMware clouds and HVM clusters.

## What's added

**Resources**
- `hpe_morpheus_cloud_affinity_group` — new, for VMware, VMware Cloud on AWS and MacStadium clouds
- `hpe_morpheus_cluster_affinity_group` — reworked, for HVM clusters. The shipped resource could not set an affinity type and could not manage membership at all

**Data sources**
- `hpe_morpheus_cloud_affinity_group` / `_groups`, `hpe_morpheus_cluster_affinity_groups` (singular cluster already existed, now exposes `servers`, `source`, `tenant_ids` and `resource_permissions`)
- `hpe_morpheus_compute_server` / `_servers` — the plural filterable by cloud, instance and bare-metal flag

**On `hpe_morpheus_instance`**
- A computed `compute_servers` attribute exposing the underlying compute server IDs

## How the pieces fit

Membership is a set of compute server IDs and is readable **only** from the affinity group side, so it is owned by the affinity group resource rather than the instance:

```hcl
resource "hpe_morpheus_cloud_affinity_group" "spread" {
  cloud_id      = 2
  pool_id       = 1
  name          = "web-spread"
  affinity_type = "KEEP_SEPARATE"
  servers       = hpe_morpheus_instance.web.compute_servers
}
```

Provisioning-time placement is a separate mechanism and is tracked under MORPH-15596.

## Behaviours worth reviewing

Each of these fails silently if got wrong, and each is commented at the call site:

- **`servers` is wholesale-replace.** Omitting the key or sending an empty array removes every member, so an update must always send the complete desired set. `resolveUpdateServers` falls back to the recorded state when the planned value is unknown — without it, renaming a group would wipe its membership. Unit-tested in both packages.
- **`active` must always be sent on create**, otherwise the group is created inactive.
- **`affinity_type` and `pool_id` are create-only** and marked `RequiresReplace`; the API ignores them on update.
- **Groups imported from vCenter DRS rules are refused.** A cloud sync rewrites their type, active flag and membership every cycle, so Terraform would fight it indefinitely. The data sources return them instead.
- **`servers` deliberately has no `UseStateForUnknown`.** Membership changes out of band, and pinning a stale set risks "inconsistent result after apply". The resulting plan noise is the lesser evil.
- **`pool_id` is effectively required** on the cloud resource — Morpheus rejects a create without it.

## Breaking change

`description` is removed from `hpe_morpheus_cluster_affinity_group`. It does not exist on the API object, appears in no request or response schema, and was never read or sent by the provider — any value set was silently discarded.

## Testing

All eight packages pass acceptance tests against a 9.0.2 appliance, covering create/read/update/delete, membership add/remove/replace, import verification and `RequiresReplace`, on both a VMware cloud and an HVM cluster.

Tests are gated on a new `affinity_group` capability alongside `vmware` / `hvm`, and take their infrastructure IDs from environment variables rather than hardcoded values.

## Related PRs

- Companion API spec fix: https://github.com/HewlettPackard/morpheus-openapi/pull/318

That PR corrects the server interface `subnet`, `networkGroup`, `networkPool` and `networkDomain` types, which are objects rather than strings. Without it the compute server list data source cannot decode a response containing an interface with a subnet. **This PR should not merge until that one has.**

The bundled SDK update in this branch is generated from that fix.


morpheus/framework/resources/cloudaffinitygroup/resource.go
```diff
@@ -0,0 +1,971 @@
...
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"strconv"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
```
matthew-c-hpe: 🔴 **Gap**: `morpheus/framework/resources/cloudaffinitygroup/` is missing `create.go`, `read.go`, `update.go`, `delete.go`, and `import.go`. The CRUD operations are currently bundled within `resource.go` but must be split into their own separate files to match provider structure conventions.

morpheus/framework/resources/cloudaffinitygroup/resource_test.go
```diff
@@ -0,0 +1,232 @@
+package cloudaffinitygroup_test
```
matthew-c-hpe, line 3: 🟠 **Violation**: Missing copyright header. All new files must include the standard HPE copyright header.

specs/datasources/cloud_affinity_group/config.yaml
```diff
@@ -0,0 +1,77 @@
+cloud_affinity_group:
```
matthew-c-hpe, line 1: 🔴 **Gap**: Resource vs. Data Source Schema Inconsistency. The resource flattens the `pool` object into a scalar `pool_id` (via `template-pool: pool_id`), but the data sources (singular and plural) do not. This exposes a nested `pool { id }` block here instead. Add a `pool` template (`int64`) and `template-pool: pool_id` move to both data source configs to ensure users can pass data source outputs directly into resource inputs.

specs/datasources/compute_server/config.yaml
```diff
@@ -0,0 +1,165 @@
...
+      description: The power state of the compute server (e.g. on, off).
+    status:
+      description: The status of the compute server (e.g. provisioned, running).
+    uuid:
+      description: The UUID of the compute server.
+    visibility:
+      description: The visibility of the compute server (public or private).
+    agent_installed:
+      description: Whether the Morpheus agent is installed on this compute server.
+    managed:
```
matthew-c-hpe: 🟡 **Warning**: Orphaned Override. The `overrides:` section contains an entry for `managed`. However, the OpenAPI `server` model does not have a top-level `managed` property (only `computeServerType.managed`, which is handled via `compute_server_type_managed`). This override is silently ignored by `go-split`.

matthew-c-hpe (commented):
---

## 📋 Review: `hpe_morpheus_cloud_affinity_group` / `hpe_morpheus_compute_server`

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 96 files · agents: patterns, spec-coverage, structure · UI val: full · connected: [#318](https://github.com/HewlettPackard/morpheus-openapi/pull/318)

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 2 |
| 🟠 Violations | 1 |
| 🟡 Warnings | 2 |
| 🔵 Recommendations | 0 |

**Verdict:** 🔴 **FAIL** — 5 issues found (3 critical)

### 🔗 3. UI Validation
🟡 **Warning**: For `compute_server(s)` (mapped to `servers` API): The API implementation returns several properties not documented in the API contract (e.g., `accountName`, `appliance`, `maintenanceMode`, `managed`, `type`, `uefi`), which could affect data source attribute coverage.
🟡 **Warning**: For `compute_server(s)`: The API contract documents properties `parentServer` and `secureMetadataDatastore`, but the API implementation does not return them.

- 🔴 **Gap**: `morpheus/framework/resources/clusteraffinitygroup/` is missing `create.go`, `read.go`, `update.go`, `delete.go`, and `import.go` splits. (This resource was modified, but the lack of file split is pre-existing).
- 🟠 **Violation**: Missing copyright headers in `morpheus/framework/resources/clusteraffinitygroup/resource_test.go`, `morpheus/testhelpers/config.go`, and `morpheus/framework/resources/storagebucket/resource.go`.
- 🟡 **Warning**: Line length exceeds 120 characters in the pre-existing `example.go` files.

- The generated SDK changes mapping to network interfaces (e.g., `model_list_hosts_200_response_all_of_servers_inner_interfaces_inner.go`) match the reference type corrections introduced by the connected companion PR, successfully syncing the provider SDK with the updated API contract.

---

morpheus/framework/resources/cloudaffinitygroup/resource.go
```diff
@@ -0,0 +1,971 @@
+// (C) Copyright 2026 Hewlett Packard Enterprise Development LP
```
jasonquigleyhpecom, line 1: 🔴 Gap — This resource package is missing separate `create.go`, `read.go`, `update.go`, `delete.go`, and `import.go` files. The CRUD methods are monolithic in `resource.go`.

morpheus/framework/resources/clusteraffinitygroup/resource.go
```diff
@@ -18,11 +18,18 @@ import (
 	sdk "github.com/HPE/terraform-provider-hpe/internal/sdk/oapigen"
 
 	"github.com/HPE/terraform-provider-hpe/morpheus/configure"
+	"github.com/HPE/terraform-provider-hpe/morpheus/utils/constants"
```
jasonquigleyhpecom: 🔴 Gap — This resource package is missing separate `create.go`, `read.go`, `update.go`, `delete.go`, and `import.go` files. The CRUD methods remain monolithic.

specs/resources/cluster_affinity_group/config.yaml
```diff
@@ -1,77 +1,92 @@
...
-    - source
-    - success
-    - tenants
-    - tenant_permissions
+    - ref_type                                         # (components/schemas/clusterAffinityGroup.yaml) read-only
+    - ref_id                                           # (components/schemas/clusterAffinityGroup.yaml) read-only
+    - external_id                                      # (components/schemas/clusterAffinityGroup.yaml) read-only
+    - tenants                                          # (components/schemas/clusterAffinityGroup.yaml) replaced by tenant_ids
+    - pool                                             # (components/schemas/clusterAffinityGroup.yaml) replaced by pool_id
+    - description                                      # EXPLICIT REMOVAL: does not exist on the domain class (v2.0.0 breaking change)
```
jasonquigleyhpecom: 🔴 Gap — The `description` attribute is explicitly removed here. Removing a previously-committed attribute from a Terraform provider constitutes a backward-incompatible breaking change for any practitioners currently passing this attribute.

specs/datasources/compute_server/config.yaml
jasonquigleyhpecom: 🟠 Violation — This override points to a non-existent root attribute in the schema. The `managed` boolean exists on the nested `computeServerType` object and is extracted as `compute_server_type_managed`. This override should target `compute_server_type_managed` instead.

morpheus/framework/resources/cloudaffinitygroup/resource_test.go
jasonquigleyhpecom, line 3: 🟠 Violation — Missing the standard copyright header: `// (C) Copyright 2026 Hewlett Packard Enterprise Development LP`

morpheus/framework/resources/clusteraffinitygroup/resource_test.go
```diff
@@ -22,18 +22,19 @@ func TestMain(m *testing.M) {
 	os.Exit(code)
 }
 
+// TestAccMorpheusClusterAffinityGroupResourceExampleOk tests create, read, and import.
```
jasonquigleyhpecom, line 27: 🟠 Violation — Missing the standard copyright header: `// (C) Copyright 2026 Hewlett Packard Enterprise Development LP`

morpheus/testhelpers/config.go
```diff
@@ -1,12 +1,55 @@
 package testhelpers
 
-import "os"
+import (
```
jasonquigleyhpecom, line 5: 🟠 Violation — Missing the standard copyright header: `// (C) Copyright 2026 Hewlett Packard Enterprise Development LP`

jasonquigleyhpecom (changes requested):
---

## 📋 Review: `hpe_morpheus_cloud_affinity_group`, `hpe_morpheus_cluster_affinity_group`, `compute_server(s)`

**PR:** [#1332 — Add affinity group resources and data sources](https://github.com/HPE/terraform-provider-hpe/pull/1332)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 94 files · agents: patterns, spec-coverage, structure · UI val: config-check · connected: [#318](https://github.com/HewlettPackard/morpheus-openapi/pull/318)

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 3 |
| 🟠 Violations | 2 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

**Verdict:** 🔴 **FAIL** — 5 issues found (3 critical)

---

### 🔴 Gaps

*   **Missing standard CRUD file split:** `morpheus/framework/resources/cloudaffinitygroup/resource.go` is implemented as a monolith instead of splitting operations into `create.go`, `read.go`, `update.go`, `delete.go`, and `import.go`.
*   **Missing standard CRUD file split:** `morpheus/framework/resources/clusteraffinitygroup/resource.go` is implemented as a monolith (same missing files as above).
*   **Removed Schema Attribute (Backward-Incompatible Break):** The `description` attribute has been completely removed from the resource schema in `specs/resources/cluster_affinity_group/config.yaml` (`schema_gen.go`). Removing a previously-committed attribute is a breaking change for existing practitioners.

### 🟠 Violations

*   **Dangling Override (Spec-to-Code):** In `specs/datasources/compute_server/config.yaml`, the `managed` override points to a non-existent root attribute in the spec/schema. It should target `compute_server_type_managed` as defined by the template flattener.
*   **Missing copyright headers:** The standard copyright header is missing in several newly added or rewritten files (`morpheus/framework/resources/cloudaffinitygroup/resource_test.go`, `morpheus/framework/resources/clusteraffinitygroup/resource_test.go`, `morpheus/testhelpers/config.go`).

*   **UI Validation (`zones-affinity-groups` & `clusters-affinity-groups`):** The API implementation relies on `tenantPermissions` for Create/Update, but it is missing from the API contract (spec gap). Fields `affinityType`, `name`, `servers`, `pool` are in the API contract but unused by the API implementation.
*   **UI Validation (`servers`):** The API implementation returns numerous fields on GET that are absent from the API contract (e.g. `managed`, `accountName`, `hardwareCpuModel`), and the contract lists possible dead fields (e.g. `parentServer`). 

---

morpheus/framework/resources/cloudaffinitygroup/resource.go
eamonnotoole: Fixed in `75ddc276`. Split into `create.go` / `read.go` / `update.go` / `delete.go` / `import.go`, following `networkrouterbgpneighbor`. `resource.go` is now 50 lines — `NewResource`, `Metadata`, `Schema`.

This was a pure move with no logic changes; the unit tests and the full acceptance suite pass unmodified, which is the evidence for that.

morpheus/framework/resources/cloudaffinitygroup/resource_test.go
eamonnotoole: Fixed in `75ddc276`.

specs/datasources/compute_server/config.yaml
eamonnotoole: Fixed in `75ddc276`. Retargeted to `compute_server_type_managed`, which is what the template flattener actually produces from the nested `computeServerType.managed`.

Good catch — the plural data source (`specs/datasources/compute_servers/config.yaml`) had the identical dangling override, so that is fixed too. A sweep of both files confirms every remaining override now targets an attribute that exists.

morpheus/framework/resources/cloudaffinitygroup/resource.go
eamonnotoole: Fixed in `75ddc276`. Split into `create.go` / `read.go` / `update.go` / `delete.go` / `import.go`, following `networkrouterbgpneighbor`. `resource.go` is now 50 lines — `NewResource`, `Metadata`, `Schema`.

This was a pure move with no logic changes; the unit tests and the full acceptance suite pass unmodified, which is the evidence for that.

morpheus/framework/resources/clusteraffinitygroup/resource.go
eamonnotoole: Fixed in `75ddc276`. Split into `create.go` / `read.go` / `update.go` / `delete.go` / `import.go`, following `networkrouterbgpneighbor`. `resource.go` is now 50 lines — `NewResource`, `Metadata`, `Schema`.

This was a pure move with no logic changes; the unit tests and the full acceptance suite pass unmodified, which is the evidence for that.

specs/datasources/compute_server/config.yaml
eamonnotoole: Fixed in `75ddc276`. Retargeted to `compute_server_type_managed`, which is what the template flattener actually produces from the nested `computeServerType.managed`.

Good catch — the plural data source (`specs/datasources/compute_servers/config.yaml`) had the identical dangling override, so that is fixed too. A sweep of both files confirms every remaining override now targets an attribute that exists.

morpheus/framework/resources/cloudaffinitygroup/resource_test.go
eamonnotoole: Fixed in `75ddc276`.

morpheus/framework/resources/clusteraffinitygroup/resource_test.go
eamonnotoole: Fixed in `75ddc276`.

morpheus/testhelpers/config.go
eamonnotoole: Fixed in `75ddc276`.

specs/resources/cluster_affinity_group/config.yaml
eamonnotoole: Fixed in `75ddc276` — deprecated rather than removed. You are right that removing a shipped attribute is a backward-incompatible break, and the repo already has precedent for the deprecation route: `vdi_pool.max_session_timeout` is deprecated for exactly this reason ("not backed by the Morpheus API and has no effect"), and `server_uuids` was deprecated rather than removed in this same release.

It is now `Optional` only, never `Computed`, so it can never be marked unknown, and it stays inert in CRUD — never sent, never read. A configured value simply persists in state, and practitioners get a warning explaining why instead of a hard "Unsupported argument" failure.

For the record on why it was flagged in the first place: `description` appears in no affinity group request or response schema, is not on the domain object, and the provider never sent or read it — any value set was silently discarded. So the attribute is genuinely inert; the disagreement was only about how to retire it, and your way is better.

specs/datasources/cloud_affinity_group/config.yaml
eamonnotoole: This one is deliberate rather than an oversight, so I would like to make the case before changing it.

The resource takes `pool_id = 1` because that is what a practitioner writes, and it sits naturally beside `cloud_id` and `cluster_id` in the same schema — a single-field nested object was noticeably clumsier to author. The data sources keep `pool = {id, name}` because on the read side the name is genuinely useful, and flattening to a scalar would discard it.

Write by id, read the richer object, is a reasonably common Terraform shape. The API itself is asymmetric in the same direction: it accepts `pool: {id: N}` on write but returns `pool: {id: N, name: "..."}` on read, so the provider is following the contract rather than inventing the split.

That said, it is a judgement call and you may weigh cross-resource consistency higher than I have. If you would rather they matched, I would suggest flattening the data sources to `pool_id` plus a computed `pool_name` rather than nesting the resource — that keeps the name available and makes both sides flat. Happy to do that if you prefer it.

jasonquigleyhpecom (commented):
---

## 📋 Review: `hpe_morpheus_*_affinity_group` & `hpe_morpheus_compute_server`

**PR:** [#1332 — Add affinity group resources and data sources](https://github.com/HPE/terraform-provider-hpe/pull/1332)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 86 files · agents: patterns, spec-coverage, structure · UI val: config-check · connected: [#318](https://github.com/HewlettPackard/morpheus-openapi/pull/318)

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

- **UI Validation (API vs Spec):** The API implementation returns `tenantPermissions` for both cloud and cluster affinity groups, but this property is missing from the API contract (`zoneAffinityGroup.yaml` / `clusterAffinityGroup.yaml`).

- The API implementation returns `id` and `name` properties under `subnet`, `networkGroup`, `networkPool`, and `networkDomain` in interfaces, which are correctly omitted from the provider diff because they are addressed in the connected PR.

jasonquigleyhpecom (approved):

PR #1335, opened 2026-08-12
Add instance affinity group placement and cluster/host visibility (MORPH-15596)

eamonnotoole:
## Summary

Adds provision-time affinity group placement for instances, plus the visibility
needed to confirm that placement actually happened.

### `affinity_group_id` (MORPH-15596)

`config_vmware.affinity_group_id` and `config_hvm.affinity_group_id` place an
instance into an affinity group when it is provisioned. Create-only: changing
the value replaces the instance, because the API accepts it at provision time
only.

It is rejected alongside `config_hvm.kvm_host_id`. Morpheus applies the pinned
host and then lets the affinity group override it, discarding the pin without
warning, so accepting both would silently ignore one of them.

### `parent_host_id` / `parent_host_name`

`hpe_morpheus_compute_server` and `hpe_morpheus_compute_servers` now report the
hypervisor host a guest runs on. The plural data source can filter on it.

This closes a real gap: an instance echoes `affinity_group_id` back unchanged
whether or not placement succeeded, so the only way to verify placement was to
query the API by hand. It can now be asserted from HCL.

### `hpe_morpheus_clusters`

New plural data source. Server-side filtering by `cloud_id` and `type_id`;
client-side `filter` block over `name`, `status`, `type_code`, `type_name`,
`cloud_id`, `cloud_name` and `dynamic_placement_mode`.

`dynamic_placement_mode` is the motivating case: affinity rules are only
enforced on clusters that have it enabled, and it was previously reachable only
by string-poking an untyped `config` map on the singular data source.

`type_code` is resolved from the cluster types endpoint, because the cluster
response embeds only `{id, name}` for its type. It is worth the extra call:
codes are stable identifiers whereas display names are not — the type whose code
is `mvm-cluster` is displayed as "HVM".

## Verification

Exercised end to end against a 9.0.2 appliance with real VMware and HVM
infrastructure, not just unit tested.

- Acceptance tests: 3 new (VMware placement, HVM placement, RequiresReplace)
  plus 4 validator unit tests — all pass.
- `hpe_morpheus_clusters`: 3 acceptance tests, pass.
- VMware placement confirmed by host comparison: seed and instance landed on
  different hosts under `KEEP_SEPARATE`.
- HVM `KEEP_TOGETHER` confirmed by pinning the seed to a host the scheduler
  would not otherwise have chosen and observing the new instance follow it —
  which rules out the scheduler's own preference as an explanation.

## Behaviour worth knowing

Two platform behaviours are documented on the resource, because both are easy to
misread as provider faults.

**VMware requires the group to have an existing member.** A vSphere DRS rule
needs two virtual machines, so provisioning the first machine into an empty
group builds a one-machine rule that vCenter rejects. Provisioning fails at
power on, and the resulting instance cannot be removed with `terraform destroy`
because it never started.

**HVM applies `KEEP_SEPARATE` late.** Only `KEEP_TOGETHER` reliably affects
placement at creation. A `KEEP_SEPARATE` violation is resolved afterwards by the
cluster's placement loop, which live-migrates the instance two to six minutes
later. Reading placement immediately after apply can therefore look like a
failure when it is only a convergence delay.

## Notes for review

Squashed into a single commit covering three related changes: the affinity
attribute, the host attributes, and the clusters data source. The latter two
exist to make the first verifiable.

No spec change was required for `parent_host_*` or `hpe_morpheus_clusters` — the
underlying fields and operations were already present.

## Related PRs

- Companion API spec PR: https://github.com/HewlettPackard/morpheus-openapi/pull/324 (merged)
- Follow-up API spec PR, read-side fields: https://github.com/HewlettPackard/morpheus-openapi/pull/327

This PR merges first: the SDK is generated and vendored in-tree, so it is
self-contained and does not depend on the spec PR merging.


morpheus/framework/resources/instance/instance_create.go
```diff
@@ -172,6 +172,14 @@ func (g *Resource) Create(
 			configHvm.KvmHostId = plan.ConfigHvm.KvmHostId.ValueInt64Pointer()
 		}
 
+		if !plan.ConfigHvm.AffinityGroupId.IsNull() {
```
matthew-c-hpe: 🟠 **Violation**: Add `&& !plan.ConfigHvm.IsUnknown()` to ensure safe retrieval before calling `.ValueInt64Pointer()`.

morpheus/framework/resources/instance/instance_create.go
```diff
@@ -195,6 +203,10 @@ func (g *Resource) Create(
 			configVMware.CreateUser = *sdk.NewNullableBool(&createUser)
 		}
 
+		if !plan.ConfigVmware.AffinityGroupId.IsNull() {
```
matthew-c-hpe: 🟠 **Violation**: Add `&& !plan.ConfigVmware.IsUnknown()` to ensure safe retrieval before calling `.ValueInt64Pointer()`.

morpheus/framework/datasources/clusters/datasource.go
```diff
@@ -0,0 +1,401 @@
...
+// identifiers while display names are not: the type whose code is
+// "mvm-cluster" is displayed as "HVM".
+func clusterTypeCodes(
+	ctx context.Context,
+	apiClient *sdk.APIClient,
+) (map[int64]string, error) {
+	out := map[int64]string{}
+
+	types, _, err := apiClient.ClustersAPI.ListClusterTypes(ctx).Execute()
+	if err != nil {
```
matthew-c-hpe, line 304: 🟠 **Violation**: The SDK call must capture the HTTP response to format the error correctly: `types, hresp, err := apiClient...`. Then return `providererrors.ErrMsg(err, hresp)`.

matthew-c-hpe (commented):
---

## 📋 Review: `hpe_morpheus_instance` / `hpe_morpheus_clusters`

**PR:** [#1335 — Add instance affinity group placement and cluster/host visibility (MORPH-15596)](https://github.com/HPE/terraform-provider-hpe/pull/1335)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 47 files · agents: patterns, spec-coverage, structure · UI val: full · connected: [#324](https://github.com/HewlettPackard/morpheus-openapi/pull/324)

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 1 |
| 🟠 Violations | 2 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

**Verdict:** 🔴 **FAIL** — 3 issues found (3 critical)

---

### 🔍 2. Attribute Coverage

🔴 **Gap: Missing read-back for `affinity_group_id` (HVM and VMware)**
In `morpheus/framework/resources/instance/instance_read.go`, the `getInstanceHVMConfig()` and `getInstanceVMwareConfig()` helper functions construct the typed config object from the API response but fail to initialize the `AffinityGroupId` attribute. Because these struct fields are zero-initialized, their state defaults to `Unknown` (0). Returning an unknown value from a `Read` operation causes Terraform to error out (`Provider produced inconsistent result after apply`).
**Fix:** Explicitly initialize the field to `types.Int64Null()` alongside the other fields, as it is a write-only property not returned by the API contract. (Note: This file is not currently in the PR diff).

### ⚙️ 4. Pattern Compliance

🟠 **Violation: Missing `IsUnknown()` null guards in create logic**
When reading `affinity_group_id` from the plan during instance creation, the code only checks if the field `!IsNull()`. The system convention requires `!IsNull() && !IsUnknown()` to ensure safe retrieval before calling `.ValueInt64Pointer()`.
*See inline comments for specifics.*

🟠 **Violation: Missing `errfmt` on API call**
The helper function `clusterTypeCodes()` makes an SDK call but ignores the HTTP response object and returns the bare error. This breaks the proven pattern for error handling: it must capture the HTTP response and return `providererrors.ErrMsg(err, hresp)` to ensure the detailed API error payload is included in the Terraform diagnostic.
*See inline comments for specifics.*

**UI Validation: `clusters` (List)**
- The API implementation returns several properties omitted from the API contract (e.g., `autoScalerEnabled`, `autoScalerGroups`, `clusterIdentifier`, `computeServerType`, `nodeType`, `serverGroupMemberStatus`, `stats`, `typeSet`, `username`).
- The API contract defines properties not seen in the API implementation (possible dead entries: `containersCount`, `deploymentsCount`, `jobsCount`, `namespacesCount`, `podsCount`, `servicesCount`, `volumesCount`, `workersCount`).

**UI Validation: `servers` (Read / List)**
- The API implementation returns properties omitted from the API contract (e.g., `accountName`, `appliance`, `code`, `displayName`, `externalDelete`, `format`, `guestAgentStatus`, `hardwareCpuFrequency`, `hardwareCpuModel`, `hardwareProductName`, `hardwareProductVendor`, `hugepages`, `iScsiInitiatorName`, `maintenanceMode`, `managed`, `managedBy`, `maxSockets`, `minimumLogLevel`, `placementStrategy`, `secureBoot`, `serverGroupMemberStatus`, `systemServer`, `tpm`, `type`, `uefi`).

- **`affinity_group_id` / `affinityGroup`**: These fields are missing from the baseline API contract for HVM and VMware configurations. Since they are included in the connected companion PR's added set, their presence in the provider schema and SDK is expected and safely consumed by the provider.

---

morpheus/framework/resources/instance/instance_create.go
eamonnotoole: Fixed in f34a23e5 — now guards on both `IsNull()` and `IsUnknown()` before
dereferencing.

morpheus/framework/resources/instance/instance_create.go
eamonnotoole: Fixed in f34a23e5 — same guard applied on the VMware path.

morpheus/framework/datasources/clusters/datasource.go
eamonnotoole: Fixed in f34a23e5. The call now captures the HTTP response and returns
`providererrors.ErrMsg(err, hresp)`, so the API error payload reaches the
diagnostic. Thanks — this one would have made a cluster-types failure hard to
diagnose in the field.

specs/datasources/compute_server/config.yaml
```diff
@@ -36,6 +36,21 @@ compute_server:
       string:
         computed_optional_required: computed
         description: The name of the service plan.
+    parent_host_id:
```
matthew-c-hpe, line 39: 🟡 **Warning** — The PR maps the API's `parentServer.id` and `parentServer.name` to these new data source fields. However, API contract validation testing indicates that the `parentServer` object is currently missing from the actual implementation response (`ONLY in spec (possible dead)`). As a result, these newly flattened fields may evaluate to `null` at runtime.

matthew-c-hpe (commented):
---

## 📋 Review: `hpe_morpheus_instance` (and `clusters`, `compute_server`, `compute_servers`)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 40 files · agents: review-patterns, review-spec-coverage, review-structure · UI val: full for instance, config-check for data sources · connected: [#324](https://github.com/HewlettPackard/morpheus-openapi/pull/324)

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 1 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 1 |
| 🔵 Recommendations | 0 |

---

### 🔴 1. Gaps

*   **`morpheus/framework/resources/instance/instance_read.go`** (Not in diff)
    The newly added attributes `config_hvm.affinity_group_id` and `config_vmware.affinity_group_id` are sent to the API during create (`instance_create.go`), but they are not mapped back into the Terraform state in `instance_read.go` when parsing the API response `apiConfig`. Since they were added to the `instance` resource schema but are missing from the `getInstanceHVMConfig` and `getInstanceVmwareConfig` read paths, this will cause spurious drift when the resource is refreshed or imported.

*   **`affinityGroup` / `affinityGroupId` Spec Sync**
    The new `affinityGroup` and `affinityGroupId` fields implemented for HVM and VMWare configurations map correctly to the API extensions being added in the connected PR [#324](https://github.com/HewlettPackard/morpheus-openapi/pull/324). No spec-behind violation exists.

---

specs/datasources/compute_server/config.yaml
eamonnotoole: Checked this against a live 9.0.2.94 appliance — `parentServer` is returned, so
the fields do populate.

Single read:

```
GET /api/servers/571819
  parentServer present: True
  parentServer = {"id": 561857, "name": "hvm-c1-b1"}
```

And through the data sources themselves:

```
guest_parent_host_id   = 561857
guest_parent_host_name = "hvm-c1-b1"
guests_on_b1           = ["mc-feat-hvm-ubuntu-1 -> hvm-c1-b1"]
```

The second line is the plural data source filtering on `parent_host_name`, so
the flattening works on the list path too.

I think I can explain the validation result. On the list endpoint the field is
populated for guests but null for hosts, because a hypervisor host has no
parent:

```
autof-mvm-resize RBD 0723   parentServer=hvm-c1-b2
hvm-c1-b1                   parentServer=None
hvm-c1-b2                   parentServer=None
```

If the sample it inspected happened to be host records, `parentServer` would be
absent from every one of them and it would look like a dead spec entry. That
null is correct and intentional, and both attribute descriptions say so — "or
null for a host itself".

Happy to add an acceptance assertion pinning a known guest to its host if you
would like it covered explicitly, though it would depend on that guest staying
put, which is exactly the kind of external dependency that broke the VMware test
in this PR.

specs/datasources/compute_server/config.yaml
```diff
@@ -51,6 +66,8 @@ compute_server:
     - template-plan_id: plan_id                                                # (template) flatten plan.id
     - template-plan_code: plan_code                                            # (template) flatten plan.code
     - template-plan_name: plan_name                                            # (template) flatten plan.name
+    - template-parent_host_id: parent_host_id                                  # (template) flatten parentServer.id
```
matthew-c-hpe, line 69: 🟡 **Warning:** The PR maps `parent_host_id` and `parent_host_name` from the `parentServer` API contract property. However, API validation flags `parentServer` as documented but possibly missing from the implementation. Please verify whether the API actually returns this object during READ/LIST operations; if it doesn't, these mapped fields will always be null.

matthew-c-hpe (commented):
---

## 📋 Review: `hpe_morpheus_instance` (and `clusters`, `compute_servers` data sources)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 50 files · agents: patterns, structure, spec-coverage · UI val: config-check · connected: [#324](https://github.com/HewlettPackard/morpheus-openapi/pull/324)

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 1 |
| 🔵 Recommendations | 0 |

---

### 🟡 3. Warnings

- **`specs/datasources/compute_server/config.yaml`**: The PR maps `parent_host_id` and `parent_host_name` from the `parentServer` spec property. However, UI validation (config-check depth) flags `parentServer` as `ONLY in spec (possible dead)`. Please verify whether the API implementation actually returns this object during READ/LIST operations; if it doesn't, these data source fields will always be null. See inline comments for specifics.

**API Validation — `clusters` (LIST)**
- Documented in API contract, missing from implementation: `containersCount`, `deploymentsCount`, `jobsCount`, `namespacesCount`, `podsCount`, `servicesCount`, `volumesCount`, `workersCount`
- Returned by API implementation, missing from API contract: `autoScalerEnabled`, `autoScalerGroups`, `clusterIdentifier`, `computeServerType`, `nodeType`, `serverGroupMemberStatus`, `stats`, `typeSet`, `username`

**API Validation — `servers` (READ/LIST)**
- Returned by API implementation, missing from API contract: `accountName`, `appliance`, `code`, `displayName`, `externalDelete`, `format`, `guestAgentStatus`, `hardwareCpuFrequency`, `hardwareCpuModel`, `hardwareProductName`, `hardwareProductVendor`, `hugepages`, `iScsiInitiatorName`, `maintenanceMode`, `managed`, `managedBy`, `maxSockets`, `minimumLogLevel`, `placementStrategy`, `secureBoot`, `serverGroupMemberStatus`, `systemServer`, `tpm`, `type`, `uefi`

- **`internal/sdk/oapigen/model_*.go`**: The SDK files were updated with `affinityGroup` and `affinityGroupId`. The spec-coverage agent initially flagged these as hand-edited since they were missing from the base spec, but they have been correctly added by the connected companion PR [#324](https://github.com/HewlettPackard/morpheus-openapi/pull/324).

---

specs/datasources/compute_server/config.yaml
eamonnotoole: Same point as the thread on `parent_host_id` above — verified against a live
9.0.2.94 appliance, and `parentServer` is returned on both read and list.

```
GET /api/servers/571819
  parentServer = {"id": 561857, "name": "hvm-c1-b1"}
```

`parent_host_name` specifically, read through the data source:

```
guest_parent_host_name = "hvm-c1-b1"
guests_on_b1           = ["mc-feat-hvm-ubuntu-1 -> hvm-c1-b1"]
```

The second line is the plural data source filtering on `parent_host_name`, so it
populates on the list path too.

The likely source of the validation result is that the field is null for
hypervisor hosts, which have no parent, and populated only for guests. A sample
consisting of host records would show it absent throughout. That null is
deliberate and both descriptions state it — "or null for a host itself".

matthew-c-hpe (commented):
---

## 📋 Review: `hpe_morpheus_instance`, `clusters`, `compute_server(s)`

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 50 files · agents: patterns, spec-coverage, structure · UI val: full · connected: [#324](https://github.com/HewlettPackard/morpheus-openapi/pull/324)

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

- **API Contract Discrepancies (`instances`)**: The full-depth API contract validation pass found pre-existing discrepancies in the `instances` endpoint. Several properties are defined in the spec but absent in the API implementation, and numerous implementation-level properties are missing from the spec. These do not intersect with the changes in this PR.
- **API Contract Discrepancies (`clusters`)**: The API contract validation pass for the `clusters` endpoint highlighted a few pre-existing discrepancies (e.g. `autoScalerEnabled` present in implementation but absent in the spec). These are out of scope for the newly added data source.

- The newly consumed `affinityGroup` and `affinityGroupId` fields in the `instance` resource are fully supported by the provider's vendored SDK (and the underlying OpenAPI contract) via the connected PR [#324](https://github.com/HewlettPackard/morpheus-openapi/pull/324).

morpheus/framework/datasources/clusters/datasource.go
```diff
@@ -0,0 +1,402 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	config.Clusters = setVal
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &config)...)
+}
+
+func compileFilters(
```
matthew-c-hpe, line 159: Should we consider moving this out to a separate package, this seems useful for all plural datasources that we implement in framework.

matthew-c-hpe (approved):

morpheus/framework/datasources/clusters/datasource.go
eamonnotoole: Probably/

PR #1337, opened 2026-08-13
Manage affinity group membership one member at a time (MORPH-15728)

eamonnotoole:
## Summary

Affinity group membership can now be managed one member at a time, so Terraform
stops evicting members it did not create.

Two new resources: `hpe_morpheus_cloud_affinity_group_member` and
`hpe_morpheus_cluster_affinity_group_member`.

## The problem

`servers` on the affinity group resources was authoritative: it owned the whole
membership set, so anything added by other means was drift and was removed on
the next apply.

That is not limited to changes made outside Terraform. Two supported provider
paths add members the group resource knows nothing about:

- an instance provisioned with `config_vmware.affinity_group_id` or
  `config_hvm.affinity_group_id`
- a node added with `hpe_morpheus_instance_node`

So a practitioner using only Terraform, never touching the UI, could lose
membership silently. The end-to-end examples needed a
`lifecycle { ignore_changes = [servers] }` block to avoid it, which was the
signal that the model was wrong rather than the examples.

## What changed

**New membership resources.** Create, Read and Delete only — a membership either
exists or it does not, so every attribute forces replacement. State records the
pairing, never the group's full list: a copy of the list in each membership
would go stale the moment any other member changed, reintroducing the same
problem at N times the scale. Import takes
`<parent>_id/affinity_group_id/server_id`.

**`servers` is no longer configurable** on the group resources. It remains as a
read-only report of current membership.

**Update re-reads membership live.** This is the part worth reviewing closely.
The API has no encoding for "leave membership alone" — omitting `servers` and
sending an empty array both empty the group:

```
group created with servers = [570236]   -> members [570236]
update sending only description         -> members []
```

So every update must send the full list even when only the name is changing.
Taking that list from Terraform state was not safe: state only records what the
resource last saw, so a member added since then was evicted by an unrelated
update. It is now read live, immediately before the write.

**Shared locking and read-back.** Because membership can only be written as a
whole, the group resources and the membership resources share a per-group lock
around their read-modify-write cycles. The lock covers one provider process,
which is where Terraform runs every resource in an apply; it cannot cover two
concurrent applies, so each write is followed by a read-back assertion that
reports a concurrent writer rather than silently losing to it.

## Also in this change

**Terminology.** A compute server may be a guest virtual machine *or* a
hypervisor host, which is not obvious from names like `compute_server_id` or
`servers`, and has caused repeated confusion when reading placement results.
Every page exposing those attributes now states the model: an instance is what
you provision, a compute server is a machine record backing it, a host is a
compute server that hosts others.

**Shared filter block.** Raised in review of #1335: the plural data source
filter implementation was duplicated across six data sources, around 340 lines
differing in two places. Compilation and matching now live in a shared package,
with the generated block type and the field accessor supplied by the caller. The
`clusters` data source is migrated as the first user (-78/+43); the other five
are unchanged, since each needs its own verification and batching them would
obscure the one migration that has been tested.

## Verification

Exercised against a live 9.0.2 appliance.

- Four new acceptance tests, all passing. Each resource gets a lifecycle test
  with import, and a coexistence test that adds a second server to the group
  behind Terraform's back, re-applies, and asserts it survived. That second test
  is the behaviour these resources exist to provide, and it fails against the
  previous approach.
- The update path verified by hand: with a member added out of band, renaming
  the group through Terraform left both members intact. Under the previous
  state-based behaviour the rename evicted the out-of-band member.
- Existing affinity group and clusters acceptance tests still pass.

The tests deliberately use existing compute servers rather than provisioning
instances. Membership does not require a new instance, and provisioning is the
slowest and least reliable part of the environment — tying these tests to it
would make them fail for unrelated reasons.

## Notes for review

**No membership sweeper.** A membership carries no name, tags or metadata, so a
leaked one cannot be told apart from a legitimate one and a sweeper could remove
real membership. The tests create their own affinity group instead, so anything
left behind is reclaimed by the existing group sweepers.

**Two resources rather than one.** Cloud and cluster affinity groups are
distinct API surfaces with distinct request and response types. A single
resource would need mutually exclusive attributes and runtime dispatch to hide
that, so this mirrors the existing split.

## Breaking changes on the 2.0 line

Neither affects a released version, but both affect anyone tracking the
development line, and both belong in the release notes.

**`servers` becomes read-only.** Configurations that set it will break. Replace
`servers = [...]` with one membership resource per entry.

**`resource_permissions.groups.id` becomes required** on the cluster resource,
matching the cloud one. Configurations that list a group without giving its id
will break. In practice nothing can have been working: the id defaulted to `0`,
which is not a group, so the permission it described was never the one intended.

MORPH-15728


morpheus/framework/resources/cloudaffinitygroupmember/delete.go
```diff
@@ -0,0 +1,164 @@
...
+				"attribute at the same time as this resource.",
+		)
+	}
+}
+
+// Update exists only to satisfy the interface.
+//
+// Every attribute forces replacement, so Terraform never calls this: a
+// membership cannot be changed, only created or destroyed.
+func (r *Resource) Update(
```
matthew-c-hpe: 🟠 **Violation:** This file houses the `Update()` and `ImportState()` methods. These must be placed in their own `update.go` and `import.go` files, even if the resource is force-replace only and `Update()` merely returns an error.

morpheus/framework/resources/clusteraffinitygroupmember/delete.go
```diff
@@ -0,0 +1,164 @@
...
+				"attribute at the same time as this resource.",
+		)
+	}
+}
+
+// Update exists only to satisfy the interface.
+//
+// Every attribute forces replacement, so Terraform never calls this: a
+// membership cannot be changed, only created or destroyed.
+func (r *Resource) Update(
```
matthew-c-hpe: 🟠 **Violation:** Similarly, this file houses `Update()` and `ImportState()`. Please relocate them to `update.go` and `import.go`.

morpheus/framework/resources/cloudaffinitygroupmember/create.go
```diff
@@ -0,0 +1,152 @@
...
+
+// writeMembership replaces the group's server list.
+func writeMembership(
+	ctx context.Context,
+	client *sdk.APIClient,
+	cloudID, groupID int64,
+	servers []int64,
+	diags *diag.Diagnostics,
+) bool {
+	body := sdk.UpdateCloudAffinityGroupRequest{
```
matthew-c-hpe, line 204: 🟡 **Warning:** `writeMembership` sends an update payload containing ONLY the `Servers` array. In the parent resource, it's noted that the API applies an unconditional boolean check for `active` on create. If the Morpheus API applies a similar unconditional default or requires `Name`/`Active` on update, omitting them here could accidentally rename or deactivate the affinity group. Consider reading and echoing `Name` and `Active` back in this payload to be safe.

matthew-c-hpe (commented):
---

## 📋 Review: `hpe_morpheus_cloud_affinity_group_member`

**PR:** [#1337 — Affinity group membership resources](https://github.com/HPE/terraform-provider-hpe/pull/1337)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 58 files · agents: patterns, structure, spec-coverage · UI val: config-check · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 2 |
| 🟡 Warnings | 2 |
| 🔵 Recommendations | 0 |

**Verdict:** 🟡 **PASS WITH WARNINGS** — 4 issues found (0 critical)

---

### ⚙️ 4. Pattern Compliance

Overall, the new implementations are robust, and they navigate complex Terraform state and synchronization issues with an excellent design using shared locks to prevent parallel Terraform workers from overwriting array-based affinity group properties.

### 📁 1. File Completeness & Structure

*(Note: Missing `schema_gen.go` and paired data sources for the `_member` resources are expected since they are synthetic resources without direct API endpoints. The `servers` breaking change is also accepted as it was never shipped.)*

### 🔍 2. Attribute Coverage

🟡 **Warning:** In `morpheus/framework/resources/cloudaffinitygroup/update.go`, if the Morpheus API processes `tenants` (which the API contract validator flags as potentially unused), it likely suffers from the same "wholesale replace" race condition as `servers`. If so, concurrent updates from different Terraform resources could clobber the `tenants` array. This might need the same live-read treatment as `servers`.

---

**Next:** Fix 🟠 Violations before merge. 🟡 Warnings are advisory.

morpheus/framework/resources/cloudaffinitygroupmember/delete.go
eamonnotoole: Fixed in 00424be2 — `Update()` and `ImportState()` now live in `update.go` and
`import.go`. Both membership resources match the standard layout: create, read,
update, delete, import, resource.

morpheus/framework/resources/clusteraffinitygroupmember/delete.go
eamonnotoole: Fixed in 00424be2 — same split applied here.

morpheus/framework/resources/cloudaffinitygroupmember/create.go
eamonnotoole: Good thing to check, and I tested it rather than reasoning about it — but it
does not happen. Scalars are left alone when omitted; only the collection is
destructive.

On a 9.0.2.95 appliance:

```
before: name='omit-fields-test' active=True type='KEEP_TOGETHER' servers=[570236]
PUT {"affinityGroup":{"servers":[570236,571819]}}
after:  name='omit-fields-test' active=True type='KEEP_TOGETHER' servers=[570236,571819]
```

`name`, `active` and `affinityType` all survive a servers-only update, so
echoing them back is not needed.

I would argue against adding them, in fact. This resource deliberately owns one
membership and nothing else. Sending `Name` and `Active` would give it the
ability to overwrite fields it does not manage, which is the failure mode the
resource exists to avoid — the group resource owning the whole membership set is
what caused the eviction problem in the first place.

Worth noting what *is* destructive, since it is the same shape of question:
omitting `servers` from a group update empties the group. That is confirmed on
9.0.0 and 9.0.2.95, and is why the group resource re-reads membership live and
re-asserts it on every update.

matthew-c-hpe (commented):
---

## 📋 Review: `hpe_morpheus_cloud_affinity_group_member` / `hpe_morpheus_cluster_affinity_group_member`

**PR:** [#1337 — Manage affinity group membership one member at a time (MORPH-15728)](https://github.com/HPE/terraform-provider-hpe/pull/1337)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 63 files · agents: patterns, spec-coverage, structure · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 4 |
| 🟠 Violations | 1 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

**Verdict:** 🔴 **FAIL** — 5 issues found (4 critical)

---

### 🔴 Gaps

- **Missing post-create failure cleanup (cloud)**: In `morpheus/framework/resources/cloudaffinitygroup/create.go`, the read-back logic was refactored into `fetchAffinityGroup`. However, `cleanup.TaintResourceState()` was removed. If `fetchAffinityGroup` fails after a successful `Create()` POST, the function now returns early without saving state, leaving the newly created affinity group orphaned. The established pattern requires tainting the resource state when read-back fails.
- **Missing post-create failure cleanup (cluster)**: See above for `clusteraffinitygroup/create.go`.
- **404s raise diagnostics instead of removing resource (cloud member)**: In `morpheus/framework/resources/cloudaffinitygroupmember/create.go`, `readMembership` queries the single-item endpoint but fails to check for HTTP 404 before calling `errfmt.CheckResponse`. A diagnostic is appended, causing `Read()` to fail with an error rather than calling `resp.State.RemoveResource(ctx)`. This breaks drift detection when the parent group is deleted out of band. Add an `errfmt.IsNotFound(httpResp)` check before falling back to `IsSingleItemRenderFailure`.
- **404s raise diagnostics instead of removing resource (cluster member)**: See above for `clusteraffinitygroupmember/create.go`.

### 🟠 Violations

- **Missing `schema_gen.go`**: `schema_gen.go` is missing for both `cloudaffinitygroupmember` and `clusteraffinitygroupmember`. Schemas and models have been hand-coded directly into their `resource.go` files, violating the rule that schemas must be generated (`schema_gen.go` never hand-edited). Note: Since these are synthetic list-manipulation resources, this might require a synthetic configuration path, but the current state violates provider standards.

---

morpheus/framework/resources/cloudaffinitygroup/create.go
```diff
@@ -154,25 +143,14 @@ func (r *cloudAffinityGroupResource) Create(
...
-		})
-
-		return
-	}
-
-	readAg := readResult.AffinityGroup
-	if readAg == nil {
-		resp.Diagnostics.AddError("API returned nil", "AffinityGroup is nil in the response")
+	readAg, ok := fetchAffinityGroup(ctx, client, cloudID, id, &resp.Diagnostics)
+	if !ok {
```
matthew-c-hpe: 🔴 Gap — `cleanup.TaintResourceState()` is missing. If `fetchAffinityGroup` fails, the function returns early without saving state, leaving the newly created affinity group orphaned.

morpheus/framework/resources/clusteraffinitygroup/create.go
```diff
@@ -157,25 +146,14 @@ func (r *clusterAffinityGroupResource) Create(
...
-		})
-
-		return
-	}
-
-	readAg := readResult.AffinityGroup
-	if readAg == nil {
-		resp.Diagnostics.AddError("API returned nil", "AffinityGroup is nil in the response")
+	readAg, ok := fetchAffinityGroup(ctx, client, clusterID, id, &resp.Diagnostics)
+	if !ok {
```
matthew-c-hpe: 🔴 Gap — `cleanup.TaintResourceState()` is missing. If `fetchAffinityGroup` fails, the function returns early without saving state, leaving the newly created affinity group orphaned.

morpheus/framework/resources/cloudaffinitygroupmember/create.go
```diff
@@ -0,0 +1,228 @@
...
+		for _, s := range result.AffinityGroup.Servers {
+			if s.Id != nil {
+				servers = append(servers, *s.Id)
+			}
+		}
+
+		return servers, true
+	}
+
+	if !affinityread.IsSingleItemRenderFailure(httpResp) {
```
matthew-c-hpe, line 131: 🔴 Gap — Add `errfmt.IsNotFound(httpResp)` check here. Otherwise, a 404 raises a diagnostic via `errfmt.CheckResponse`, causing `Read()` to fail instead of properly removing the resource from state when drift occurs.

morpheus/framework/resources/clusteraffinitygroupmember/create.go
```diff
@@ -0,0 +1,228 @@
...
+		for _, s := range result.AffinityGroup.Servers {
+			if s.Id != nil {
+				servers = append(servers, *s.Id)
+			}
+		}
+
+		return servers, true
+	}
+
+	if !affinityread.IsSingleItemRenderFailure(httpResp) {
```
matthew-c-hpe, line 131: 🔴 Gap — Add `errfmt.IsNotFound(httpResp)` check here. Otherwise, a 404 raises a diagnostic via `errfmt.CheckResponse`, causing `Read()` to fail instead of properly removing the resource from state when drift occurs.

morpheus/framework/resources/cloudaffinitygroupmember/resource.go
```diff
@@ -0,0 +1,155 @@
...
+// Every attribute forces replacement. A membership is a fact rather than a
+// thing with settings: it either exists or it does not, so there is nothing to
+// update in place. Pointing it at a different group or a different server is a
+// different membership.
+func (r *Resource) Schema(
+	_ context.Context,
+	_ resource.SchemaRequest,
+	resp *resource.SchemaResponse,
+) {
+	resp.Schema = schema.Schema{
```
matthew-c-hpe, line 80: 🟠 Violation — Schema is hand-coded. Provider standards require schemas to be generated (`schema_gen.go`), driven by `specs/resources/*/config.yaml`.

morpheus/framework/resources/cloudaffinitygroup/create.go
eamonnotoole: You're right, and this was a regression I introduced. Fixed in 26b23954.

`cleanup.TaintResourceState()` was there before the refactor and I dropped it
when the read-back moved into `fetchAffinityGroup`. What made it slip past me is
worth recording: the removal surfaced only as an "imported and not used" error
for the `cleanup` package, and I resolved that by letting goimports drop the
import rather than asking why it had become unused.

The consequence was exactly as you describe — a group created in Morpheus but
never written to state, invisible to Terraform and never retried. Worth noting
the odds got worse rather than better, since the read-back can now fail on a
group whose single-item endpoint is broken.

The taint is restored on both group resources.

morpheus/framework/resources/clusteraffinitygroup/create.go
eamonnotoole: Same regression, same fix — restored in 26b23954.

morpheus/framework/resources/cloudaffinitygroupmember/create.go
eamonnotoole: Correct, and also a regression of mine. Fixed in 26b23954.

`readMembership` checked for the render failure but not for not-found, so a 404
fell through to `CheckResponse`, raised a diagnostic, and `Read` then returned an
error instead of calling `RemoveResource`. The original member read did have the
`IsNotFound` check; I lost it when routing that path through the fallback.

It now returns absence without a diagnostic, so a group deleted outside
Terraform surfaces as drift and the next plan offers to recreate, rather than
failing.

One caveat on verification: I could not exercise this against a live appliance,
because all three I have access to are currently unreachable. The change
restores the exact behaviour the resource had before the refactor, and unit
tests, vet and lint are clean, but the drift path itself is verified by
inspection rather than by a live run. I'll confirm it live once an appliance is
back.

morpheus/framework/resources/clusteraffinitygroupmember/create.go
eamonnotoole: Same fix as the cloud member — restored in 26b23954.

morpheus/framework/resources/cloudaffinitygroupmember/resource.go
eamonnotoole: Agreed that generated schemas are the standard, and I'd rather not special-case
it lightly — but I think hand-writing is right here, and there is precedent.

Two shipped resources already do this: `instancenode` and `instancepowerstate`
have no `schema_gen.go`, for the same reason you identify — they are synthetic,
with no single API resource behind them. `instancenode` is the closest analogue
and is what I modelled these on, including the file layout.

The generator derives a schema from an API schema. A membership has none: it is
a pairing that exists, not an object the API models, and it is manipulated by
rewriting the parent group's server list. There is nothing to generate from, so
a synthetic path would mean inventing a spec entry describing a resource the API
does not have.

Happy to revisit if you'd prefer that, but it seemed a bigger commitment than
this PR should make unilaterally — and it would want to cover the two existing
hand-written resources too, rather than leaving three exceptions with two
different rationales.

matthew-c-hpe, line 80: 🔴 **Gap:** The schema and `memberModel` were hand-written. Schemas must be generated via the code generator and never hand-edited.

morpheus/framework/resources/cloudaffinitygroupmember/create.go
```diff
@@ -0,0 +1,235 @@
...
+	var plan memberModel
+
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &plan)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	client, err := r.NewClient(ctx)
+	if err != nil {
+		resp.Diagnostics.AddError("could not create sdk client", err.Error())
```
matthew-c-hpe: 🟠 **Violation:** Bypassing the established `errfmt` package. Use `errfmt.DiagClientError(&resp.Diagnostics, err)` instead of manually appending a diagnostic.

matthew-c-hpe (commented):
---

## 📋 Review: `hpe_morpheus_*_affinity_group_member`

**PR:** [#1337 — hpe_morpheus_*_affinity_group_member](https://github.com/HPE/terraform-provider-hpe/pull/1337)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 60 files · agents: patterns, structure, spec-coverage · UI val: skipped (target not found in OpenAPI KB) · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 3 |
| 🟠 Violations | 2 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

**Verdict:** 🔴 **FAIL** — 5 issues found (5 critical)

---

### 📁 1. File Completeness

🔴 **Gap: Missing generated schemas**
The schema and `memberModel` were hand-written in `resource.go` (for both `cloudaffinitygroupmember` and `clusteraffinitygroupmember`). The provider architecture dictates that schemas must be generated via the code generator and never hand-edited. Define a `config.yaml` for each membership resource and use the code generator.

🔴 **Gap: Missing data sources for membership resources**
Every resource is required to have a matching data source. There is no data source for `cloudaffinitygroupmember` or `clusteraffinitygroupmember`. While a user could extract membership from the parent affinity group's `servers` attribute, the architectural convention requires dedicated plural data sources for feature parity.

### 🔍 2. Attribute Coverage

🔴 **Gap: Missing `required` override for `resource_permissions.groups.id`**
*File: `specs/resources/cluster_affinity_group/config.yaml` (line 89)*
Unlike `cloud_affinity_group` where `resource_permissions` was fully redefined, `cluster_affinity_group` reused the existing OpenAPI schema via a `moves` operation. Because the OpenAPI schema does not mark the ID of the `sites` item as required, `resource_permissions.groups.id` falls back to `Optional: true, Computed: true` in `schema_gen.go`. It should be `required` to prevent omitting the group ID.

### ⚙️ 4. Pattern Compliance

🟠 **Violation: Missing null guard for `g.Id` in `buildSitesPayload`**
*File: `morpheus/framework/resources/clusteraffinitygroup/update.go` (line 181)*
Due to the schema issue above, `groups.id` is optional and computed. The `buildSitesPayload` loop unconditionally calls `g.Id.ValueInt64()`.
If a practitioner omits the `id` field in HCL, this evaluates to `0` and blindly sends `id: 0` to the API.

🟠 **Violation: Bypassing `errfmt` on SDK client creation failure**
The `Create`, `Read`, `Update`, and `Delete` functions for both `cloudaffinitygroupmember` and `clusteraffinitygroupmember` bypass the established `errfmt` package when failing to instantiate the SDK client. They manually append a diagnostic error instead of using `errfmt.DiagClientError(&resp.Diagnostics, err)`.

---

morpheus/framework/resources/cloudaffinitygroupmember/create.go
eamonnotoole: Fixed in 41d22672 — all four CRUD methods on both membership resources now use
`errfmt.DiagClientError(&resp.Diagnostics, err)`, matching the group resources.

morpheus/framework/resources/cloudaffinitygroupmember/resource.go
eamonnotoole: I've left this hand-written, and I'd rather set out the reasoning than quietly
comply or quietly ignore it.

There is precedent: `instancenode` and `instancepowerstate` are shipped
resources with no `schema_gen.go`, for the reason you identified in the previous
review — they are synthetic, with no single API resource behind them.
`instancenode` is the closest analogue and is what these were modelled on,
including the file layout.

The generator derives a schema from an API schema. A membership does not have
one: it is a pairing that exists, not an object the API models, and it is
manipulated by rewriting the parent group's server list. Adding a `config.yaml`
would mean describing a resource the API does not have, so the generated result
would be authored by hand in a different file rather than genuinely generated.

If the standard should be "synthetic resources get a synthetic spec entry", I
think that is a reasonable position — but it should be decided once and applied
to the two existing hand-written resources too, rather than leaving three
exceptions with two different rationales. That felt like a bigger call than this
PR should make on its own.

Happy to do it either way; it needs a decision rather than my preference.

matthew-c-hpe (commented):
---

## 📋 Review: `hpe_morpheus_cloud_affinity_group` / `hpe_morpheus_cluster_affinity_group`

**PR:** [#1337 — Manage affinity group membership one member at a time (MORPH-15728)](https://github.com/HPE/terraform-provider-hpe/pull/1337)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 65 files · agents: patterns, spec-coverage, structure · UI val: skipped (resource not found in KB) · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 2 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

### 📁 1. File Completeness

🔴 **Gap: Missing `schema_gen.go`**
The `cloudaffinitygroupmember` and `clusteraffinitygroupmember` resources bypass the code generation pipeline entirely. There are no `config.yaml` files under `specs/` for these resources, and they define their schema manually in `resource.go` instead of embedding a generated schema. Procedural rules require all resources to define their schema via `config.yaml` to generate a `schema_gen.go`.

🔴 **Gap: Missing Paired Data Sources**
Neither `cloudaffinitygroupmember` nor `clusteraffinitygroupmember` have a paired data source implementation (missing `datasource.go`, `schema_gen.go`, `*_test.go`, and registration). Procedural rules mandate a paired data source for every resource. If the parent group's data source is intended to suffice for this purpose, this exemption must be explicitly documented.

---

matthew-c-hpe (commented):
---

## 📋 Review: `hpe_morpheus_cloud_affinity_group` and `hpe_morpheus_cluster_affinity_group`

**PR:** [#1337 — Affinity group member resources](https://github.com/HPE/terraform-provider-hpe/pull/1337)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 61 files · agents: patterns, spec-coverage, structure · UI val: config-check · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 1 |
| 🔵 Recommendations | 0 |

---

### 🟡 Warnings

- **Line length:** `morpheus/framework/datasources/clusters/datasource.go:130` — Line exceeds the 120-character maximum limit (121 chars).

#### 🔗 UI Validation (`zones-affinity-groups` and `clusters-affinity-groups`)
- **Spec gaps (ONLY in impl):** `tenantPermissions` (CREATE/UPDATE), `_active` (UPDATE)
- **Dead spec entries (ONLY in spec):** `affinityType`, `tenants`, `pool`
*(Note: `servers` is also flagged as ONLY in spec by the validator, but omitted here as the PR explicitly documents its required presence for list replacement.)*

---

morpheus/framework/datasources/clusters/datasource.go
```diff
@@ -132,7 +127,9 @@ func (d *DataSource) Read(
 		cl := &rs.Clusters[i]
 
 		// Client-side: generic filter blocks
-		if !matchesFilters(cl, typeCodes, filters) {
+		if !dsfilter.Matches(cl, filters, func(c *sdk.ListClusters200ResponseAllOfClustersInner, field string) (string, bool) {
```
matthew-c-hpe, line 130: 🟡 **Warning**: Line exceeds the 120-character maximum limit (121 chars).

morpheus/framework/resources/clusteraffinitygroup/create_unit_test.go
```diff
@@ -0,0 +1,189 @@
...
+			out += ","
+		}
+
+		out += `{"id":` + itoa(id) + `,"name":"g` + itoa(id) + `"}`
+	}
+
+	return out + `],"meta":{"total":` + itoa(int64(len(ids))) + `}}`
+}
+
+func itoa(i int64) string {
```
matthew-c-hpe: Hallucination ??

strconv.Itoa

specs/resources/cluster_affinity_group/config.yaml
```diff
@@ -82,6 +87,10 @@ cluster_affinity_group:
       computed_optional_required: computed_optional
     resource_permissions.groups.id:
       description: Group ID.
+      # Required, matching the cloud resource. The spec leaves the id of a sites
+      # entry optional, which left this Optional+Computed: a practitioner could
+      # list a group without saying which, and the payload would carry id 0.
+      computed_optional_required: required
```
matthew-c-hpe, line 94: 🔵 Recommendation — The `resource_permissions.groups.id` attribute was changed from `computed_optional` to `required`. While the comment correctly notes that this prevents an invalid `0` ID from being sent, this is a backward-incompatible breaking change that will affect configurations omitting the ID. Consider adding this to the breaking changes note in the PR description alongside the `servers` change.

matthew-c-hpe (commented):
---

## 📋 Review: `hpe_morpheus_cloud_affinity_group` and `hpe_morpheus_cluster_affinity_group`

**PR:** [#1337 — Manage affinity group membership one member at a time (MORPH-15728)](https://github.com/HPE/terraform-provider-hpe/pull/1337)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 65 files · agents: patterns, spec-coverage, structure · UI val: config-check · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 1 |

---

### 🔵 4. Recommendations

---

morpheus/framework/resources/clusteraffinitygroup/create_unit_test.go
```diff
@@ -0,0 +1,189 @@
...
+)
+
+// testClient points the SDK at a stub appliance.
+func testClient(t *testing.T, handler http.HandlerFunc) *sdk.APIClient {
+	t.Helper()
+
+	srv := httptest.NewServer(handler)
+	t.Cleanup(srv.Close)
+
+	return clientfactory.NewAPIClient(
```
matthew-c-hpe: can we do this without clientfactory, if we rip it out eventually this will get caught in the fallout

morpheus/framework/resources/clusteraffinitygroup/create_unit_test.go
eamonnotoole: Fair hit, and thank you — fixed in `0df8e8b5`.

No excuse for it. I wasn't unaware of the standard library; I was assembling the
fixture as a JSON string, needed an integer formatted mid-expression, and wrote a
helper instead of stopping to use the obvious one. Then I copied the file into the
cloud package and duplicated it.

It was also quietly broken: the loop ran while the value was positive, so any
negative input returned an empty string. Only positive ids were ever passed, so
nothing failed. That is the real argument against hand-rolling a primitive — the
defect had nowhere to show itself.

One correction to the suggestion: the value is an `int64`, so it would have been
`strconv.FormatInt(id, 10)` rather than `strconv.Itoa`, which takes an `int`.

In the event neither is needed. The helper only existed because the fixture was
built by string concatenation, which was the actual mistake, so the listing is now
constructed from the generated types and marshalled. The fixture can no longer
drift from the shape the SDK parses, and there is no integer formatting left.

While in there I removed two more of the same kind: `containsServer` is now
`slices.Contains`, which this provider already uses elsewhere, and `withoutServer`
is `slices.DeleteFunc` over a clone.

Worth noting that build, lint and tests all passed this. Nothing mechanical was
going to catch it — it needed review, so this one is squarely on me for not
re-reading the diff.

morpheus/framework/resources/clusteraffinitygroup/create_unit_test.go
eamonnotoole: Agreed, and worth doing regardless of whether `clientfactory` ever goes: a unit
test for id recovery has no business depending on how the provider builds its
production client, with the auth and transport that come with it. All the test
needs is an SDK client pointed at a stub.

I can't change code from this mode. Queued as the next edit — `sdk.NewConfiguration()`
with `Servers[0].URL` set to the test server, passed to `sdk.NewAPIClient`, which
drops the import entirely. I'll push it shortly.

specs/resources/cluster_affinity_group/config.yaml
eamonnotoole: Agreed — added. The PR description now carries a `## Breaking changes on the 2.0
line` section covering both this and `servers`, rather than mentioning `servers`
alone.

Worth stating plainly for the release notes: nothing can currently be relying on
the old behaviour. Omitting the id left it defaulting to `0`, which is not a
group, so the permission it described was never the one intended. The break is
real, but it only affects configurations that were already not doing what they
appeared to.

morpheus/framework/datasources/clusters/datasource.go
eamonnotoole: That line is mine, and it is 121 characters — but it is within this repository's
configured limit rather than over it. `.golangci.yml` sets `lll.line-length` to
132, so it passes both locally and in CI, which is why it did not surface as a
lint failure on my side.

So there is a standards question here rather than a defect: if 120 is the
intended limit, the setting is what needs changing, and there will be other lines
between 120 and 132 already in the tree. If 132 is intended, this line is fine as
it stands.

Happy to shorten it either way — it reads better wrapped — but I would rather not
special-case one line against a limit the configuration does not enforce. Tell me
which way you want it and I will make the change consistently.

specs/resources/cloud_affinity_group/config.yaml
```diff
@@ -2,14 +2,20 @@ cloud_affinity_group:
...
+          hpe_morpheus_cloud_affinity_group_member, one resource per member, so
+          that servers added by other means -- an instance provisioned into the
+          group, or a node added to one -- are not evicted.
         element_type:
           int64: {}
     tenantIds:
       set:
         computed_optional_required: computed_optional
-        description: List of tenant account IDs that are allowed access.
+        description: "Deprecated: has no effect. The Morpheus API rejects tenant assignment on affinity groups and does not apply it, so the value is retained in Terraform state only and does not reflect the appliance. Tracked as MORPH-15806."
```
matthew-c-hpe, line 17: 🔵 **Recommendation** — This long string (and the deprecation message below) will exceed the 120-character limit when generated into `schema_gen.go`. Consider using a block string (`|-`) to wrap the text across multiple lines and keep the YAML readable, matching the approach you used for `servers` above.

matthew-c-hpe (commented):
---

## 📋 Review: `hpe_morpheus_cloud_affinity_group` & `hpe_morpheus_cluster_affinity_group`

**PR:** [#1337 — Manage affinity group membership one member at a time (MORPH-15728)](https://github.com/HPE/terraform-provider-hpe/pull/1337)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 68 files · agents: patterns, spec-coverage, structure · UI val: config-check · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 2 |

**Verdict:** 🟢 **PASS** — 2 issues found (0 critical)

---

### ✨ 5. Code Quality

🔵 **Recommendations**
- **Line length in `config.yaml`**: The `description` and `deprecation_message` strings for `tenantIds` (and `description` on clusters) are written on a single line. Consider using block strings (`|-`) to keep the YAML cleanly formatted. *(See inline comments)*
- **Sweeper documentation**: The PR correctly omits sweepers for the synthetic membership resources because they carry no identifiable metadata. Consider adding a brief comment in the code to document this deliberate exclusion so future reviewers understand why they are absent.

### 🔗 3. UI Validation
*(`config-check` depth ran on `cloud_affinity_group` and `cluster_affinity_group` since their `config.yaml` changed)*

🟢 **Changed fields confirmed**:
- `servers`: The UI validator confirms `servers` is `ONLY in spec (possible dead)` for both POST and PUT, matching the PR's shift to read-only since the API does not accurately process it.
- `tenants`: The UI validator confirms `tenants` is `ONLY in spec (possible dead)` for both POST and PUT, validating the new `tenantIds` deprecation message.

**UI Validation (`cloud_affinity_group` & `cluster_affinity_group`)**
- `tenantPermissions` is `ONLY in impl (spec gaps)` for POST and PUT.
- `active`, `affinityType`, `id`, `name`, `refId`, `refType`, `source`, `visibility` are `ONLY in spec (possible dead)` for GET (Read and List).
- `_active` is `ONLY in impl (spec gaps)` for UPDATE.

---

specs/resources/cluster_affinity_group/config.yaml
```diff
@@ -2,14 +2,20 @@ cluster_affinity_group:
...
+          hpe_morpheus_cluster_affinity_group_member, one resource per member, so
+          that servers added by other means -- an instance provisioned into the
+          group, or a node added to one -- are not evicted.
         element_type:
           int64: {}
     tenantIds:
       set:
         computed_optional_required: computed_optional
-        description: List of tenant account IDs that are allowed access.
+        description: "Deprecated: has no effect. The Morpheus API rejects tenant assignment on affinity groups and does not apply it, so the value is retained in Terraform state only and does not reflect the appliance. Tracked as MORPH-15806."
```
matthew-c-hpe, line 17: 🔵 **Recommendation** — This long string (and the deprecation message below) will exceed the 120-character limit when generated into `schema_gen.go`. Consider using a block string (`|-`) to wrap the text across multiple lines and keep the YAML readable.

matthew-c-hpe (approved):

PR #1338, opened 2026-08-13
Make API non-returnable attributes write-only (Tier 2)

eamonnotoole:
## Summary

Five attributes are accepted by the Morpheus API on write but never returned on read. They were stored in Terraform state anyway, so state held values the provider could not refresh and drift was undetectable. This marks them write-only and adds a `_version` companion for each.

| Resource | Write-only attributes |
|---|---|
| `hpe_morpheus_network_router_nat` | `action`, `firewall`, `service` |
| `hpe_morpheus_load_balancer` | `group_id`, `network_server_id` |
| `hpe_morpheus_instance_clone` | `source_instance_id` |

Write-only values are nullified in the plan by the framework, so Create and Update now read them from `req.Config`. Reading them from the plan still compiles and silently sends nulls, so every affected path was switched explicitly.

Requires Terraform >= 1.11.

## Behaviour notes

- **`network_router_nat`** — `firewall` defaults to `MATCH_INTERNAL_ADDRESS` on create, because the underlying API field is required and the option type's `defaultValue` is a UI form default the API does not apply. On update an omitted `firewall` is left out of the payload entirely, so the value already on the router is preserved rather than overwritten. Getting this backwards would silently overwrite a live NSX-T value, so the create and update payload builders are extracted and tested directly.
- **`load_balancer`** — `group_id_version` and `network_server_id_version` force replacement; neither value can be changed on an existing load balancer.
- **`instance_clone`** — the clone source cannot be changed after creation, so `source_instance_id_version` has no effect and exists only so a value may be supplied without error. The best-effort recovery of `source_instance_id` from `config.cloneInstanceId` is removed: a write-only attribute must be null in state, so writing a recovered value there is invalid.

`source_instance_id`, `group_id` and `network_server_id` are dropped from the `ImportStateVerifyIgnore` lists. Write-only values are null in state on both sides of an import, so there is nothing to ignore and the import tests now verify the behaviour instead of skipping it.

## Three problems found while verifying this

These all blocked testing the change and are fixed here.

**Clone failures were invisible.** The clone runs as a `cloning` process on the source instance, but Create only polled `ListInstances` for the clone's name and never looked at that process. On one appliance the clone failed server-side after 27 seconds while Terraform kept polling for the full 15-minute timeout; on another it was still running when the timeout expired. Both produced the same diagnostic, and because `errors.Unwrap` returns nil for the error `backoff` yields, that diagnostic read `<nil>` and said nothing at all.

Create now checks the source instance's history each time round the poll and aborts immediately when the clone process has failed, reporting the process's own error text. On timeout it reports the underlying error and names the process status, so a slow clone (raise the timeout, and expect an untracked instance to clean up) is distinguishable from a failed one.

**The NSX-T test fixture was pinned to identifiers that exist nowhere.** Tier-0 router 28, group 3, network server 5 and a named edge cluster were hard-coded in three places. No appliance the tests are run against has them, so every NSX-T test failed at the first data source with `Router not found with id 28` — which reads like a provider bug but is only a stale fixture. They now come from the environment via `testhelpers.RequireNsxtFixture`:

```
TF_ACC_NSXT_TIER0_ROUTER_ID
TF_ACC_NSXT_GROUP_ID
TF_ACC_NSXT_NETWORK_SERVER_ID
TF_ACC_NSXT_EDGE_CLUSTER
```

Tests skip with a message naming the missing variables rather than failing.

**The documented `instance_clone` examples were not valid for the schema.** `volumes` and `network_interfaces` are `ListNestedAttribute`s, but the example template and every worked example in the docs used block syntax, so anyone copying them got `Blocks of type "volumes" are not expected here`. This is well-formed HCL and `terraform fmt` accepts it — only validation against the provider schema rejects it, which is why it survived. Converted to list-of-object syntax throughout.

## Testing

Unit tests for the NAT payload builders and the clone process rules were each verified by sabotage — breaking the production rule makes them fail, so they exercise the real code path rather than a copy.

Acceptance, with the NSX-T fixture supplied:

| Suite | Result |
|---|---|
| `networkrouternat` | pass |
| `loadbalancermonitor`, `loadbalancerprofile` (NSX-T LB path, exercises `network_server_id`) | pass |
| `loadbalancer` write-only plan tests | pass |
| `instance_clone` | **not verified end-to-end** |

`instance_clone` could not be verified on either appliance available to me: on one the clone fails server-side within seconds, on the other it hangs indefinitely in `cloning`. Neither failure is caused by this change — both reproduce on the pre-change code — and the clone diagnostics added here are what made the distinction visible. The clone code paths are covered by unit tests only.

Reviewers with a working clone environment should exercise `hpe_morpheus_instance_clone` create and import.

## Notes

- No `CHANGELOG.md` entry; release notes to be written for 2.0.0 as a whole. These conversions and the removal of the `ImportStateVerifyIgnore` entries are user-visible and should be covered there.
- The same stale `network_integration_id = 5` fixture remains in the firewall-rule and DHCP-server tests. Out of scope here; worth a follow-up.


RDavisHPE (approved):

specs/resources/network_router_nat/config.yaml
```diff
@@ -11,17 +14,46 @@ network_router_nat:
...
         description: The service (protocol) applied to the NAT rule (nested under config).
+    action_version:
+      int64:
+        computed_optional_required: optional
+        description: >-
+          Version marker for the write-only action attribute. Because action is write-only it is
+          never stored in state, so editing action on its own produces no plan diff. Increment
+          this value whenever action changes: the change to action_version is what produces the
+          plan diff, and the update then sends the current action value to the API.
+    firewall_version:
```
RDavisHPE, line 31: Running `task generate` I'm getting some drift between this description and the generated schema.

RDavisHPE (approved):

specs/resources/network_router_nat/config.yaml
eamonnotoole: We need to do a general task generate I think.

PR #1341, opened 2026-08-14
Fix provider Read consistency for hpegl → hpe migration import

RDavisHPE:
Fixes state inconsistencies discovered during end-to-end testing of `hpegl_vmaas_*` → `hpe_morpheus_*` migration. After migrating HCL config and running `terraform apply` with `import` blocks, several resources hit Read-vs-plan consistency errors that blocked the apply. These are all cases where the Morpheus API returns values that don't align with what the Terraform plan expects.

**Migration test result:** 16 resources imported (instance, instance\_clone, load\_balancer, 3 LB profiles, 2 LB monitors, LB pool, virtual\_server, network, network\_router, BGP neighbor, NAT rule, static route, firewall rule group) — 0 add, 0 destroy, 0 replacements.

### `RequiresNonEmptyStringAlsoRequiresInt64At`

Removed `IsUnknown()` from the null-or-unknown guard. During import, sibling attributes resolve as Unknown (the framework hasn't populated them yet). Treating Unknown the same as null caused the validator to fire incorrectly on import, blocking resources like NAT rules where `action` is write-only and the sibling attribute is legitimately set.

### `RequiresNonZeroInt64At`

Added a `== 0` early return. Reference-ID fields like `ssl_client_profile` and `ssl_server_profile` use `0` as a sentinel meaning "not configured." The migrated HCL correctly carries `ssl_client_profile = 0` from the old provider state, but the validator was treating `0` the same as a real value and requiring the sibling attribute to be set.

## Virtual server `vip_type` — UseStateForUnknown

`vip_type` is `Computed` + `Optional` with `RequiresReplace`. During import, the API returns the real value (e.g., `null` for NSX-T), but without `UseStateForUnknown` the plan phase re-resolves it as Unknown on subsequent applies, creating a perpetual diff that triggers replacement. Added `UseStateForUnknown` to the plan modifier list so the state value is preserved across plan cycles. The codegen config (`config.yaml`) was updated with `state_for_unknown: true`.

## Instance `network_interfaces` — Read consistency for `network_type_id` and `ip_mode`

The instance Read function has two code paths for populating `network_interfaces`: one from `instance.Interfaces` (used during import) and one from `server.interfaces` (used during normal CRUD). These return different subsets of fields:

- **`network_type_id`**: Present on `instance.Interfaces` but missing from `server.interfaces`. After import, subsequent reads via the server path returned null, causing `was 8, but now null`.
- **`ip_mode`**: The Morpheus API omits `ipMode` from the JSON response entirely when it's empty/unset — in *both* response paths. Both SDK models have `IpMode *string`, so both produce `types.StringNull()`. But the schema declares `Default: stringdefault.StaticString("")`, so the plan always has `ip_mode = ""`. The mismatch caused `was "", but now null`.

**Fix:** Added a gap-fill loop after the server-path read that copies `NetworkTypeId` from the instance-level response when the server response has null. For `IpMode`, since both API paths omit it, the gap-fill also defaults null to `""` to match the schema default.

## Load balancer profile `tags` — Computed + null guard

The LB profile `tags` attribute had two issues:

1. **Schema:** `tags` was `Optional`-only but the Morpheus API populates tags server-side (NSX-T injects sentinel entries like `{name:"",value:""}` even when no tags were configured). Made it `Computed` + `Optional` in the codegen config so the provider is allowed to set it.

~~2. **Read null guard:** When the user never configured tags (prior state is null), the Read function was calling `readTagsFromConfig` which returned API-injected sentinels, creating a perpetual diff. Added a guard: if `prior.Tags.IsNull()`, return `types.SetNull(...)` instead of reading from the API response. This also avoids surfacing an upstream Morpheus API bug where NSX-T tag CRUD sends `{name,value}` instead of `{scope,tag}`, causing 400 errors on any profile create/update that includes tags.~~ Unnecessary with optional_computed

## Virtual server `vip_hostname` — Coerce empty string to null

`vip_hostname` is `Optional`-only (not `Computed`). The migrated HCL doesn't set `vip_hostname` (it wasn't configured in the old provider). The plan resolves it to null. But the Morpheus API returns `""` for an empty hostname, and `convert.StrToType` converts that to `types.StringValue("")`. Since the attribute isn't `Computed`, the framework rejects a provider-driven change from null to `""`.

**Fix:** Inline coercion in Read — when the API returns nil or `""`, return `types.StringNull()` instead. An empty hostname is semantically "not set."

## Known remaining issue

**Instance clone Update** — The `instanceclone` Update function unconditionally calls the `ResizeInstance` API on every update, regardless of what actually changed. Unlike the `instance` resource (which has `isAPIUpdateNeeded` guards, plan-vs-state volume comparison, and a separate `UpdateInstance` call for metadata), the clone Update has no such logic. Any diff — even a name change or computed field resolution — triggers a resize, which returns a Morpheus 500 error. This is a pre-existing provider bug unrelated to migration; the migration works around it by stripping attributes that would cause diffs (via cleanup config). A proper fix requires rewriting the clone Update function to match the instance Update's pattern.

### DIFF
--- a/morpheus/framework/resources/loadbalancervirtualserver/resource_read.go
+++ b/morpheus/framework/resources/loadbalancervirtualserver/resource_read.go
@@ -160,7 +160,15 @@ func getVirtualServerAsState(
 	state.VipAddress = convert.StrToType(vs.VipAddress)
 	state.VipPort = convert.Int64ToType(vs.VipPort)
 	state.VipProtocol = convert.StrToType(vs.VipProtocol)
-	state.VipHostname = convert.StrToType(vs.VipHostname.Get())
+	// VipHostname: coerce empty string to null. The API returns "" for an
+	// unset hostname; since vip_hostname is Optional (not Computed), Read must
+	// return null when the user never configured it to avoid an inconsistent
+	// result (null in plan vs "" from API).
+	if h := vs.VipHostname.Get(); h == nil || *h == "" {
+		state.VipHostname = types.StringNull()
+	} else {
+		state.VipHostname = types.StringValue(*h)
+	}
 
 	// Computed fields
 	state.Active = convert.BoolToType(vs.Active)
--- a/specs/resources/load_balancer_profile/config.yaml
+++ b/specs/resources/load_balancer_profile/config.yaml
@@ -13,7 +13,7 @@ load_balancer_profile:
     # injects it into the active config block). Replaces the per-block tags.
     tags:
       set_nested:
-        computed_optional_required: computed_optional
+        computed_optional_required: optional
         description: NSX-T tags applied to the profile.
         nested_object:
           attributes:
@@ -95,7 +95,7 @@ load_balancer_profile:
         - LBClientSslProfile
         - LBServerSslProfile
     tags:
-      computed_optional_required: computed_optional
+      computed_optional_required: optional
       description: NSX-T tags applied to the profile.
     # cookie_type is exposed with friendly values; the provider translates
     # session<->LBSessionCookieTime and persistence<->LBPersistenceCookieTime.
--- a/specs/resources/load_balancer_profile/config.yaml
+++ b/specs/resources/load_balancer_profile/config.yaml
@@ -13,7 +13,7 @@ load_balancer_profile:
     # injects it into the active config block). Replaces the per-block tags.
     tags:
       set_nested:
-        computed_optional_required: optional
+        computed_optional_required: computed_optional
         description: NSX-T tags applied to the profile.
         nested_object:
           attributes:
@@ -95,7 +95,7 @@ load_balancer_profile:
         - LBClientSslProfile
         - LBServerSslProfile
     tags:
-      computed_optional_required: optional
+      computed_optional_required: computed_optional
       description: NSX-T tags applied to the profile.
     # cookie_type is exposed with friendly values; the provider translates
     # session<->LBSessionCookieTime and persistence<->LBPersistenceCookieTime.
--- a/morpheus/framework/resources/instance/instance_read.go
+++ b/morpheus/framework/resources/instance/instance_read.go
@@ -1746,6 +1746,22 @@ func getStateInterfaces(
 		return nil, diags
 	}
 
+	// The server interfaces response does not carry all the fields that the
+	// instance-level interfaces response does (e.g. NetworkInterfaceTypeId,
+	// and IpMode may be absent). Fill gaps from intfsFromInstance so that
+	// Read returns consistent values regardless of which code path is taken.
+	for i := range intfsFromServer {
+		if i >= len(intfsFromInstance) {
+			break
+		}
+		if intfsFromServer[i].NetworkTypeId.IsNull() || intfsFromServer[i].NetworkTypeId.IsUnknown() {
+			intfsFromServer[i].NetworkTypeId = intfsFromInstance[i].NetworkTypeId
+		}
+		if intfsFromServer[i].IpMode.IsNull() || intfsFromServer[i].IpMode.IsUnknown() {
+			intfsFromServer[i].IpMode = intfsFromInstance[i].IpMode
+		}
+	}
+
 	// Get []NetworkInterfacesValue from the plan
 	var intfsFromPlan []NetworkInterfacesValue
 	pd := plan.NetworkInterfaces.ElementsAs(ctx, &intfsFromPlan, false)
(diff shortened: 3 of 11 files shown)


morpheus/framework/resources/loadbalancervirtualserver/resource_read.go
```diff
@@ -160,7 +160,15 @@ func getVirtualServerAsState(
 	state.VipAddress = convert.StrToType(vs.VipAddress)
 	state.VipPort = convert.Int64ToType(vs.VipPort)
 	state.VipProtocol = convert.StrToType(vs.VipProtocol)
-	state.VipHostname = convert.StrToType(vs.VipHostname.Get())
+	// VipHostname: coerce empty string to null. The API returns "" for an
```
eamonnotoole, line 163: 🔴 **Gap: `vip_hostname` cannot be cleared on update (incomplete fix)**

The PR correctly coerces an empty string from the API into `Null` here, but misses the corresponding `Update` logic needed to actually clear the value when a user removes it.

In `resource_update.go`, if a user removes `vip_hostname`, `plan.VipHostname.IsNull()` becomes true. The current guard skips it, leaving `instance.VipHostname` as a `nil` pointer. Because `VipHostname` is `*string` with `omitempty` in the SDK update model, the field is omitted from the JSON request. The backend ignores the omission and retains the old hostname, leading to a perpetual diff.

**Fix:** Modify `resource_update.go` to explicitly send `""` to the API when the field is cleared:
```go
if !plan.VipHostname.IsUnknown() {
    if plan.VipHostname.IsNull() {
        instance.VipHostname = sdk.PtrString("")
    } else {
        instance.VipHostname = plan.VipHostname.ValueStringPointer()
    }
}
```

specs/resources/load_balancer_profile/config.yaml
```diff
@@ -13,7 +13,7 @@ load_balancer_profile:
     # injects it into the active config block). Replaces the per-block tags.
     tags:
       set_nested:
-        computed_optional_required: optional
+        computed_optional_required: computed_optional
```
eamonnotoole: 🔴 **Gap: `tags` cannot be cleared due to `Computed` block limitation**

`tags` is a `SetNestedAttribute` (an HCL block). Terraform HCL does not permit assigning empty collections to blocks (e.g., `tags = []` is invalid syntax). If a user attempts to clear their tags by deleting all `tags {}` blocks from their configuration, Terraform treats the missing `Computed` block as unconfigured and falls back to the prior state (planning it as `Unknown`).

Consequently, in `update.go`, `tags.IsUnknown()` evaluates to true in `tagsFromPlan`, the SDK slice is left `nil`, and the update request omits the tags payload. The backend never clears the tags.

**Fix:** Revert `tags` to just `Optional` (not `Computed`). The PR's excellent new logic in `readTagsFromConfig` already filters out the `{name:"", value:""}` sentinels and returns `types.SetNull()` when the result is empty. That alone is sufficient to fix the perpetual diff for unconfigured tags without sacrificing the user's ability to clear them.

eamonnotoole (commented):
---

## 📋 Review: `hpe_morpheus_load_balancer_profile`, `hpe_morpheus_load_balancer_virtual_server`, `hpe_morpheus_instance`

**PR:** [#1341 — Fix provider Read consistency for hpegl → hpe migration import](https://github.com/HPE/terraform-provider-hpe/pull/1341)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 10 files · agents: patterns, spec-coverage, structure · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 2 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

- 🟡 **Naming Deviation**: The `instance` resource prefixes its primary logic files (`instance_create.go`, `instance_read.go`, etc.) instead of using the standard `create.go`, `read.go`.
- 🟡 **Naming Deviation**: The `loadbalancervirtualserver` resource similarly uses a `resource_` prefix for its CRUD files (`resource_create.go`, `resource_read.go`, etc.).

---

eamonnotoole (commented):
---

## 📋 Review: `hpe_morpheus_load_balancer_profile`, `hpe_morpheus_load_balancer_virtual_server`, `hpe_morpheus_instance`

**Scope:** 🎯 PR diff vs `origin/dev-2.0` (Latest commits) — 9 files · agents: patterns, spec-coverage, structure · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

- 🟡 **Missing from Spec**: `vip_type` is correctly configured in the resource but is missing from the OpenAPI `loadBalancerInstanceCreate` schema (it is only present in the read schema).
- 🟡 **Naming Deviation**: The `instance` resource prefixes its primary logic files (`instance_create.go`, `instance_read.go`, etc.) instead of using the standard `create.go`, `read.go`.
- 🟡 **Naming Deviation**: The `loadbalancervirtualserver` resource similarly uses a `resource_` prefix for its CRUD files (`resource_create.go`, `resource_read.go`, etc.).

---

eamonnotoole (approved):

PR #1346, opened 2026-08-17
Add hpe_morpheus_instance_disk_type data source (MORPH-15864)

jasonquigleyhpecom:
## Summary

Adds a `hpe_morpheus_instance_disk_type` data source — the hpe equivalent of hpegl's `hpegl_vmaas_instance_disk_type`. It resolves a disk type (Morpheus storage volume type) by name within a cloud, layout and group, and returns its `id` for use as `storage_type_id` on `hpe_morpheus_instance` volumes.

This continues the hpegl parity lookup work, alongside `instance_storage_controller`, `network_proxy` and `network_interface_type`.

## Usage

```hcl
data "hpe_morpheus_cloud" "example" {
  name = "hvm"
}

data "hpe_morpheus_instance_disk_type" "example" {
  name      = "Standard"
  cloud_id  = data.hpe_morpheus_cloud.example.id
  layout_id = 77
  group_id  = 1
}

resource "hpe_morpheus_instance" "example" {
  volumes = [{
    storage_type_id = data.hpe_morpheus_instance_disk_type.example.id
  }]
}
```

## Migrating from hpegl

The lookup is the same; the cloud, layout and group ids that hpegl takes as strings are `int64` here, so they compose directly with other hpe data sources.

| hpegl | hpe |
|---|---|
| `hpegl_vmaas_instance_disk_type` | `hpe_morpheus_instance_disk_type` |
| `name`, `cloud_id`, `layout_id`, `group_id` (strings) | same argument names, `int64` ids |
| `.id` | `.id` (use as `storage_type_id`) |

## Implementation notes

- The disk types available depend on the provisioning context, so the lookup is scoped by cloud, layout and group and reads `GET /api/instances/service-plans` — the same endpoint hpegl uses.
- The schema is generated into `schema_gen.go` from the new data source config under `specs/datasources/instance_disk_type/`. Disk types are nested at `plans[].storageTypes[]` with no singular component to generate from, so the matched disk type's fields are surfaced as top-level attributes and the `plans[]` envelope is dropped.
- Morpheus repeats the same storage types in every service plan, so the read flattens and dedupes them by `id` before matching.
- Matching is case-insensitive and whitespace-trimmed, matching hpegl. Exactly one match is required: zero returns `no instance disk type found`, more than one returns `multiple instance disk types were returned`.

## Attribute coverage

Every scalar on the underlying model is exposed (35 attributes), matching the convention set by `network_type`. Six attributes the API returns were missing from the API schema entirely — `nameEditable`, `planResizable`, `hasISO`, `noStorage`, `multiAttachSupported` and `hasActiveReplica` — so they are added there and the vendored SDK regenerated. Types were confirmed against the API implementation, where all six are booleans (`hasActiveReplica` additionally nullable).

Two model fields are deliberately not exposed:

- `optionTypes` — an array of free-form maps, which `network_type` omits for the same reason.
- `createDatastore` — the API returns a JSON boolean where the schema declares a nullable string, so the SDK surfaces the coerced string `"0"`/`"1"`. Exposing that would be misleading, so it is omitted until the schema types it as a boolean.

## Testing

- `make lint` — clean (0 issues).
- Unit tests cover dedupe-by-id, name matching (exact, case-insensitive, whitespace-trimmed, zero match, multiple matches), and the state mapping across each shape the model mixes (plain pointer, nullable string, nullable bool, int32 widening, and absent values staying null).
- Acceptance tests pass against a live appliance: `TestAccMorpheusFindInstanceDiskType` and `TestAccMorpheusFindInstanceDiskTypeNotFound`. They use long-lived seeded fixtures (cloud resolved by name, layout 77, group 1, the `Standard` disk type) and require no test-specific environment variables.
- `make docs` regenerated; re-running it leaves a clean tree.

## Related PRs

- Companion API spec PR: https://github.com/HewlettPackard/morpheus-openapi/pull/336

This PR vendors its own generated SDK, so it is self-contained and merges first; the spec PR follows.

## Follow-up

A plural `hpe_morpheus_instance_disk_types` list data source is planned for a future PR against the same endpoint.


morpheus/framework/datasources/instancedisktype/datasource_internal_test.go
```diff
@@ -0,0 +1,185 @@
...
+	in := diskType{
+		Id:           int32Ptr(1),
+		Name:         strPtr("Standard"),
+		Code:         strPtr("standard"),
+		DisplayOrder: int32Ptr(3),
+		VolumeType:   strPtr("disk"),
+		Enabled:      boolPtr(true),
+		HasISO:       boolPtr(false),
+	}
+	in.Description.Set(strPtr("Standard"))
```
matthew-c-hpe: 🔵 **Recommendation** — The standard idiomatic style for constructing SDK models in tests is inline via struct literals using `sdk.NewNullable*` wrappers (e.g., `Description: *sdk.NewNullableString(strPtr("Standard"))`). Here, the test relies on calling `.Set()` on the zero-valued nullable struct field instead. While fully functional, updating to the inline constructor format keeps model declarations strictly declarative.

matthew-c-hpe (commented):
---

## 📋 Review: `hpe_morpheus_instance_disk_type`

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 19 files · agents: patterns, spec-coverage, structure · UI val: config-check · connected: [#336](https://github.com/HewlettPackard/morpheus-openapi/pull/336)

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 1 |

**Verdict:** 🟢 **PASS** — 1 issues found (0 critical)

---

### ⚙️ 4. Pattern Compliance

🔵 **Recommendation** — **Use inline nullable constructors**
- **File**: `morpheus/framework/datasources/instancedisktype/datasource_internal_test.go`
- **Details**: The standard idiomatic style for constructing SDK models in tests is inline via struct literals using `sdk.NewNullable*` wrappers (e.g., `Description: *sdk.NewNullableString(strPtr("Standard"))`). Here, the test relies on calling `.Set()` on the zero-valued nullable struct field instead. While fully functional, updating to the inline constructor format keeps model declarations strictly declarative. See inline comments for specifics.

* 🔵 **Directory Naming for Examples**: The `go:generate` command outputs to `examples/data-sources/morpheus_instance_disk_type/`, which drops the `hpe_` prefix. While this follows an existing idiosyncratic pattern across the provider, Terraform registry documentation typically expects the directory name to match the full data source name (`hpe_morpheus_instance_disk_type`).

The following new API fields were successfully captured from the companion-added set in the connected PR, built into the SDK `diskType` model, and fully wired into the Terraform schema and state mapper:
* `nameEditable` (as `name_editable`)
* `planResizable` (as `plan_resizable`)
* `hasISO` (as `has_iso`)
* `noStorage` (as `no_storage`)
* `multiAttachSupported` (as `multi_attach_supported`)
* `hasActiveReplica` (as `has_active_replica`)

---

morpheus/framework/datasources/instancedisktype/datasource_internal_test.go
```diff
@@ -0,0 +1,185 @@
...
+			}
+		})
+	}
+}
+
+// TestDiskTypeAsState covers the three mapping shapes the model mixes: plain
+// pointers, NullableString/NullableBool (which must be read through Get), and
+// int32 widening. It also pins that absent values stay null rather than being
+// flattened to a zero value.
+func TestDiskTypeAsState(t *testing.T) {
```
matthew-c-hpe: I dont think we need this really

matthew-c-hpe (approved):

PR #1347, opened 2026-08-17
Make the datastore and service plan lookups work on Private Cloud appliances

eamonnotoole:
## Summary

`hpe_morpheus_service_plan` and `hpe_morpheus_datastore` could not resolve a
name on a Private Cloud appliance. Both now can.

## The problem

Both looked the object up by name, got it back from the listing, then discarded
it and fetched it again by id. That second request is the one that fails:

```
GET /api/service-plans?provisionType=vmware   lists 407 G1-Small
GET /api/service-plans/407                    404 "Service Plan not found with id 407"
```

Plan 407 is real -- every instance running on that appliance uses it. The same
holds for datastores associated with a cloud rather than a cluster.

The datastore data source had a second problem. It refused any datastore
without `refType` and `refId`, and that appliance reports neither: the listing
carries only an id and a name, and the single-item endpoint 404s. So every
datastore lookup failed, by either route.

### This is the whole VMware surface, not an edge case

The 404 is not spread evenly. On both appliances checked, the datastores whose
single-item read fails are exactly the cloud-associated ones, and every VMware
datastore is cloud-associated:

| type | associated with | single-item GET |
|---|---|---|
| `vmfs` (VMware) | Cloud | **404** |
| `Directory Pool` (KVM/HVM cluster storage) | Cluster | 200 |
| Kubernetes storage classes | Cluster | 200 |

So this was not "some datastores cannot be read". For anyone provisioning
VMware, it was every datastore they could actually use -- the two `vmfs`
datastores on the Private Cloud appliance are precisely the two that failed,
while the cluster-associated ones resolved fine throughout.

## What changed

**The re-read is gone.** The listing entry and the single-item response are
generated from the same API object, so the entry is converted and mapped as
before. Conversion is by re-encoding rather than field by field -- thirty-odd
fields and several nested structures per side, and a hand-written copy would be
silently wrong the first time the SDK gained a field.

**The datastore fetch became an optimisation.** Older appliances answer the
listing with only an id and a name, and the fetch carries more, so it is still
attempted -- but a failure now falls back to the entry the listing already
returned rather than failing the read.

**A missing association is no longer fatal.** `refType`/`refId` drive
`associated_resource_type`, `associated_resource_id`, `tenants` and
`resource_permissions`. Those are now null when the API reports no association,
which is how the plural `hpe_morpheus_datastores` has always treated the same
absence. An unrecognised `refType` is treated the same way.

**Explicit paging.** Both lookups ask for a page rather than accepting the
default of 25, and report when more objects share a name than were fetched.
`name` is an exact server-side filter, so `meta.total` counts matches rather
than the collection. Paging further would not help -- the practitioner still
has to say which one they meant -- so it asks for an ID instead of quietly
using the first match.

**A latent bug, fixed.** The existing error paths built their null tenants set
with `types.SetNull(TenantsType{})`, which carries no attribute types and so
produces an empty object type instead of the schema's. Terraform reports that
as a provider bug rather than as an absent value. The type is now derived from
the value.

## Verification

Against two appliances, plus unit tests for the conversions and the sparse
listing path.

| | 8.0.7 (Private Cloud) | 9.0.1 |
|---|---|---|
| `service_plan` "G1-Small" | **407** -- previously 404 | unchanged |
| `datastore` (cloud-associated) | **resolves** -- previously failed | unchanged |
| `associated_resource_type` | null (not reported) | `"Cloud"` |

The relaxation is a behaviour change: anyone reading `associated_resource_type`
now gets null instead of an error on appliances that do not report it. Null is
the honest answer -- the appliance did not say -- but it is silent, logged only
at debug level. A warning was left out deliberately, since it would fire on
every read for every Private Cloud user.

Separately, the appliance side deserves attention regardless of what the
provider does: a single-item GET returning 404 for an object the listing
returns, and that running instances actively use, is wrong.

## Out of scope

`hpe_morpheus_datastore` and `hpe_morpheus_service_plan` only. The other plural
data sources cap their listing and none checks `meta.total`, so a truncated
answer is silently a partial one. That is worth its own change.

### DIFF
--- a/morpheus/framework/datasources/datastore/convert_internal_test.go
+++ b/morpheus/framework/datasources/datastore/convert_internal_test.go
@@ -1,0 +1,161 @@
+// (C) Copyright 2026 Hewlett Packard Enterprise Development LP
+
+package datastore
+
+import (
+	"testing"
+
+	sdk "github.com/HPE/terraform-provider-hpe/internal/sdk/oapigen"
+)
+
+func strp(s string) *string { return &s }
+func boolp(b bool) *bool    { return &b }
+func i64p(i int64) *int64   { return &i }
+
+// TestDatastoreFromListEntry pins the conversion the by-name path relies on.
+//
+// The listing entry and the single-item shape are generated from the same API
+// object, so this asserts the fields actually survive re-encoding rather than
+// silently arriving as zero values.
+func TestDatastoreFromListEntry(t *testing.T) {
+	t.Parallel()
+
+	in := &sdk.ListDatastores200ResponseAllOfDatastoresInner{
+		Id:           4,
+		Name:         "datastore1",
+		Type:         "vmfs",
+		Status:       "available",
+		RefType:      strp("ComputeZone"),
+		RefId:        i64p(1),
+		Active:       boolp(true),
+		Online:       boolp(true),
+		ExternalId:   strp("datastore-11"),
+		ExternalType: strp("vmfs"),
+		DatastoreType: sdk.ListDatastores200ResponseAllOfDatastoresInnerDatastoreType{
+			Id:   7,
+			Code: "vmware-datastore",
+		},
+	}
+
+	out, err := datastoreFromListEntry(in)
+	if err != nil {
+		t.Fatalf("conversion failed: %v", err)
+	}
+
+	if out.Id != in.Id {
+		t.Errorf("id = %d, want %d", out.Id, in.Id)
+	}
+
+	if out.Name != in.Name {
+		t.Errorf("name = %q, want %q", out.Name, in.Name)
+	}
+
+	if out.Type != in.Type {
+		t.Errorf("type = %q, want %q", out.Type, in.Type)
+	}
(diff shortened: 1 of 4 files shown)


jasonquigleyhpecom (commented):
---

## 📋 Review: `hpe_morpheus_datastore` / `hpe_morpheus_service_plan`

**PR:** [#1347 — Make the datastore and service plan lookups work on Private Cloud appliances](https://github.com/HPE/terraform-provider-hpe/pull/1347)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 4 files · agents: review-patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

matthew-c-hpe (approved):

jasonquigleyhpecom (approved):

PR #1348, opened 2026-08-18
Stop an imported instance planning changes nobody made

eamonnotoole:
## Summary

Importing an instance and planning it reported changes nobody had made, and on
appliances before 8.1.2 escalated them to a replacement -- destroying and
recreating a running VM.

## The problem

Two attributes declare a schema default the read path could not produce:

| attribute | schema default | read path produced |
|---|---|---|
| `network_interfaces[].ip_mode` | `""` | `null` |
| `config_vmware.nested_virtualization` | `"off"` | `null` |

The API omits both when they were never set. Null in state against a default in
the plan is a change Terraform acts on, so the diff could never settle. On
`network_interfaces` that meets the `>= 8.1.2` constraint, which turns a
cosmetic difference into `1 to destroy`.

`ip_mode` was already handled -- but in one caller rather than at the source.
The normalisation sat in the loop that merges server and instance interfaces,
and the import path returns before reaching it:

```go
// If this is an import, then return intfsFromInstance
if plan.Name.IsNull() || plan.Name.IsUnknown() {
    return intfsFromInstance, diags
}
```

`nested_virtualization` had no normalisation at all. Its comment reasoned that
null was *"valid for this Optional+Computed attribute"*, which overlooked the
default.

## What changed

**Both are resolved where the value is built**, so every caller sees the same
thing regardless of path. The special case in the merge loop is gone rather
than duplicated.

**The version constraint now compares the way the plan modifiers already do.**
`checkForAttributeUpdate` compared plan against state strictly, while
`restoreUnchangedComputedAttributes` deliberately fills unknowns from state
first. The framework marks every null computed attribute unknown whenever
anything in the resource differs, so editing an unrelated attribute left the
interface and volume collections unknown and looked like a reconfiguration. The
two are now consistent.

## Verification

Against a Private Cloud appliance at 8.0.7, which is below the constraint and so
shows the worst case:

| | before | after |
|---|---|---|
| plan after import | `1 to import, 1 to add, 1 to destroy` | `1 to import, 0 to add, 0 to change, 0 to destroy` |
| plan after apply | — | `No changes` |

Found while migrating a configuration from the hpegl provider, but it is not
specific to that: `terraform import` followed by `terraform plan` showed the
same diff.

The version-gate change is the one to scrutinise. It makes a replacement less
likely, which is the safe direction, but it means a genuine reconfiguration
must now differ in a *known* value to trigger the constraint. That is the
intent -- unknowns carry no practitioner intent -- and it matches the reasoning
already documented in `instance_plan_modifiers.go`.

## Out of scope

`hpe_morpheus_instance` declares 17 schema defaults. Two are fixed here; the
rest (`create_user`, `no_agent`, `public_ip_type`, `instance_profile`,
`enforce_raid_boot_volume`, `wait_for_ip_address`) have not been checked for the
same shape, and the pattern is not confined to this resource. Any attribute with
a schema default whose read path can write null will diff after import. That
sweep deserves its own change.

### DIFF
--- a/morpheus/framework/resources/instance/instance_read.go
+++ b/morpheus/framework/resources/instance/instance_read.go
@@ -970,16 +970,25 @@ func getNoAgent(
 	return nil, diags
 }
 
+// nestedVirtualizationDefault mirrors the schema default for
+// config_vmware.nested_virtualization and config_hvm.nested_virtualization.
+var nestedVirtualizationDefault = "off"
+
 func getNestedVirtualization(
 	id int64,
 	apiConfig *apiConfigType,
 ) (*string, diag.Diagnostics) {
-	// nestedVirtualization is optional: hpegl never sets it, so Morpheus omits
-	// it from the GET response entirely (IsSet() == false).  Treat absence as
-	// nil rather than an error — the caller passes nil to convert.StrToType
-	// which produces a null value, valid for this Optional+Computed attribute.
+	// nestedVirtualization is optional and Morpheus omits it from the GET
+	// response entirely when it was never set (IsSet() == false).
+	//
+	// Absence resolves to the schema default rather than to nil. The attribute
+	// is Optional+Computed with a default of "off", so a configuration that
+	// leaves it out plans "off"; returning nil would put null in state, and
+	// null against "off" is a change Terraform acts on. On an instance that
+	// means a permanent diff on config_vmware after import, for an attribute
+	// nobody set. Same reasoning as ipModeOrDefault.
 	if !apiConfig.NestedVirtualization.IsSet() {
-		return nil, nil
+		return &nestedVirtualizationDefault, nil
 	}
 
 	return apiConfig.NestedVirtualization.Get(), nil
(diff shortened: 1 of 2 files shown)


jasonquigleyhpecom (commented):
---

**PR:** [#1348 — Stop an imported instance planning changes nobody made](https://github.com/HPE/terraform-provider-hpe/pull/1348)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 2 files · agents: patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

🟠 **`morpheus/framework/resources/instance/instance_read.go`** (line 752)
The `isEC2` pointer (`apiConfig.IsEC2`) is dereferenced as `*isEC2` without a `nil` check before passing it to `convert.StringToBool`. If the API omits this optional field, the dereference will cause a runtime panic.

🟡 **`morpheus/framework/resources/instance/instance_read.go`** (lines 289-290)
Duplicate assignment for `state.InstanceTypeId` (mapped twice from `instance.InstanceType.Id`).

🟡 **`morpheus/framework/resources/instance/resource.go`** (lines 44-45)
Duplicate assignment for `resp.TypeName` in `Metadata`.

---

jasonquigleyhpecom (approved):

matthew-c-hpe (approved):

PR #1349, opened 2026-08-18
Triage MORPH-16049: Use a real provider block in no-search-attrs tests

jasonquigleyhpecom:
The `*NoSearchAttrs` / `*NoSearchTerms` data source tests could never match the
error they asserted on, because the provider bailed out before the code that
produces that error ever ran.

## Symptom

39 data source tests failed, each on an `ExpectError` that never matched:

```
Error: Missing Morpheus connection details
```

## Cause

Each test built its config from a package-local `providerConfigOffline` const
with `url = ""`. `morpheus/provider.go:105-112` treats an empty URL as fatal:

```go
if m.URL.ValueString() == "" {
    resp.Diagnostics.AddError(
        "Missing Morpheus connection details",
        missingConnectionDetails,
    )

    return
}
```

`ConfigureProvider` therefore never completes, so the data source `Read` — the
only producer of `ErrorNoValidSearchTerms` — never runs, and the error the test
expects is never emitted.

## Fix

Replace `providerConfigOffline` with `testhelpers.ProviderBlock()` in the failing
test only. This mirrors
`morpheus/framework/datasources/networkproxy/datasource_test.go:211-240`, which
already used this pattern and passed. `policy` additionally drops
`IsUnitTest: true` and moves to `testhelpers.GetAccTestFactories(...)`.

The `providerConfigOffline` const is removed only from the 18 packages where
nothing referenced it any more; the other 21 touched packages keep it.

## Deliberately not changed

The `*BothSearchAttrs` / `*IdAndName` / `*IdAndVersion` tests keep the offline
block. Their `Invalid Attribute Combination` is raised at
`ValidateDataResourceConfig`, before `ConfigureProvider`, so an empty URL never
reaches them — all 27 passed already. Converting them would risk regressing 27
green tests to fix none.

## Tests

Verified by acceptance rerun: all 39 previously failing tests now pass, the 27
regression-guard tests above still pass, and nothing regressed.

## Notes

A follow-up sweep is worth doing: 12 further data source packages carry the same
`providerConfigOffline` pattern but are capability-skipped on the test appliance,
so they never failed. They will surface the moment those capabilities are enabled.

Closes MORPH-16049.


jasonquigleyhpecom (commented):
## 📋 Review: `hpe_morpheus_*` (Bulk test update)

**PR:** [#1349 — Triage MORPH-16049: Use a real provider block in no-search-attrs tests](https://github.com/HPE/terraform-provider-hpe/pull/1349)

**Scope:** 🎯 PR diff vs `origin/dev-2.1` — 39 files · agents: review-patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 1 |

### ⚙️ 4. Pattern Compliance

🔵 **Recommendation: Standardize provider factories in updated tests**
While `policy/datasource_test.go` was successfully migrated to use `testhelpers.GetAccTestFactories(t, adapter.NewMorpheus(), nil)` inside its updated `resource.TestCase` block, several other data source tests modified in this PR (such as `user`, `role`, `cloud`, `group`, `environment`, and `instancetypelayout`) still refer to local `testAccProtoV6ProviderFactories` implementations within their `*NoSearchAttrs` tests. Consider taking this opportunity to align them with the `testhelpers.GetAccTestFactories` standard.

- **Test infrastructure (`TestMain` definition)**: Numerous files touched by this PR still implement `code := m.Run()` instead of the required `code := testhelpers.TestMain(m)` behavior within their `TestMain` definitions. This gap exists in the tests for `cloud`, `cluster`, `datastore`, `group`, `networkdomain`, `networkpool`, `networkserver`, `networktype`, `ostype`, `policy`, `role`, `securitygroup`, and `user`.
- **Test infrastructure (SDK Factories)**: A wide variety of data source tests still maintain a local `var testAccProtoV6ProviderFactories` map rather than completely adopting `testhelpers.GetAccTestFactories(...)` across all underlying test definitions.
- **Test infrastructure (`IsUnitTest`)**: The `IsUnitTest: true` flag continues to be improperly set on acceptance tests attempting actual API interaction. This is present in `morpheus/framework/datasources/policy/datasource_test.go` (under `TestAccMorpheusPolicyDataSourceBothSearchAttrs`).

---

matthew-c-hpe (approved):

PR #1350, opened 2026-08-18
Triage MORPH-16051: Fix password_wo test never reaching the plan

jasonquigleyhpecom:
`TestAccMorpheusUserResourceMissingPasswordWo` failed in the nightly acceptance run. The cause is a single broken test fixture, not missing validation.

## Symptom

The test asserts that creating an `hpe_morpheus_user` without `password_wo` produces a `'password_wo' not set` diagnostic. It got a different error instead:

```
Expected an error with pattern ('password_wo' not set)
  | Error: Missing Morpheus connection details
--- FAIL: TestAccMorpheusUserResourceMissingPasswordWo (0.55s)
```

## Cause

The test built its own provider block inline, with empty values:

```hcl
provider "hpe" {
  morpheus {
    url = ""
    username = ""
    password = ""
  }
}
```

Terraform calls `ConfigureProvider` before `PlanResourceChange`. With no URL the provider failed configuration outright and the test terminated, so the plan was never produced and the diagnostic the test waits for could never be emitted.

## Fix

Use `testhelpers.ProviderBlock()`, as the rest of the package does, so the provider configures and planning is actually reached. Test-only — no provider source, `schema_gen.go` or `specs/` changes.

## The enforcement being tested already exists

Worth recording, because the initial diagnosis was wrong in a way that would have done damage.

Triage first concluded that `password_wo` had no create-time enforcement, and proposed adding a `ValidateConfig`. That came from grepping `resource.go`, which indeed does not contain the check — but the schema does:

- `morpheus/framework/resources/user/schema_gen.go:87-96` attaches `modifiers.RequireOnCreateModifier{}` to `password_wo`
- `utils/modifiers/modifiers.go:24-50` emits `attribute 'password_wo' not set (this attribute is optional for some operations, eg import, but needed during create)`

The enforcement was there all along. Only the test was broken.

### Why `ValidateConfig` would have been the wrong fix

On terraform-plugin-framework v1.19.0, `resource.ValidateConfigRequest` carries only `Config` and `ClientCapabilities` — there is no `State` and no `Plan`. It therefore cannot distinguish create from update, refresh or import, and would have fired on every operation. That would have broken `TestAccMorpheusUserResourceRequiredAttrsOk` and `TestAccMorpheusUserResourceImportOk`, both of which deliberately omit `password_wo` on their later steps.

`RequireOnCreateModifier` scopes itself correctly because it runs at `PlanResourceChange`, where prior state is available:

```go
diags := req.State.GetAttribute(ctx, path.Root("id"), &id)
```

A null `id` means create, and the check fires; a populated `id` — update, refresh, import — means it does not. It reads the value from `req.ConfigValue`, which `NullifyWriteOnlyAttributes` never touches (that nulls the planned state only), so the write-only attribute is still visible to the modifier.

## Tests

Verified by acceptance rerun: all 12 tests in `morpheus/framework/resources/user` pass, nothing regressed. `Example`, `RequiredAttrsOk` and `ImportOk` all pass — these were the regression risk had the `ValidateConfig` route been taken.

## Notes

Closes MORPH-16051. The story's scope was corrected before filing to match this finding: it was originally drafted as "add create-time validation for `password_wo`", but that validation already exists — the actual defect is the test fixture.


jasonquigleyhpecom (commented):
---

## 📋 Review: `hpe_morpheus_user`

**PR:** [#1350 — Triage MORPH-16051: Fix password_wo test never reaching the plan](https://github.com/HPE/terraform-provider-hpe/pull/1350)

**Scope:** 🎯 PR diff vs `origin/dev-2.1` — 1 file · agents: patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

RDavisHPE (approved):

matthew-c-hpe (approved):

PR #1351, opened 2026-08-18
Triage MORPH-16050: Fix provider-block tests asserting a stale message

jasonquigleyhpecom:
Two provider-block acceptance tests asserted a diagnostic the provider no longer emits.

## Symptom

```
Expected an error with pattern (Must set a configuration value for the
morpheus\[0\].url attribute as the\nprovider has marked it as required.)

  | Error: Missing Morpheus connection details
--- FAIL: TestAccMorpheusProviderBlockMissingURL (0.51s)
```

`TestAccMorpheusProviderBlockNoneSet` failed the same way.

## Cause

The expected text is a schema-level required-attribute violation. `url` is now
**Optional** — an identity block may supply it — so that violation never fires.
The same condition is instead detected in `Configure()` and reported as
`Missing Morpheus connection details`.

The provider is behaving correctly; only the test expectations were stale.
`morpheus/provider.go` is deliberately unchanged.

## Fix

Both call sites now assert a single commented const matching
`Missing\s+Morpheus\s+connection\s+details`.

The assertion is pinned to the **summary line only**, never the multi-line detail
body. The summary occurs exactly once in the tree (`morpheus/provider.go:107`) and
deliberately mirrors the SDKv2 provider's message, so it is stable. The detail body
carries example HCL and identity attribute names (`pce_identity`,
`pce_disconnected_identity`) — exactly the churn-prone text that left these tests
stale in the first place. The `\s+` absorbs Terraform's line-wrapping of the summary.

## Tests

Verified by acceptance rerun: all 8 `TestAccMorpheusProviderBlock*` tests pass — the
two that were failing plus the six siblings that guard this change. Nothing regressed.

## Notes

Review nit, deliberately not actioned to keep this change test-only: the new const
is named `missingConnectionDetails`, which the `morpheus` package already uses for
the *detail* body. Different packages, so there is no conflict, but
`missingConnectionDetailsSummaryPattern` would read better.

Closes MORPH-16050.


jasonquigleyhpecom (commented):
---

## 📋 Review: `hpe_morpheus_provider`

**PR:** [#1351 — Triage MORPH-16050: Fix provider-block tests asserting a stale message](https://github.com/HPE/terraform-provider-hpe/pull/1351)

**Scope:** 🎯 PR diff vs `origin/dev-2.1` — 1 file · agents: patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

RDavisHPE (approved):

PR #1356, opened 2026-08-20
pce_disconnected_identity: rename issuer_url to token_issuer_url

matthew-c-hpe:
Renames the `issuer_url` attribute in the `pce_disconnected_identity` provider
block to `token_issuer_url`, matching the name GreenLake gives the value.

`pce_identity` is deliberately left alone: it shipped in v1.6.0, so renaming its
`issuer_url` would break existing configurations. The two blocks are configured
independently, so the divergence is contained.

This is not a breaking change — `pce_disconnected_identity` is not present in
the v1.6.0 tag, so no released configuration can be using it. No deprecation
shim or state upgrade is needed.

## Changes

- Framework provider schema: attribute renamed, along with the three
  cross-references to it (the `client_id` / `client_secret` "also requires"
  rules and the `iam_token` conflict rule). The description text is unchanged.
- Mirrored SDKv2 schema: the same key and its four `RequiredWith` /
  `ConflictsWith` paths.
- `pceDisconnectedIdentityConfig` now reads `token_issuer_url`. This is the one
  functional change rather than a rename: `main.go` replaces the SDKv2 schema
  with one derived from the framework schema, so the config keys follow the
  framework names. Left as `issuer_url` this would have silently dropped the
  issuer URL from the Disconnected token exchange, with no error and a failure
  only much later at token exchange time.
- `PCEDisconnectedIdentityModel.IssuerURL` renamed to `TokenIssuerURL`. The
  internal `pce.Config.IssuerURL` is unchanged: it is shared with the Connected
  flow and is not a Terraform attribute name.
- The shared `IdentityCredentialsOrTokenValidator` credential list now spans
  both spellings. The existing loop already skips an attribute the block does
  not have, so one list serves both blocks; the diagnostic names both.

## Tests

- Tests that cover both blocks resolved the issuer attribute per block instead
  of hardcoding one name — without this the Disconnected subtests fail on an
  unknown attribute.
- Added a validator test proving a block carrying only `token_issuer_url`, with
  no `issuer_url` key at all, is accepted.
- Parity and example-attribute tests rekeyed.
- The examples test needed a `#nosec G101`: the key now contains "token" and the
  value is a URL literal, which trips gosec's hardcoded-credentials heuristic.
  Suppressed narrowly, with justification, rather than excluding gosec from all
  test files.

## Docs

`templates/index.md.tmpl` names both spellings; the example `.tf` was renamed
and realigned; `docs/index.md` was regenerated with `make docs`.

## Verification

- `go build ./...`
- `make lint` — 0 issues
- `go test -count=1` on the affected packages — all pass
- `make docs` is idempotent, so the docs check is clean
- Audited: every remaining bare `issuer_url` is either in a `pce_identity`
  context or in a comment that deliberately names both spellings

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -43,7 +43,8 @@ For the first two the URL of the Morpheus instance must be provided as `url`.
 
 For the last two the URL and access token are obtained from GreenLake, so `url` must not be set.
 Provide a `pce_identity` or `pce_disconnected_identity` block instead, using either GreenLake API
-client credentials (`client_id`, `client_secret` and `issuer_url`) or a pre-generated `iam_token`.
+client credentials (`client_id`, `client_secret` and `issuer_url`, or `token_issuer_url` in the
+Disconnected block) or a pre-generated `iam_token`.
 The two blocks are mutually exclusive, and neither can be combined with `url`, `username`,
 `password`, `access_token` or `tenant_subdomain`.
 
@@ -223,12 +224,12 @@ provider "hpe" {
   # Provide morpheus block if you want to create morpheus resources
   morpheus {
     pce_disconnected_identity {
-      client_id     = "client_id"
-      client_secret = "client_secret"
-      issuer_url    = "https://issuer.example.com"
-      location      = "location"
-      workspace_id  = "workspace_id"
-      broker_url    = "https://broker.example.com"
+      client_id        = "client_id"
+      client_secret    = "client_secret"
+      token_issuer_url = "https://issuer.example.com"
+      location         = "location"
+      workspace_id     = "workspace_id"
+      broker_url       = "https://broker.example.com"
     }
   }
 }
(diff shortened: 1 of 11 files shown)


eamonnotoole (commented):
---

## 📋 Review: `hpe` provider

**PR:** [#1356 — pce_disconnected_identity: rename issuer_url to token_issuer_url](https://github.com/HPE/terraform-provider-hpe/pull/1356)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 11 files · agents: review-patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

eamonnotoole (approved):

PR #1357, opened 2026-08-20
Add a practitioner guide to affinity groups

eamonnotoole:
## Summary

A practitioner guide to affinity groups, published to the registry as
`docs/guides/affinity_groups.md`.

Affinity groups changed shape in v2.0.0 and the reasons are not self-evident
from the schema. This covers creating a group, getting guests into it, and
proving where they actually landed.

## Why a guide rather than more reference docs

Three things in particular cost time to work out, and none of them is visible
from an attribute description:

**Membership is a resource per member, and `servers` is read-only.** The API
treats a supplied server list as authoritative and removes anything absent from
it, so a group that named its own members would evict guests that joined by any
other route -- an instance provisioned into it, a node added to a cluster -- on
the next apply, silently. The guide gives the reasoning, not just the rule.

**Ordering has to be stated explicitly.** An instance using `affinity_group_id`
references the group, so Terraform creates the group first. Nothing tells it the
group must be *populated* first, and membership is a sibling resource rather
than a dependency of the instance. Without an explicit `depends_on` the guest
can be placed against an empty group, which decides nothing and makes the result
a race. The v1.6.0 form got that ordering for free, because the instance
referenced a group whose `servers` referenced the seed -- so this is a trap
specific to configurations being migrated.

**The two platforms apply rules at different times.** VMware builds a DRS rule
and honours it at placement. HVM records membership and lets the cluster's
placement loop reconcile afterwards, observed at anywhere from ninety seconds to
six minutes. An apply that returns with two guests still co-located has not
necessarily failed, and without that stated it looks like a bug.

## Also covered

- The two group types, and that they are not interchangeable
- Both join routes: `affinity_group_id` at provision time, membership resource afterwards
- Proving placement from `parent_host_name` rather than the UI
- What forces replacement -- including that `affinity_group_id` replaces the
  **instance**, with the `active = false` pattern for moving a running guest
  between rules without rebuilding it
- Migrating from the v1.6.0 `servers` / `ignore_changes` form
- `tenant_ids` is deprecated and has no effect, with the consequence that state
  will show tenants the appliance does not have

## Verification

Every HCL block was extracted from the rendered guide and checked with
`terraform fmt` -- all ten parse and are canonically formatted, so they are
pasteable rather than merely plausible. That check caught one example written as
an indented fragment, now a complete resource.

The behaviour described was exercised against an appliance: KEEP_TOGETHER
binding at provision time on HVM, KEEP_SEPARATE converging afterwards, and
KEEP_SEPARATE applying at placement on VMware.

`templates/` and `docs/` are in step; `make docs` produces no drift.

### DIFF
--- a/docs/guides/affinity_groups.md
+++ b/docs/guides/affinity_groups.md
@@ -1,0 +1,353 @@
+---
+page_title: "Affinity Groups"
+subcategory: "Guides"
+---
+
+## What This Guide Covers
+
+Affinity groups tell Morpheus to keep a set of guests on the same hypervisor, or
+to keep them apart. This guide covers how to express that in Terraform with the
+HPE provider: creating a group, getting guests into it, and proving where they
+actually landed.
+
+It also covers what changed in v2.0.0, because membership works differently from
+v1.6.0 and existing configuration will need editing.
+
+- [The two group types](#the-two-group-types)
+- [Membership is a resource](#membership-is-a-resource)
+- [Two ways to join a group](#two-ways-to-join-a-group)
+- [Ordering matters](#ordering-matters)
+- [Proving placement](#proving-placement)
+- [Platform differences](#platform-differences)
+- [What forces replacement](#what-forces-replacement)
+- [Migrating from v1.6.0](#migrating-from-v160)
+- [tenant_ids is deprecated](#tenant_ids-is-deprecated)
+
+---
+
+## The Two Group Types
+
+Affinity groups belong either to a cloud or to a cluster, and the provider has a
+separate resource for each. They are not interchangeable — pick the one that
+matches where your guests live.
+
+| Resource | Belongs to | Typical use |
+|---|---|---|
+| `hpe_morpheus_cloud_affinity_group` | a cloud | VMware |
+| `hpe_morpheus_cluster_affinity_group` | a cluster | HVM |
+
+```hcl
+resource "hpe_morpheus_cluster_affinity_group" "web" {
+  cluster_id    = var.cluster_id
+  name          = "web-tier"
+  affinity_type = "KEEP_TOGETHER"
+  active        = true
+}
+```
+
+```hcl
+resource "hpe_morpheus_cloud_affinity_group" "web" {
+  cloud_id      = var.cloud_id
+  pool_id       = var.pool_id
+  name          = "web-tier"
+  affinity_type = "KEEP_SEPARATE"
+  active        = true
+}
+```
+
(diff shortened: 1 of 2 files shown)


jasonquigleyhpecom (commented):
---

## 📋 Review: `hpe_morpheus_affinity_groups_guide`

**PR:** [#1357 — Add a practitioner guide to affinity groups](https://github.com/HPE/terraform-provider-hpe/pull/1357)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 2 files · agents: structure, patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 1 |
| 🔵 Recommendations | 0 |

---

### ⚙️ 4. Pattern Compliance

🟡 **Warning: Incorrect HCL guidance for Set indexing**
- **Files**: `docs/guides/affinity_groups.md`, `templates/guides/affinity_groups.md`
- **Details**: The guide text states: *"An instance may have more than one; `one()` is correct for a single-node instance, otherwise index the list explicitly."* However, the `compute_servers` attribute on `hpe_morpheus_instance` is typed as a `Set of Number`, not a `List`. In HCL, sets cannot be accessed by an index directly (e.g., attempting `compute_servers[0]` will result in a Terraform error: `Elements of a set cannot be accessed by index`).
- **Recommendation**: Rephrase the sentence to acknowledge that it is a set. Advise users that they must convert it to a list first if they want to access specific indices, e.g., *"otherwise convert the set to a list to index it explicitly (e.g., `tolist(hpe_morpheus_instance.app.compute_servers)[0]`)"*.

---

**Next:** Fix 🔴 Gaps and 🟠 Violations before merge. 🟡 Warnings are advisory. see inline comments for specifics.

jasonquigleyhpecom (commented):
---

## 📋 Review: `hpe_morpheus_affinity_groups_guide`

**PR:** [#1357 — Add a practitioner guide to affinity groups](https://github.com/HPE/terraform-provider-hpe/pull/1357)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 2 files · agents: structure, patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

jasonquigleyhpecom (approved):

PR #1358, opened 2026-08-20
Treat an unknown value as absent when converting to Go types

eamonnotoole:
Fixes MORPH-16244.

## Summary

`ValueToAny` guarded null but not unknown. An unknown `NumberValue` holds a nil
`*big.Float`, so `ValueBigFloat().Float64()` dereferenced nil and panicked the
provider.

## How it was reached

The path is a dynamic attribute. Its validator guards the value it is given:

```go
if request.ConfigValue.IsNull() || request.ConfigValue.IsUnknown() {
    return
}
```

but `config = { templateId = var.image_id }` is an object that is itself
**known** — only its member is unknown. The guard passes, `ObjectToAny` walks
the members, and the unknown number panics.

That explains a confusing symptom: a literal works, and so does arithmetic
between literals because Terraform folds it at parse time. The crash appears
only once a value is deferred to apply — a variable, a `count`/`for_each`
reference, a map lookup, `tonumber()`. Taking an id from a `.tfvars` file is
enough.

## What changed

Unknown is treated as absent, the same as null, in `ValueToAny` and in the five
collection helpers beside it (`ListToAny`, `SetToAny`, `MapToAny`,
`ObjectToAny`, `TupleToAny`), which all had the same omission.

Callers get `nil` rather than a zero value, so "not yet known" cannot be
mistaken for "set to zero".

## Verification

Reverting the guard reproduces the panic under test:

```
--- FAIL: TestUnitValueToAnyUnknown/number
panic: runtime error: invalid memory address or nil pointer dereference
```

Tests cover every type the conversion handles, plus the nested case actually
reached in practice.

## Two properties of the number conversion, pinned rather than changed

A Terraform number is a `big.Float`, and this converts through `float64`. Two
consequences are now recorded in tests so they are visible rather than
discovered:

- a whole number still arrives as `float64` — fine for JSON, which does not
  distinguish, but it is not an integer round-trip
- an integer beyond float64's exact range (2^53) is **refused** with a loss of
  precision error rather than silently rounded

The second means a sufficiently large id cannot pass through a dynamic
attribute at all. That is the safe behaviour, but it is a real limit and worth
knowing about.

## Scope

Not specific to any one attribute or provider. Any dynamic attribute given a
computed value whose conversion dereferences an unknown would panic.

### DIFF
--- a/utils/convert/valuetoany.go
+++ b/utils/convert/valuetoany.go
@@ -14,7 +14,18 @@ import (
 // This function handles null values, primitive types, and complex types
 // like objects, lists, sets, tuples and maps.
 func ValueToAny(ctx context.Context, v attr.Value) (any, error) {
-	if v == nil || v.IsNull() {
+	// Unknown is treated as absent, the same as null.
+	//
+	// A value that is not yet known cannot be converted, and the accessors do
+	// not tolerate being asked: an unknown NumberValue returns a nil
+	// *big.Float, so ValueBigFloat().Float64() dereferences nil and panics.
+	//
+	// The callers are validators, which have nothing to validate until the
+	// value is resolved. Guarding only null was enough while every value came
+	// from a literal, and became a crash the moment one came from a variable,
+	// a count or for_each reference, or any other expression deferred to
+	// apply.
+	if v == nil || v.IsNull() || v.IsUnknown() {
 		return nil, nil
 	}
 
@@ -50,7 +61,7 @@ func ValueToAny(ctx context.Context, v attr.Value) (any, error) {
 }
 
 func ListToAny(ctx context.Context, l basetypes.ListValue) ([]any, error) {
-	if l.IsNull() {
+	if l.IsNull() || l.IsUnknown() {
 		return nil, nil
 	}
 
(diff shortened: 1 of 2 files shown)


jasonquigleyhpecom (approved):

jasonquigleyhpecom (commented):
---

## 📋 Review: `utils/convert`

**PR:** [#1358 — Treat an unknown value as absent when converting to Go types](https://github.com/HPE/terraform-provider-hpe/pull/1358)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 2 files · agents: review-patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

🔵 **Recommendation: Missing `Int64Value` support in `ValueToAny`**
- **File**: `utils/convert/valuetoany.go`
- **Context**: The new test `TestUnitValueToAnyUnknown` explicitly includes a case for `"int64": types.Int64Unknown()`, which gracefully returns `nil` thanks to the new `IsUnknown()` guard at the top of the function. However, the underlying `switch val := v.(type)` block in `ValueToAny` lacks a case for `basetypes.Int64Value`. If a *known* `types.Int64Value(123)` is ever passed to `ValueToAny`, it will fall through to the `default` block and return an `"unsupported type for ValueToAny conversion"` error. Consider adding support for it alongside `NumberValue` and `Float64Value` so primitive type support is complete.

---

PR #1359, opened 2026-08-20
Mark image tenants as known so they remain distinguishable

eamonnotoole:
Fixes MORPH-16245.

## Summary

Reading a virtual image with **two or more tenants** failed with
`Duplicate Set Element` — whether or not anything was actually duplicated.

## Root cause

Tenants were built as struct literals without setting `state`:

```go
return TenantsValue{
    Name: convert.StrToType(in.Name),
    Id:   convert.Int64ToType(in.Id),
}
```

The zero value of `attr.ValueState` is `ValueStateNull`, so every tenant was a
**null object** — and a null object lowers to a null `tftypes` value regardless
of the attributes set on it. `Name` and `Id` were discarded on the way out,
leaving every tenant byte-identical to every other.

## Where the error actually comes from

Not `types.SetValue` at construction — that accepts these values. It is
`SetType.Validate`, which runs **after** the value has left the provider and
compares elements pairwise as `tftypes.Value`.

That has a direct consequence for testing: the fault is **invisible at the
`attr.Value` layer**. A test written there passes against the broken code. This
one lowers the values and compares them, which is where the difference has to
show.

## What changed

`state: attr.ValueStateKnown` when constructing the value. One field.

The mapping is now a named function rather than an inline closure, so the test
can drive the production code rather than a copy of it.

## Verification

Removing the field fails the guard with the diagnosis in the message:

```
--- FAIL: TestUnitImageTenantsAreDistinguishable
    two tenants with different ids and names compare equal;
    Terraform will reject the set as containing duplicates
```

**The tags mapper in the same file already sets `state` correctly.** So this was
an omission in one of two adjacent mappings, not a misunderstanding of the type
— which makes it the kind of thing that is likely to recur.

Other generated value types built as struct literals may carry the same
omission. That sweep is deliberately **not** attempted here; it wants doing
separately, with its own verification, because the failure mode is silent until
a collection happens to contain more than one element.

## A rejected approach, recorded so it is not revisited

Deduplicating inside `convert.ToSetType` looks like an obvious fix and is
actively harmful. Because every tenant lowers to the same null value, a dedup
pass compares elements that all appear equal, collapses every tenant to one, and
returns **successfully** — converting a loud, correct error into silent data
loss.

### DIFF
--- a/morpheus/framework/datasources/image/image.go
+++ b/morpheus/framework/datasources/image/image.go
@@ -360,15 +360,7 @@ func parseAsData(
 	data.Tags = tags
 
 	// tenants
-	tenants, d := convert.ToSetType(
-		ctx, image.Accounts,
-		func(in sdk.GetVirtualImage200ResponseVirtualImageAccountsInner) TenantsValue {
-			return TenantsValue{
-				Name: convert.StrToType(in.Name),
-				Id:   convert.Int64ToType(in.Id),
-			}
-		},
-	)
+	tenants, d := convert.ToSetType(ctx, image.Accounts, tenantValue)
 	diags.Append(d...)
 	data.Tenants = tenants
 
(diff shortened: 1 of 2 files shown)


jasonquigleyhpecom (commented):
---

## 📋 Review: `hpe_morpheus_image`

**PR:** [#1359 — Mark image tenants as known so they remain distinguishable](https://github.com/HPE/terraform-provider-hpe/pull/1359)

**Scope:** 🎯 PR diff vs `origin/dev-2.0` — 2 files · agents: patterns · UI val: skipped · connected: none

| Severity | Count |
|----------|-------|
| 🔴 Gaps | 0 |
| 🟠 Violations | 0 |
| 🟡 Warnings | 0 |
| 🔵 Recommendations | 0 |

---

jasonquigleyhpecom (approved):

PR #1360, opened 2026-08-25
Adding documentation for v2.0.0 of tfmigrator + hpegl migration

RDavisHPE:
Adding docs for new tfmigrator workflow and hpegl migration, including a table mapping each resource and datasource in hpegl to its hpe equivalent.

Also standardized 2.0.0 versioning for opsramp

### DIFF
--- a/docs/guides/hpegl_to_hpe_migration.md
+++ b/docs/guides/hpegl_to_hpe_migration.md
@@ -23,6 +23,11 @@ provider (hpegl used the Terraform Plugin SDK v2; the HPE provider uses the Plug
 resource therefore requires Terraform-generated configuration and manual post-migration review. The
 process documented for [tfmigrator](./tfmigrator_migration.md) handles this generation for you.
 
+## Prerequisites
+
+- Terraform >= 1.5.0 (required for `import` blocks)
+- HPE provider >= 2.0.0
+
 ## Provider Block Transformation
 
 The largest difference between an hpegl configuration and an HPE configuration is the provider block.
@@ -66,6 +71,17 @@ provider "hpe" {
 ### GreenLake IAM - Disconnected (`iam_version = "glp"`)
 
 ```HCL
+# Before (hpegl)
+provider "hpegl" {
+  iam_version      = "glp"
+  iam_service_url  = "https://client.greenlake.hpe.com/..."
+  vmaas {
+    location   = "HPE-Data-Center"
+    space_name = "my-space"
+    broker_url = "https://vmaas-broker.example.com"
+  }
+}
+
 # After (hpe)
 provider "hpe" {
   morpheus {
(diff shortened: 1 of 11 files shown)


eoghanino (dismissed):

eoghanino (commented):
My review came across the following: 
- `docs/guides/tfmigrator_migration.md` / `templates/guides/tfmigrator_migration.md`, "Install Scripts" links: restore a real ref in the URLs — `blob/-/scripts/...` was presumably a stray edit; use `blob/main/scripts/install-tfmigrator.sh` (and the windows/macos equivalents).
- `docs/guides/hpegl_to_hpe_migration.md` / `templates/...`, "Composite Import IDs" table: confirm and make consistent the separator — `load_balancer_pool`/`load_balancer_profile` use `{lb_id}/{id}` while every other row uses `.`; if that difference is real, add a sentence saying so, otherwise fix it.
- `docs/guides/hpegl_to_hpe_migration.md` / `templates/...`, "Direct-connect" section: the prose says "no `pce_identity` sub-block is emitted" but the preceding sections are `pce_identity` and `pce_disconnected_identity` — reword to cover both, and add the "Before (hpegl)" snippet for the `glp` and direct-connect examples so all three sections match the format of the `glcs` one.
- `docs/guides/hpegl_to_hpe_migration.md` / `templates/...`, "Prerequisites"-equivalent content: add the `hpe` provider version constraint (`>= 2.0.0`) that the other migration guides now state, so this guide isn't the only one without it.
- `docs/guides/tfmigrator_migration.md` / `templates/...`, "Full Pipeline (`migrate`)" paragraph: split the sentence beginning "As part of this it also removes..." — the parenthetical ordering explanation plus the "keep the output quiet" aside is hard to follow; state the ordering as a short list.
- `docs/guides/tfmigrator_migration.md` / `templates/...`, "Command Summary": `migrate`'s bullet lists the pipeline as ending at `migrate-datasources` while the following text says `migrate` also runs `cleanup-providers`; include it in the arrow chain.
- `scripts/install-tfmigrator-windows.ps1`, `$extract_dir`: `Expand-Archive $archive_path -Force` with no `-DestinationPath` extracts relative to the current location, but `$extract_dir` is a bare name while `$archive_path` is under `$tmp_dir` — verify the binary path still resolves after the rename.

RDavisHPE:
Thanks for the review @eoghanino — addressed the actionable points:

- **Direct-connect section:** reworded the prose to cover both sub-blocks (now reads "neither a `pce_identity` nor a `pce_disconnected_identity` sub-block emitted"), and added `# Before (hpegl)` snippets to the `glp` and direct-connect sections so all three match the `glcs` before/after format. Field names verified against the tfmigrator `provider_migration.yaml` mappings.
- **Provider version constraint:** added a `## Prerequisites` section to the hpegl guide (Terraform `>= 1.5.0`, HPE provider `>= 2.0.0`), matching the opsramp guide.
- **Full Pipeline (`migrate`) paragraph:** split the run-on "As part of this it also removes…" sentence into a short ordered list (before/after `migrate-datasources`), with the interleave rationale as a trailing note.
- **Command Summary:** the `migrate` bullet's arrow chain now includes `→ cleanup-providers`, consistent with the individual-steps chain.

Not changed, intentionally:

- **Install Scripts links (`blob/-/scripts/...`):** this is not a stray edit — leaving as-is.
- **Composite Import IDs separator (`{lb_id}/{id}`):** these discriminators have already shipped, so we can't change them now.
- **`install-tfmigrator-windows.ps1` `$extract_dir`:** out of scope for this docs PR.

Docs regenerated via `make docs`; `templates/` and `docs/` are in sync.

eoghanino (approved):

